PR #273, opened 2023-06-28
Implement `ots_getContractCreator`

JamesHinshelwood:
This API requires doing some time travel and transaction replays. If contracts were not able to create contracts themselves, this would be simpler, because the contract creator would always be the `from_addr` of the creation transaction. However, this is not the case so we need to replay transactions with a custom event listener to work out which account created a contract.


JamesHinshelwood:
Will come back to this later.

PR #366, opened 2023-08-08
Include the high_qc in persistence

theo-zil:
Probably worth waiting till after mutable committee is merged to reduce conflict burden


theo-zil:
Closing this since it was decided we should just derive it from the latest known block.

PR #374, opened 2023-08-14
Abstract database serialization and access into separate module

JamesHinshelwood:
All references to `sled` and (de)serialization code now live inside `db::Db`, which exposes a simpler interface to `Consensus`, `BlockStore` and `State`. These structs now store an `Arc<Db>`.

I renamed the existing `SledDb` to `TrieStorage` to avoid confusion with my new struct and to hopefully make the purpose clearer.

We make use of `paste` in the implementation of `Db` to cut down on some repetition. Feel free to complain if you think this doesn't help.


zilliqa/src/db.rs
```diff
@@ -1,19 +1,181 @@
...
+    block: Tree,
+    canonical_block_number: Tree,
+    transaction: Tree,
+    transaction_receipt: Tree,
+    /// An index of address to a list of transaction hashes, for which this address appeared somewhere in the
+    /// transaction trace. The list of transations is ordered by execution order.
+    touched_address_index: Tree,
+}
+
+macro_rules! get_and_insert_methods {
```
n-hutton, line 25: Not too sure I am a big fan of this macro as it just reduces a little duplication, but if it is going to be in here could it add a commment?

theo-zil (approved):
Definitely a welcome change, would be great to get this in.

Don't have a strong opinion on the macro either way.

theo-zil (approved):
I've come to agree with Nathan's point regarding the macros I think, but merging this in now as refactoring the macros out will probably cause relatively minimal disruption and/or can be done at a point where not too many open PRs are open.

n-hutton (approved):

PR #418, opened 2023-09-07
Define infrastructure for ZQ2 development network

JamesHinshelwood:
See https://github.com/Zilliqa/devops/pull/736 for usage example.

### DIFF
--- a/infra/tf/.gitignore
+++ b/infra/tf/.gitignore
@@ -1,0 +1,3 @@
+.terraform/
+.terraform.lock.hcl
+terraform.tfstate*.backup
--- a/infra/tf/apps.tf
+++ b/infra/tf/apps.tf
@@ -1,0 +1,107 @@
+## Ancilliary applications for ZQ2 networks are deployed to a GKE cluster.
+
+resource "google_compute_network" "zq2_apps" {
+  name                    = "zq2-apps"
+  auto_create_subnetworks = false
+}
+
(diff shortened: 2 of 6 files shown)


zilliqa/src/state.rs
```diff
@@ -45,7 +45,7 @@ impl State {
         let _ = state.set_gas_price(default_gas_price().into());
 
         for (address, balance) in genesis_accounts {
-            state.set_native_balance(*address, (*balance).into())?;
```
theo-zil, line 48: I'm not very familiar with TF so I'm sure there must've been a good reason to change this to a `String`, but this seems like the kind of thing I'd love to validate early rather than late. Maybe even with a custom serde deserializer function specified in cfg.rs, to validate it on parse? Actually, I'm not sure off the top of my head, but can it be specified as a string but deserialized directly into an `u128`?

I'm just not super keen on having a string that can fail parsing at any time be stored in the config.

But also pragmatically speaking we are almost certainly going to construct the `State` very soon after booting the node, so in practical terms it will still fail quite fast, so only worth looking into if you think you can add the extra type-safety without too much effort.

mauromedda (approved):
What I will do is probably remove the hardcoded values re: domain names, region and CIDR probably moving them as default values in the variables to let 3rd parties to customize the deployment and run their ZQ2 devnets on GCP. That said, it's currently fitting the purpose of what we need so I am approving it :)

PR #433, opened 2023-09-12
Implement `eth_getLogs`

JamesHinshelwood:
The current implementation is fairly basic and slow. We just iterate through each block, receipt and log, filtering results out when we can.


zilliqa/src/api/eth.rs
```diff
@@ -273,6 +276,119 @@ fn get_block_transaction_count_by_number(
     Ok(block.map(|b| b.transactions.len().to_hex()))
 }
 
+#[derive(Deserialize, Default)]
+#[serde(default, rename_all = "camelCase")]
+struct GetLogsParams {
+    from_block: Option<BlockNumber>,
+    to_block: Option<BlockNumber>,
+    address: Option<OneOrMany<H160>>,
```
theo-zil, line 284: With `Option<OneOrMany<>>`, is there a meaningful difference between `None` and `Some(Null)`? Otherwise the `Option<>` seems superfluous here (unless this is necessary for param parsing)

zilliqa/src/api/eth.rs
```diff
@@ -273,6 +276,119 @@ fn get_block_transaction_count_by_number(
...
+                .enumerate()
+                .map(move |(i, t)| (i, t, block_number, block_hash)))
+        })
+        .flatten_ok();
+
+    // Get the receipts for each transaction hash. This is an iterator of (receipt, txn_index, txn_hash, block_number, block_hash).
+    let receipts = txn_hashes.map(|h: Result<_>| {
+        let (txn_index, txn_hash, block_number, block_hash) = h?;
+        let receipt = node
+            .get_transaction_receipt(txn_hash)?
```
theo-zil, line 343: Since I recently changed the way receipts are stored, I think it's likely worth  using `node.get_transaction_receipts_in_block()` rather than collecting txn_hashes and then individual receipts for each txn.

theo-zil (changes requested):

zilliqa/src/api/eth.rs
JamesHinshelwood: Good idea, thanks for spotting.

zilliqa/src/api/eth.rs
JamesHinshelwood: There is a difference (though maybe not a useful one)
https://github.com/Zilliqa/zq2/blob/3655674d671cf0b70bb7c2a9c0e652446ec8bf1a/zilliqa/src/api/eth.rs#L360-L364

When `params.address = None`, this returns `true`. This means the log in question is always included.

When `params.address = Some(Null)`, this returns `false` because `OneOrMany::Null.contains(x) == false` for all `x`. This means the log in question is always excluded.

So passing `null` for the `address` will always result in an empty response, while passing nothing will do no extra filtering. Arguably we could return an error in the `null` case because its always nonsense, but I left it in for consistency with the `topics: Vec<OneOrMany<_>>` parameter.

zilliqa/src/api/eth.rs
JamesHinshelwood: Ah one sticking point is I need to know the index of the transaction in the block, which `TransactionReceipt` currently doesn't contain. Do you think it makes sense to add that just for this API? On the one hand it is basically redundant data if you consider the `TransactionReceipt` tied to the `Transaction`. However, it does mean we can implement this slightly more efficiently...

zilliqa/src/api/eth.rs
theo-zil: Makes sense, cheers for the explanation. If this is something API clients might feasibly expect then certainly makes sense to keep it.

zilliqa/src/api/eth.rs
theo-zil: Ah yes I see the problem. Hmm - honestly I don't see any downside to including it in the TransactionReceipt. The index isn't stored on `Transaction`s either, and that means that any time we'd want it, we'd have to fetch the block, iterate through the block's tx hash list and find the position of the tx. Storing it in the receipt makes the most sense to me.

JamesHinshelwood: Ah I found I'd also need to add the block's number to the receipts as well as the block hash. I'm wondering whether the extra complexity is worth it. The fact that we need to add this 'redundant' data to the receipts makes me feel un-easy like our data model is not quite right/natural.

@n-hutton It might be useful to get your opinion here too.

theo-zil: Hm, what about querying the block number (using the hash) for every block in the RPC?
So if I understand it right, something along the lines of (in pseudocode)
```
let receipts: Vec<Vec<(Receipt, u64)>> = blocks.map(|block| {
  (block.get_transaction_receipts_in_block(), block.view())
});
let logs = receipts.map(|(receipt_vec, view)| {
  receipt_vec.map(|receipt| {
    receipt.logs.map(|log| { view, ...log })
  })
}).flatten().flatten()
// map logs to output format
```

JamesHinshelwood: That makes sense - I was trying to avoid getting the block at all and just sticking with the block hash, but your suggestion makes more sense. Updated the PR, please take another look when you have a moment :)

n-hutton (approved):
Kinda annoying the eth api forces this OneOrMany pattern. Great that this PR includes testing for emmitted eth events as well

zilliqa/src/api/eth.rs
theo-zil: Ah I see! I think your previous commit was already fetching the blocks though which is why I suggested it this way.

Either way looks good now!

theo-zil (approved):

PR #437, opened 2023-09-14
improve logging in evm

n-hutton:


zilliqa/src/exec.rs
```diff
@@ -280,7 +286,10 @@ impl State {
             }
 
             backend.origin = call_args.caller;
-            result = run_evm_impl_direct(call_args.clone(), &backend);
+
+            result = tracing::subscriber::with_default(tracing_subscriber(), || {
```
n-hutton, line 290: I didn't see a better way to do it than this - ideally, we would be able to get the global subscriber, but there didn't appear to be a way and I got bored of looking.

PR #438, opened 2023-09-14
Consensus from genesis hash

theo-zil:
Small intermediate PR which adjusts consensus such that it becomes possible to launch a node (on an existing network) supplying only the genesis hash.
This avoids having to specify the public keys and peer IDs of the genesis committee, and additionally is a building block towards later specifying the genesis in a contract on the main shard to allow configuration-less joining of child shards.


JamesHinshelwood:
What is the use case for specifying a genesis hash, rather than the full genesis block? Surely to generate (and trust) the genesis hash you would need to know the full block anyway?

theo-zil:
To generate, definitely. But to trust, I feel like having just the hash is much easier/more convenient to specify in a config, rather than the details of the full committee. Basically a matter of ergonomics.
Once you have the hash, you can have the node fetch the genesis block and trust it if it matches, then proceed as normal.

JamesHinshelwood:
Sorry it took me 2 weeks to reply here :sob: 

> To generate, definitely. But to trust, I feel like having just the hash is much easier/more convenient to specify in a config, rather than the details of the full committee. Basically a matter of ergonomics.
Once you have the hash, you can have the node fetch the genesis block and trust it if it matches, then proceed as normal.

I think the detail I am still iffy about is that in order to trust a hash, you need to download the block and check it matches. But you need to check it matches against _something_, and that thing is the genesis block itself, so just having the hash doesn't achieve anything?

JamesHinshelwood:
Regarding the goal of this PR:
> This avoids having to specify the public keys and peer IDs of the genesis committee, and additionally is a building block towards later specifying the genesis in a contract on the main shard to allow configuration-less joining of child shards.

Would it not still be possible to do this, even if the genesis block was specified in full? The contract on the main shard could include the genesis block?

JamesHinshelwood:
cc @DrZoltanFazekas since it might be useful to have a second opinion here, in case I'm missing something.

theo-zil:
> Sorry it took me 2 weeks to reply here 😭

No worries :)

> I think the detail I am still iffy about is that in order to trust a hash, you need to download the block and check it matches. But you need to check it matches against _something_, and that thing is the genesis block itself, so just having the hash doesn't achieve anything?

The way I'm thinking it, the hash trust comes from external channels. I.e. the human operator sets the genesis hash in their config from knowledge they get e.g. from a website. Or maybe after mainnet launches we hardcode that in the codebase, and then the trust comes from running the official binary (or compiling signed code from the official repository).
You then use that trust to verify the genesis block you get, extend the trust to _that_, and from there you can trust the rest of the blockchain you download.
Another way to look at it is that the genesis hash uniquely identifies a blockchain. If your genesis hash is wrong, chances are you just won't be able to start a node, but if it did match a different genesis block, you would simply follow a different blockchain*.

Largely speaking I think specifying the genesis hash is nearly equivalent to specifying the genesis committee, the two main differences are that knowing the committee lets you _create_ the genesis block while knowing only the hash merely lets you verify it, and that the hash is a more compact value that's more convenient to communicate compared to an array of public key peer ID pairs. The trust issues are the same, though: you trust the configured genesis hash as much as you would have to trust the provided genesis committee data (i.e., you basically assume it's correct for the intended blockchain, and if not then that's operator error).

> Would it not still be possible to do this, even if the genesis block was specified in full?

That's true. This is essentially a tradeoff in slightly higher complexity when constructing a `Consensus` vs. ergonomics when specifying configs.
But if you think the complexity isn't worth it then yeah we can definitely scrap this and use the full pubkey-peerID vec in the same contexts.

*In some very niche scenario I can actually imagine not having the full genesis committee to be an advantage: if you're a new operator and you intend to start several nodes on an existing network/shard, but you happen to be partitioned from the rest of the network (internet connection outage?), nodes configured with the full committee will happily construct a new genesis block, start building from there, and you will have issues once network connectivity is restored. Nodes configured only with the hash will fail early letting you know there is a problem.

JamesHinshelwood:
@theo-zil Thanks for the explanation, that makes a lot of sense and I agree :+1:

JamesHinshelwood (approved):

PR #439, opened 2023-09-14
Upgrade tx tracing and mempool filtering

### LINKED ISSUE #434: State accounts don't reflect funds contract
When you make a transfer to an address, if that is the first time the address gets funds, the state should create an account with nonce 0 (and never delete it presumably). This is so state accounts reflects who can actually execute TXs

n-hutton:
Upgrade the tx tracing to connect the evm generated traces - we can monitor how this effects performance.

Further to discussions about the mempool, use this upgraded tracing to determine the accounts that are being 'created' and put them in the state map. Closes #434 . Comments inline


evm-ds/src/evm_server_run.rs
```diff
@@ -27,7 +27,7 @@ use crate::tracing_logging::{CallContext, LoggingEventListener};
 fn build_exit_result<B: Backend>(
     executor: CpsExecutor<B>,
     runtime: &Runtime,
-    trace: &LoggingEventListener,
+    trace: Arc<Mutex<LoggingEventListener>>,
```
n-hutton, line 30: This was changed so that before you start, you create this listener, and all evm executions point to and build on the same one, essentially.

n-hutton: The listener was changed from how it was because previously it was ser/deser when communicating with the cpp code, but actually we don't need that overhead.

A slightly awkward thing is that it needs to be present in both the args, so the evm sees it, and in the result, so it gets passed back to consensus.

evm-ds/src/precompiles/scilla_call.rs
```diff
@@ -87,7 +87,9 @@ fn scilla_call_common(
     let transitions = code["contract_info"]["transitions"].to_owned();
     let Some(transitions) = transitions.as_array() else {
         return Err(PrecompileFailure::Error {
-            exit_status: ExitError::Other(Cow::Borrowed("Unable to get transitions array from contract json")),
+            exit_status: ExitError::Other(Cow::Borrowed(
```
n-hutton, line 90: my linter seems to prefer this somehow

evm-ds/src/tracing_logging.rs
```diff
@@ -166,10 +192,10 @@ impl evm::runtime::tracing::EventListener for LoggingEventListener {
                 input,
             } => {
                 intern_trace = Some(InternalOperationOtter {
-                    call_type: call_depth,
-                    from: format!("{:?}", address),
-                    to: format!("{:?}", target),
-                    value: format!("{:0X?}", balance),
+                    call_type: 0,
```
n-hutton, line 195: This was actually a bug in the old evm too which was fixed there and not here

evm-ds/src/tracing_logging.rs
```diff
@@ -166,10 +192,10 @@ impl evm::runtime::tracing::EventListener for LoggingEventListener {
...
+                    from: address,
```
n-hutton, line 196: since we don't ser/deser across a jsonrpc we can use native H160 etc

zilliqa/src/consensus.rs
```diff
@@ -434,13 +437,19 @@ impl Consensus {
...
         // already executed it in the process of proposing this block.
         if !self.transactions.contains_key(hash.0)? {
-            let mut listener = TouchedAddressEventListener::default();
-            let result = evm_ds::evm::tracing::using(&mut listener, || {
-                self.state
-                    .apply_transaction(txn.clone(), self.config.eth_chain_id, current_block)
-            })?;
-            for address in listener.touched {
-                self.touched_address_index.merge(address.0, hash.0)?;
+            let result = self.state.apply_transaction(
```
n-hutton, line 440: we no longer use the touchedAddr..Listener as the infos is now contained in the tx result.

zilliqa/src/exec.rs
```diff
@@ -574,13 +598,27 @@ impl State {
...
     }
 
     // Apply the changes the EVM is requesting for
-    fn apply_delta(&mut self, applys: Vec<evm_ds::protos::evm_proto::Apply>) -> Result<()> {
+    fn apply_delta(
+        &mut self,
+        applys: Vec<evm_ds::protos::evm_proto::Apply>,
+        new_addresses: Vec<Address>,
+    ) -> Result<()> {
+        // these accounts have been 'created' by having funds sent to them. We need to create them with nonce 0
```
n-hutton, line 613: we now require to know which addresses have been potentially created by a TX so that we can make sure that there is always an account that maps to the funding contract (so we can determine whether TXs 'from' are valid).

I actually would slightly like the idea of caching the balance of accounts too, that way you don't need to execute the evm for a balance check. what do people think?

JamesHinshelwood: IMO we should avoid adding a cache before we've shown that executing the EVM to calculate balances is a bottleneck (and even then the answer might be to speed up our EVM).

JamesHinshelwood: Regarding needing to know which accounts are 'created'. It strikes me as peculiar that we need to know. The mental model I have right now is `state: Map<Address, Account>` and accounts get accessed as `state.get(address).unwrap_or(Account::default_with_zero_balance_and_nonce())` (i.e. if an account doesn't exist in the map, we default it to the new account and if it gets mutated, it will be written back to state).

There are quite a few cases in which an account could have state, without native funds being sent to them (e.g. `CREATE2` deployments) and I'm apprehensive about us trying to enumerate those cases.

Would a generic approach in which we assume a default account if it doesn't exist work?

n-hutton: The cache wouldn't be purely (or mainly) for performance reasons but for scenarios like:
- to test if you have enough gas to execute a tx you can do it without evm execution which reduces logs
- If you are debugging more intuitive/easy to dump the accounts + balances and inspect
- calls to eth_getBalance/gasPrice which are common you likely don't care about seeing the evm churning away for
-when the evm calls to ask what balances are it doesn't have to recurse by making another call

So it's just quality of life, but just an idea.

Wrt knowing which accounts are 'created' - it is less complicated than that. All it is doing is if a transfer is made to a location, if there is currently no account there, make one. Again, it is just quality of life but I am not tied to it

zilliqa/src/exec.rs
```diff
@@ -574,13 +598,27 @@ impl State {
...
+        // get_account creates a defaulted account if it doesn't exist
+        for address in new_addresses {
+            let account = self.get_account(address).unwrap();
+            if account.nonce == 0 {
+                self.save_account(address, account)?;
+            }
```
JamesHinshelwood, line 619: Won't we save the account in `self.apply_delta` anyway?

JamesHinshelwood (approved):
LGTM with the caveats we discussed yesterday about tracing and account storage.

PR #451, opened 2023-09-20
Implement Zilliqa `GetNetworkId` API

JamesHinshelwood:


zilliqa/src/api/zilliqa.rs
```diff
@@ -21,6 +22,17 @@ fn get_current_mini_epoch(_: Params, node: &Arc<Mutex<Node>>) -> Result<String>
     Ok(node.lock().unwrap().view().to_string())
 }
 
+fn network_id(eth_chain_id: u64) -> u64 {
+    // We fix the convention the Zilliqa network ID is equal to the Ethereum chain ID minus 0x8000. This is true for
```
theo-zil, line 26: Huh, so Zilliqa network IDs aren't interoperable with Ethereum chain IDs, unless you do the conversion back?
If this is how ZQ1 is doing it then sure, I just didn't expect that.

theo-zil (approved):

zilliqa/src/api/zilliqa.rs
JamesHinshelwood: The IDs are different, but so are the APIs and transaction formats you pass them in, so yes they are completely distinct.

PR #453, opened 2023-09-20
Implement Zilliqa `GetLatestTxBlock` API

JamesHinshelwood:
I did quite a bit of moving around of the existing Ethereum API types, to make it clearer what was specific to the Ethereum/Zilliqa APIs. This might cause some noise in the diff, I'm happy to split this up if it causes issues with reviews.

### DIFF
--- a/zilliqa/src/api/erigon.rs
+++ b/zilliqa/src/api/erigon.rs
@@ -5,13 +5,13 @@ use jsonrpsee::{types::Params, RpcModule};
 
 use crate::node::Node;
 
-use super::types::EthBlock;
+use super::types::eth;
 
 pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
     super::declare_module!(node, [("erigon_getHeaderByNumber", get_header_by_number)])
 }
 
-fn get_header_by_number(params: Params, node: &Arc<Mutex<Node>>) -> Result<Option<EthBlock>> {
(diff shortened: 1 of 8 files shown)


zilliqa/src/api/erigon.rs
```diff
@@ -5,13 +5,13 @@ use jsonrpsee::{types::Params, RpcModule};
...
 
-use super::types::EthBlock;
+use super::types::eth::Block;
 
 pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
     super::declare_module!(node, [("erigon_getHeaderByNumber", get_header_by_number)])
 }
 
-fn get_header_by_number(params: Params, node: &Arc<Mutex<Node>>) -> Result<Option<EthBlock>> {
+fn get_header_by_number(params: Params, node: &Arc<Mutex<Node>>) -> Result<Option<Block>> {
```
theo-zil, line 14: I like the explicit specification of `eth::Block` in the other files, perhaps we should keep it here just for visual clarity at a glance?

I like the idea of having a convention of specifying the namespace inline for any types which whose concepts might be shared between the two APIs (so blocks, logs, transactions and similar).

JamesHinshelwood: Agreed, will change

JamesHinshelwood: Done - I also did the same for the `Otterscan`-prefixed types in `eth.rs` and moved them to `ots.rs` instead.

PR #466, opened 2023-09-28
Consensus async start

n-hutton:
various changes to consensus that allow it starting with a single node and allowing other nodes to join later. With this PR, you can start one node with a flag telling it to do its own buisness and it will just go. If not, then it will wait for others to come onto the network. They request to join the committee at some point down the line rather than getting included in the first block.


zilliqa/src/message.rs
```diff
@@ -170,6 +171,7 @@ impl ExternalMessage {
             ExternalMessage::NewTransaction(_) => "NewTransaction",
             ExternalMessage::RequestResponse => "RequestResponse",
             ExternalMessage::JoinCommittee(_) => "JoinCommittee",
+            ExternalMessage::Hello(_) => "Hello",
```
n-hutton, line 174: Now nodes send a hello rather than join request every so often

zilliqa/src/node.rs
```diff
@@ -119,10 +119,21 @@ impl Node {
...
                 ExternalMessage::Proposal(m) => {
+
+                    let m_view = m.header.view;
+
                     if let Some((leader, vote)) = self.consensus.proposal(m)? {
                         self.reset_timeout.send(())?;
                         self.message_sender
                             .send_external_message(leader, ExternalMessage::Vote(vote))?;
+                    } else {
+                        if m_view % 5 == 0 {
```
n-hutton, line 130: This is quite crude but just to show that the nodes don't need to be included immediately

zilliqa/src/consensus.rs
```diff
@@ -302,27 +298,63 @@ impl Consensus {
     }
 
     fn download_blocks_up_to(&mut self, to: u64) -> Result<()> {
+
         for view in (self.view + 1)..to {
             self.view = view;
             self.block_store.request_block_by_view(view)?;
         }
-        self.view = to + 1;
+
```
n-hutton, line 306: The prev logic here could allow you to effectively decrement your view

zilliqa/src/consensus.rs
```diff
@@ -302,27 +298,63 @@ impl Consensus {
...
+            self.view = to + 1;
+        }
 
         Ok(())
     }
 
-    pub fn timeout(&mut self) -> Result<(PeerId, NewView)> {
+    pub fn timeout(&mut self, peers_joined: bool) -> Option<((PeerId, ExternalMessage))> {
+
+        if self.view == 1  {
```
n-hutton, line 321: It doesn't make sense to increment the view past 1, so either do nothing, or if a peer has just joined, create a vote for genesis (starting the chain effectively)

zilliqa/src/consensus.rs
```diff
@@ -567,8 +556,13 @@ impl Consensus {
         trace!(block_view, self.view, %block_hash, "handling vote");
 
         // if we are not the leader of the round in which the vote counts
-        if block.committee.leader(block_view).public_key != self.secret_key.node_public_key() {
-            trace!(vote_view = block_view + 1, "skipping vote, not the leader");
+        // The vote is in the happy path (?) - so the view is block view + 1
+        if !self.are_we_leader_for_view(block_hash, block_view + 1) {
```
n-hutton, line 560: see slack discussion

zilliqa/src/consensus.rs
```diff
@@ -677,38 +679,103 @@ impl Consensus {
...
+            warn!(
+                "committee is too small {}, something might be wrong",
+                committee.len()
+            );
+        }
+
+        committee
+    }
+
+    fn are_we_leader_for_view(&mut self, parent_hash: Hash, view: u64) -> bool {
```
n-hutton, line 756: I found the old way of getting the leader by using the view didn't handle when the views were not contiguous well, because it assumed there would be a block for view - 1 - which there isn't when a proposer has failed.

zilliqa/src/consensus.rs
```diff
@@ -677,38 +679,103 @@ impl Consensus {
...
         };
         new_view.verify(sender.public_key)?;
 
         // check if the sender's qc is higher than our high_qc or even higher than our view
         self.update_high_qc_and_view(false, new_view.qc.clone())?;
 
-        let committee_size = self.committee()?.len();
+        let committee_size = committee.len();
+
+        if committee_size > 4 {
```
n-hutton, line 825: I noticed the bitvec sometimes allocated 1024 bits for some reason, I wondered if it was committee size (nothing turned up though)

zilliqa/src/consensus.rs
```diff
@@ -736,6 +809,7 @@ impl Consensus {
             qcs.push(new_view.qc);
             // TODO: New views broken
             supermajority = cosigned_weight * 3 > 99999 /* total weight of what committee? */ * 2;
```
n-hutton, line 861: This seems like an odd calculation?

zilliqa/src/consensus.rs
```diff
@@ -746,6 +820,14 @@ impl Consensus {
                 "storing vote for new view"
             );
             if supermajority {
+                if self.view < new_view.view {
```
n-hutton, line 873: This allows you to potentially catch up if your view is less but you see everyone agreed on a higher view

zilliqa/src/consensus.rs
```diff
@@ -759,6 +841,8 @@ impl Consensus {
                     let parent = self
                         .get_block(&parent_hash)?
                         .ok_or_else(|| anyhow!("missing block"))?;
+
+                    // why does this have no txn?
                     let proposal = Block::from_agg(
```
n-hutton, line 896: don't we want to put txs here?

zilliqa/src/consensus.rs
```diff
@@ -1290,11 +1393,12 @@ impl Consensus {
         Ok(block.committee.leader(view))
     }
 
-    fn check_quorum_in_bits(&self, view: u64, cosigned: &BitSlice) -> Result<()> {
-        let committee = &self
-            .get_block_by_view(view - 1)?
-            .ok_or_else(|| anyhow!("missing block"))?
-            .committee;
+    fn check_quorum_in_bits(
+        &self,
```
n-hutton, line 1447: missing block errors were more common before, now you refer to the parent's committee directly so it is less error prone

zilliqa/src/cfg.rs
```diff
@@ -45,6 +45,8 @@ pub struct NodeConfig {
     pub tx_retries: u64,
     /// The maximum duration between a recieved block's timestamp and the current time. Defaults to 10 seconds.
     pub allowed_timestamp_skew: Duration,
+    /// Whether to allow a single node to generate its own chain.
+    pub allow_single_node_network: bool,
```
JamesHinshelwood, line 49: IMO there should be no reason a network with a single node doesn't just work out of the box without any configuration.

zilliqa/src/node.rs
```diff
@@ -119,10 +119,18 @@ impl Node {
         match message {
             Message::External(external_message) => match external_message {
                 ExternalMessage::Proposal(m) => {
+                    let m_view = m.header.view;
+
                     if let Some((leader, vote)) = self.consensus.proposal(m)? {
                         self.reset_timeout.send(())?;
                         self.message_sender
                             .send_external_message(leader, ExternalMessage::Vote(vote))?;
+                    } else if m_view % 5 == 0 {
```
JamesHinshelwood, line 128: Maybe it makes sense to keep this logic inside consensus instead? It feels a bit magic here to trigger this behaviour based on not responding to a proposal.

zilliqa/src/node_launcher.rs
```diff
@@ -123,12 +121,8 @@ impl NodeLauncher {
...
                     trace!("timeout elapsed");
 
-                    if !joined {
-                        self.outbound_message_sender.send((None, self.config.eth_chain_id, Message::External(ExternalMessage::JoinCommittee(self.secret_key.node_public_key())))).unwrap();
-                        joined = true;
-                    } else {
-                        self.node.lock().unwrap().handle_timeout().unwrap();
-                    }
+                    self.node.lock().unwrap().handle_timeout().unwrap();
+                    self.outbound_message_sender.send((None, self.config.eth_chain_id, Message::External(ExternalMessage::Hello(self.secret_key.node_public_key())))).unwrap();
```
JamesHinshelwood, line 125: Does this mean we send both a `NewView` (if we're already part of the committee) and a `Hello` on timeout?

zilliqa/src/p2p_node.rs
```diff
@@ -260,8 +259,8 @@ impl P2pNode {
 
                     let topic = Self::shard_id_to_topic(shard_id);
 
-                    // Push messages back into queue if there are no peers
-                    if self.swarm.behaviour().gossipsub.all_peers().collect_vec().is_empty() {
+                    // Push messages back into queue if there are no peers and it is not to self
+                    if self.swarm.behaviour().gossipsub.all_peers().collect_vec().is_empty() && dest != Some(from) && dest.is_some()  {
```
JamesHinshelwood, line 263: nit: The `.is_some()` is unnecessary and I think its easier to read the logic if we exclude self-sends first.
```suggestion
                    if dest != Some(from) && self.swarm.behaviour().gossipsub.all_peers().collect_vec().is_empty() {
```

zilliqa/src/consensus.rs
JamesHinshelwood: Maybe `self.view = self.view.max(to + 1)` is simpler?

zilliqa/src/consensus.rs
```diff
@@ -306,23 +306,59 @@ impl Consensus {
...
+            self.view = to + 1;
+        }
 
         Ok(())
     }
 
-    pub fn timeout(&mut self) -> Result<(PeerId, NewView)> {
+    pub fn timeout(&mut self, peers_joined: bool) -> Option<(PeerId, ExternalMessage)> {
+        if self.view == 1 {
+            if self.config.allow_single_node_network || peers_joined {
```
JamesHinshelwood, line 324: IMO we should remove `peers_joined` along with the `allow_single_node_network` configuration and just always execute this block.

zilliqa/src/consensus.rs
```diff
@@ -306,23 +306,59 @@ impl Consensus {
...
+                    .unwrap()
+                    .ok_or_else(|| anyhow!("missing block"))
+                    .unwrap();
+                // If we're in the genesis committee, vote again.
+                if genesis
+                    .committee
+                    .iter()
+                    .any(|v| v.peer_id == self.peer_id())
+                {
+                    warn!("timeout in view 1, we should vote for genesis block rather than incrementing view");
```
JamesHinshelwood, line 336: I don't think this should be a warning, since it will happen whenever you start a new network (unless I'm misunderstanding?)

zilliqa/src/cfg.rs
n-hutton: I agree, although as it currently stands, it can be that the network will appear to work but in fact it is just running with the one validator and the others fail to sync/join the committee. Having it explicitly set whether you allow only single validator avoids this. If you want I can take it out since it's not likely to happen the way you use it (start everyone at same instant)

zilliqa/src/consensus.rs
n-hutton: righto

zilliqa/src/node.rs
n-hutton: This would also mean that you could send multiple messages and/or of different types when you recieve one - for example sending JoinCommittee instead of Vote when you see a block proposal but you are not part of committee

zilliqa/src/node_launcher.rs
n-hutton: Yes, although I will remove the `Hello` I think.

theo-zil:
Just a quick comment regarding nodes failing to join the network and starting their own geneses: a config setting to allow creating the genesis block would make more sense to me (e.g. `allow_constructing_genesis` or similar). Couple this with some visible error messages thrown if a node is a) not allowed to create genesis and b) failed to join an existing network - possibly a panic-quit after a few failed attempts - and this should make things visible if the network fails to form, while as a bonus also generalising for larger networks in the future.
E.g. say `allow_single_node_network` is false, you start 10 nodes, something goes wrong when they try to join each other, and they partition into one chain with 6 nodes and another with 4. If instead all but one have `allow_constructing_genesis` set to false, then only one chain will ever launch, and any nodes that didn't join will visibly error out.

PR #472, opened 2023-10-03
FIX: Solc Shanghai Version check

WuBruno:
Provides a better error requiring Solc version 0.8.20 running on the host system

### DIFF
--- a/zilliqa/tests/it/main.rs
+++ b/zilliqa/tests/it/main.rs
@@ -528,7 +528,7 @@ async fn deploy_contract(
     if let Ok(version) = sc.version() {
         if version.cmp(&SHANGHAI_SOLC).is_lt() {
             panic!(
-                "Solc Version {} required, currently set {}",
+                "solc version {} required, currently set {}",
                 SHANGHAI_SOLC, version
             );
         }
--- a/zilliqa/tests/it/main.rs
+++ b/zilliqa/tests/it/main.rs
@@ -499,6 +499,7 @@ fn format_message(
 }
 
 const PROJECT_ROOT: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/");
+const EVM_VERSION: EvmVersion = EvmVersion::Shanghai;
 
 async fn deploy_contract(
     path: &str,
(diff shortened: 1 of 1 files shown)


zilliqa/tests/it/main.rs
```diff
@@ -518,12 +519,21 @@ async fn deploy_contract(
...
 
-    let sc = ethers::solc::Solc::default();
+    let sc: ethers::solc::Solc = ethers::solc::Solc::default();
 
     let mut compiler_input = CompilerInput::new(contract_file.path()).unwrap();
     let compiler_input = compiler_input.first_mut().unwrap();
     compiler_input.settings.evm_version = Some(EvmVersion::Shanghai);
 
+    if let Ok(version) = sc.version() {
+        if version.cmp(&SHANGHAI_SOLC).is_lt() {
```
JamesHinshelwood, line 529: Is there any way to de-duplicate the `EvmVersion::Shanghai` above with the `SHANGHAI_SOLC` here, so we can't forget to update one if we update the other?

JamesHinshelwood: I'm just skimming the `ethers` docs. I think one (maybe a bit cruddy) way would be:
```
if let Ok(version) = sc.version() {
  let expected_version = compiler_input.settings.evm_version;
  if expected_version.normalize_version(version) != Some(expected_version) {
    panic!("...");
  }
}
```

If it works, it could probably do with an explanatory comment too, because I don't think `normalize_version` is a very well named method.

zilliqa/tests/it/main.rs
```diff
@@ -518,12 +519,21 @@ async fn deploy_contract(
     let mut contract_file = tempfile::Builder::new().suffix(".sol").tempfile().unwrap();
     std::io::Write::write_all(&mut contract_file, &contract_source).unwrap();
 
-    let sc = ethers::solc::Solc::default();
+    let sc: ethers::solc::Solc = ethers::solc::Solc::default();
```
JamesHinshelwood, line 522:
```suggestion
    let sc = ethers::solc::Solc::default();
```

zilliqa/tests/it/main.rs
WuBruno: Oh I came across that function as well, but got so confused by then name that I just glanced over it. Will update it to that instead.
I had also realised that was a problem being separate and was trying to find a way to link them together

JamesHinshelwood (approved):

PR #476, opened 2023-10-05
Mempool priority txs

n-hutton:
keep a heap for each from address of txs ordered by nonce, and in the case of same nonce, by higher gas price. This allows efficient collecting of txs into a block (although it doesn't fully optimize for highest fees in the block for simplicity - for that you would probably want another heap of all TXs). Comments inline.


zilliqa/src/consensus.rs
```diff
@@ -108,6 +111,64 @@ impl From<Hash> for MissingBlockError {
     }
 }
 
+#[derive(Debug, Clone)]
```
n-hutton, line 114: this just contains the minimal set of info rather than the full TX so we can define a custom ordering and don't have to copy the tx in to new_transactions and new_transactions_priority

JamesHinshelwood: I reckon there must be a cleaner way of representing this. I'll have a think/research and come back to this PR.

JamesHinshelwood:
@n-hutton I sketched the following for a separate mempool struct:
```rust
struct TransactionPool {
    /// All *valid* transactions in the pool, indexed by (sender, nonce)
    transactions: BTreeMap<(H160, u64), SignedTransaction>,
    /// Transaction indices sorted by gas price
    ready: BinaryHeap<(H160, u64)>,
}

impl TransactionPool {
    fn best_transaction(&mut self) -> Option<SignedTransaction> {
        let Some(idx) = self.ready.pop() else { return None; };
        let transaction = self.transactions.remove(idx).unwrap();

        // maybe a newtype for txn identifiers so you can implement `next_nonce`? Or just deconstruct and increment?
        let next_idx = idx.next_nonce();
        if self.transactions.get(next_idx).is_some() {
            self.ready.insert(next_idx);
        }

        return Some(transaction);
    }

    fn insert_transaction(&mut self, transaction: SignedTransaction) {
        // check chain ID, gas price, etc. are valid
        // check account nonce is *strictly less than* txn nonce
        // if okay, add to pool
    }
}
```

I think the main difference is that we order transactions in the heap by gas price and figure out nonce ordering between accounts by checking for inclusion in a separate map instead.

Does this make sense?

zilliqa/src/consensus.rs
```diff
@@ -108,6 +111,64 @@ impl From<Hash> for MissingBlockError {
...
+struct TxnOrder {
+    pub nonce: u64,
+    pub gas_price: u64,
+    pub gas_limit: u64,
+    pub hash: Hash,
+    pub retries: Arc<Mutex<u64>>, // so we can modify the element in a heap
```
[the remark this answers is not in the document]
n-hutton: The reason for the arc mutex was because I was iterating over the heap and increasing retries, the iterator is not mutable.

However, I later realised that iterating over the heap isn't actually ordered, so it has to be a pop, therefore it doesn't actually need arc mutex or AtomicU64 - my latest commit switches it to u64.

zilliqa/src/consensus.rs
```diff
@@ -123,8 +184,8 @@ pub struct Consensus {
     pending_peers: Vec<Validator>,
     /// Transactions that have been broadcasted by the network, but not yet executed. Transactions will be removed from this map once they are executed.
     new_transactions: BTreeMap<Hash, SignedTransaction>,
-    /// Transactions that cannot yet be executed due to the nonce being higher. The value is the number of retries
-    new_transactions_waiting: BTreeMap<Hash, u64>,
+    /// Transactions ordered by priority, map of address of TXn (from account) to ordered TXns to be executed.
+    new_transactions_priority: BTreeMap<Address, BinaryHeap<TxnOrder>>,
```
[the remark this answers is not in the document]
n-hutton: Not sure what extra lots means? However I do agree having a seperate dedicated mempool seems a lot cleaner and better (possibly also incorporating its own logic for getting missing txs).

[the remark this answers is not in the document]
JamesHinshelwood: Sorry, `s/extra/extract`.

n-hutton:
> @n-hutton I sketched the following for a separate mempool struct:
> ...
> Does this make sense?

You also need a lookup of hash to SignedTransaction so that you can 'getHashByTransaction' right? Although, this is minor of course.

Would collisions of (H160, u64) replace the SignedTransaction (priority tx)?

As I understand the logic of this, it is something like, in a loop, check/pop the heap for the most lucrative TX (which contains all TXs?). Then you can get it from the transactions map.

I don't understand how you are suggesting to deal with that now the nonce of interest for that from account is increased by 1. Also, how do you remove TXs that you see in others' blocks? Iterate over the whole heap looking for the (H160, u64)?

The ready heap seems problematic to me - if I understand it correctly, you are not guaranteed that what you pop off `ready` is valid as it might have the highest gas fee (heap ordering), but a nonce that is too high. It might also have a nonce that is too low, as time can pass between you adding into the `ready` heap and attempting a reap.

It also seems that during the reap you modify the mempool/heap - this could be bugprone if your proposal doesn't get accepted and now your heap is balanced as if the txs got executed.

zilliqa/src/consensus.rs
```diff
@@ -472,87 +548,111 @@ impl Consensus {
...
+                        remove_all_txs = true;
+                    }
+                }
             }
 
-            // Tx appears to be good, add it to the list
-            tx_from_addrs.insert(tx.from_addr);
-            ret.push(tx.clone());
+            // Put the txs back in the heap
+            txs.extend(temp_vec.into_iter());
```
theo-zil, line 638: Should this happen before or after `remove_all_txs` is checked?

Consider the following:
1. `max_retries` is 1000. Account A has nonce 0. A sends a transaction T1000 with nonce 1001.
2. 1000 blocks later, A sends a transaction T0 with nonce 0, and T2 with nonce 2.

What I would expect: T0 gets executed, while T2 remains with the queue, with `retries = 1`.
What seems will happen: T0 gets executed. T2 is popped and placed into `temp_vec` because it has 999 retries left. T1000 is checked, it now has 1000 retries, so it's dropped and `remove_all_txs` is set to true. Then, here, `temp_vec` (including T2) is poured back into `txs`, and in the if-statement just below `txs` gets cleared, thus T2 is dropped from the mempool.

zilliqa/src/consensus.rs
```diff
@@ -472,87 +548,111 @@ impl Consensus {
...
-                warn!(
-                    "Transaction from address {} already exists in block and will not be added",
-                    tx.from_addr
-                );
-                continue;
+                // tx nonce == iter nonce, so we can add it
+                match self.new_transactions.get(&txn.hash) {
+                    Some(tx) => {
+                        trace!("Adding tx to execute: {:?}", tx);
+                        ret.push(tx.clone());
```
theo-zil, line 624: Note that by now, if I'm reading this correctly, we will have already pushed `txn` into `temp_vec` (L587). Thus at the same time as we're adding `tx` to the ret vec and executing it this block, we're also putting the hash back into the priority queue below (L638). This doesn't seem right - but let me know if I've misunderstood how this works.

zilliqa/src/consensus.rs
```diff
@@ -472,87 +548,111 @@ impl Consensus {
...
+                        iter_nonce += 1;
+                    }
+                    None => {
+                        warn!(
+                            "Transaction {} not found in mempool, but was in priority txs",
+                            txn.hash
+                        );
+                        remove_all_txs = true;
```
theo-zil, line 632: What's the reason for clearing out the entire priority map if one inconsistent header is found? Having the data be inconsistent sounds like it would be symptom of a bug - I'd probably flag it as an error condition, then either ignore it hoping it's a one off or rebuild the entire heap by scanning `new_transactions`. I'm a bit confused at the choice here

zilliqa/tests/it/eth.rs
```diff
@@ -574,3 +574,166 @@ async fn eth_call(mut network: Network) {
...
+
+    // Send all of them
+    for tx in txs_to_send {
+        let prom = wallet.send_transaction(tx, None);
+        promises.push(prom);
+    }
+
+    // Wait for all of them to be completed
+    for prom in promises {
+        let _hash = prom.await.unwrap().tx_hash();
```
theo-zil, line 650: I think this is gonna execute them sequentially, and using `futures::join_all` might make the test more streamlined

zilliqa/tests/it/eth.rs
```diff
@@ -574,3 +574,166 @@ async fn eth_call(mut network: Network) {
...
+    let to: H160 = "0x00000000000000000000000000000000deadbeef"
+        .parse()
+        .unwrap();
+
+    let mut txs_to_send: Vec<TypedTransaction> = Vec::new();
+    let tx_send_amount = 10;
+    let tx_send_iterations = 10;
+
+    // collect up a bunch of TXs to send at once, with two per nonce (one with a priority fee)
+    // but not including the first one to allow the mempool time to see them all
```
theo-zil, line 682: Minor nit but considering we're sending pairs of transactions, "not including the first one" confused me at first - consider
```suggestion
    // but starting from nonce 1 to allow the mempool time to see them all without being able to mine them yet
```

theo-zil (changes requested):
Most of it looks good but would just like to clarify some questions about the logic in the mempool that I wasn't sure about

zilliqa/src/consensus.rs
```diff
@@ -108,6 +109,64 @@ impl From<Hash> for MissingBlockError {
...
+        if self.hash == other.hash {
+            warn!("TXnOrder hash collision {:?} vs {:?}", self, other);
+            return Ordering::Equal;
+        }
+
+        let nonce_ordering = other.nonce.cmp(&self.nonce);
+        if nonce_ordering == Ordering::Equal {
+            // If nonce is equal, compare by gas_price * fee (desire higher)
+            let self_fee = self.gas_price * self.gas_limit;
+            let other_fee = other.gas_price * other.gas_limit;
```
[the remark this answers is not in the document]
n-hutton: I guess in my mind if you order on gas price only, then txs that have a limit of 0 or very low can be valued over other TXs with a lower price. IMO you want to work out the max amount you could be getting from the tx.

We do not currently limit by block total gas limit, just by number of TXs - should be a todo, then?

zilliqa/src/consensus.rs
```diff
@@ -472,87 +548,111 @@ impl Consensus {
...
+            // Push as many txs as have a sequential nonce into the map
+            // We have to pop, as iterating over a binary heap is not in order.
+            // So we put it all back afterward
+            let mut temp_vec = Vec::new();
+            while let Some(mut txn) = txs.pop() {
+                txn.retries += 1;
+                temp_vec.push(txn.clone());
+
+                // if we have reached the block limit, break
+                if ret.len() >= self.config.consensus.block_tx_limit {
```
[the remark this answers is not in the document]
n-hutton: I agree with all of this comment - shall I create a ticket for implementing block limits?
-gas
-number of TXs
-total block serialized size

zilliqa/src/consensus.rs
n-hutton: Hi, that was my intention. Hashes are removed later when the block is executed 'for real' and not just during construction. 

In my mind, it is better to keep the heap as-is for the reap, because you are not guaranteed your block will be executed after you construct it (you can time out or other stuff). If this happens, the TXs dissapear essentially.

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
theo-zil: Yeah, specifically two things even
 - gas limit
 - change TX selection to have the execution loop pull in new TXs as soon as the last has finished executing, (constrained by the gas limit remaining in the block at the current stage), as opposed to our current two-stage process of pre-selecting transactions then executing all of them.

+1 for ticket

zilliqa/src/consensus.rs
n-hutton: Yes, that is a good point - ok, I will just drop that TX from temp_vec and have the warning

zilliqa/tests/it/eth.rs
n-hutton: It doesn't make the test run any faster, but it is cleaner 👍

zilliqa/src/consensus.rs
theo-zil: In my example there is no T1 though - T2 should remain in the queue for another 999 blocks (or until a T1 is broadcast later). The code seems like it will eject it immediately though, just because T1000 was ejected.

n-hutton: Oh right, I misread, sorry about that. However, I still do not think it will happen as you say. 

There is a break whenever the TX nonce is too high, so what will happen is that only the lowest nonced TX has its retries incremented. That is to say, the flow would be: when T2 arrives, its retries will be incremented until == 1000, T1000 will not be touched and will stay on 999.

There is no point checking higher nonces, and implicitly if you have run out of retries for the lower one, the higher ones are forefit.

theo-zil: Oh I see! Yeah I glossed over that `break` - I agree that it resolves my concern, and everything looks correct. Thanks for clearing it up for me.

theo-zil (approved):
Approving, James's comments notwithstanding - happy to merge this now if the work to separate the mempool into its own struct is gonna be in a subsequent PR

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
n-hutton: https://github.com/Zilliqa/zq2/issues/496

theo-zil (approved):

PR #486, opened 2023-10-12
Split message channels

theo-zil:
Separates the `InternalMessage` and `ExternalMessage` structs to be sent through different channels, and removes the combined Message enum.

The purpose of this is to improve the type safety of sending messages around. `InternalMessage` is intended for messages within the same running process - from a shard thread to the p2p coordinator process, or (in the future) between shard threads for cross-shard messaging. `ExternalMessage` meanwhile is the type for actual p2p messages which are sent over the network.
Without this PR, both types go through the same tokio `channel`s, and all channels accept a common `Message` type that encapsulates both of the above. This means that while we can handle messages in a type-safe manner with pattern matching, the actual communication channel does not enforce any safety.

This change splits the channels, so one dedicated `unbounded_channel()` is typed to only route `ExternalMessage`s and another is typed only for `InternalMessage`s. Each gets correspondibly separate handling in the message loop of `p2p_node`. This ensures conceptual separation, allows the functions to handle the different messages to be separate and avoids having to nest match statement in every place a message is being handled.

The tradeoff for ensuring cleanly separated channels for each type is of course that we have to hold more senders and receivers at each end. This adds a bit more boilerplate - the most significant place for this is in the test scaffolding, where I actually re-create a joint enum and redirect both types of messages into the same channel for processing in the `tick()` loop. This turned out to be more bothersome than I initially expected, so I think there's a genuine tradeoff in readability and complexity here, where simply having a union enum and a single channel for everything in the first place avoids it - so I'm not entirely convinced this is worth merging after all. Opening PR for discussion, though.

### DIFF
--- a/zilliqa/src/node_launcher.rs
+++ b/zilliqa/src/node_launcher.rs
@@ -1,3 +1,4 @@
+use crate::p2p_node::LocalMessageTuple;
 use crate::{health::HealthLayer, message::ExternalMessage};
 use jsonrpsee::RpcModule;
 use std::{
@@ -12,7 +13,7 @@ use anyhow::{anyhow, Result};
 use http::{header, Method};
 use libp2p::{futures::StreamExt, PeerId};
 
-use crate::message::Message;
+use crate::message::InternalMessage;
 use node::Node;
 use std::time::Duration;
 use tokio::{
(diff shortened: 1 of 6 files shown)


zilliqa/src/node_launcher.rs
```diff
@@ -28,9 +29,19 @@ pub struct NodeLauncher {
...
-    pub inbound_message_sender: UnboundedSender<(PeerId, Message)>,
-    pub inbound_message_receiver: UnboundedReceiverStream<(PeerId, Message)>,
+    /// The following three message streams are used for networked messages.
+    /// The sender is provided to the p2p coordinator, to forward messages to the node.
+    pub inbound_message_sender: UnboundedSender<(PeerId, ExternalMessage)>,
+    /// The corresponding receiver is handled here, forwarding messages to the node struct.
+    pub inbound_message_receiver: UnboundedReceiverStream<(PeerId, ExternalMessage)>,
+    /// This sender is provided to the node's MessageSender, to send messages back to the network.
     outbound_message_sender: UnboundedSender<OutboundMessageTuple>,
+    /// The following three message streams are used for local messages.
```
JamesHinshelwood, line 39:
```suggestion
    /// The following two message streams are used for local messages.
```
?

JamesHinshelwood (dismissed):
I like this change :) I think it makes the intent of messages clearer and thus more readable and I don't mind the extra boilerplate.

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):

PR #487, opened 2023-10-13
rename tests

n-hutton:
It seems preferable to me to not have duplicated filenames (consensus.rs in two places etc) - wdyt?


JamesHinshelwood:
IMO the `test_` prefix is unnecessary because they're already under the `test/` directory. In VSCode I use the "Quick open" (CTRL+P) dialog which lets me fuzzy search for files. So if I wanted to go to the consensus tests, I would type CTRL+P, then "test consensus" and press enter.

PR #488, opened 2023-10-15
Consensus improvements and tests

n-hutton:
Consensus improvements:
- Single node can produce blocks
- Anyone can join and sync from 0
- Anyone can die and rejoin from persistence (including whole network dying)
- Generally a lot more resillient

todos:
- unhappy path, aggQC has not been implemented
- do validators want to select leader by weight, too?
- Block sync is incredibly slow. Presumably, if you get a vec of incrementing blocks you can do something like verifying only the first and last signatures

Comments are inline. Block number is implemented and used over views as views are not guaranteed to be contigous.

Block sync is implemented by saying 'my head block number is N, give me whatever you have above that' - it's peers' discretion how many blocks to respond with.

Block sync is only triggered during timeouts when it is likely you are behind and continues until you get no more blocks.

Getting who the committee is, is achieved by following the QC hash since there were some bugs where the block proposal had a different committee than was believed (leading to invalid signature errors).

Added and tested (manually) exponential view timeout operation, does work.


zilliqa/src/api/erigon.rs
```diff
@@ -18,7 +18,7 @@ fn get_header_by_number(params: Params, node: &Arc<Mutex<Node>>) -> Result<Optio
     let header = node
         .lock()
         .unwrap()
-        .get_block_by_view(block)?
+        .get_block_by_number(block)?
```
n-hutton, line 21: view should not be used outside of consensus.rs as we index blocks by number now

zilliqa/src/api/eth.rs
```diff
@@ -68,11 +68,12 @@ fn accounts(_: Params, _: &Arc<Mutex<Node>>) -> Result<[(); 0]> {
 }
 
 fn block_number(_: Params, node: &Arc<Mutex<Node>>) -> Result<String> {
```
n-hutton, line 70: No clue why this was -= 1 before

zilliqa/src/api/eth.rs
```diff
@@ -264,8 +265,8 @@ fn get_block_transaction_count_by_number(
 
     let node = node.lock().unwrap();
     let block = match block_number {
-        BlockNumber::Number(number) => node.get_block_by_view(number),
-        BlockNumber::Earliest => node.get_block_by_view(0),
+        BlockNumber::Number(number) => node.get_block_by_number(number),
```
n-hutton, line 268: I would argue we could get rid of getting blocks by view network call, and just use number, where if there are forks you could get different results  for multiple queries of number.

zilliqa/src/consensus.rs
```diff
@@ -113,12 +116,13 @@ pub struct Consensus {
     secret_key: SecretKey,
     config: NodeConfig,
     message_sender: MessageSender,
-    block_store: BlockStore,
+    pub block_store: BlockStore,
```
n-hutton, line 119: if you make block store public, you can avoid having consensus having duplicated methods to access it

zilliqa/src/consensus.rs
```diff
@@ -325,28 +341,126 @@ impl Consensus {
         Ok(None)
     }
 
-    fn download_blocks_up_to(&mut self, to: u64) -> Result<()> {
-        for view in (self.view + 1)..to {
-            self.view = view;
-            self.block_store.request_block_by_view(view)?;
-        }
-        self.view = to + 1;
+    pub fn head_block(&self) -> Block {
```
n-hutton, line 344: Head block is now prioritized over finalized block for use. 

I think asking the DB here for the highest block number will become cleaner with james' DB pr

zilliqa/src/consensus.rs
```diff
@@ -366,14 +484,7 @@ impl Consensus {
...
                 if let Some(e) = e.downcast_ref::<MissingBlockError>() {
-                    trace!(?e, "missing block");
-                    match e.0 {
-                        BlockRef::Hash(hash) => self.block_store.request_block(hash)?,
-                        BlockRef::View(view) => self.block_store.request_block_by_view(view)?,
-                    }
-
-                    self.add_block(block)?;
-
+                    info!(?e, "missing block when checking block proposal");
```
n-hutton, line 487: IMO the network should send as few messages as possible. It should not request blocks etc when it detects a floating block, as this could cause network cascades (for example, when you start up, you will get a bunch of blocks from peers that will trigger this logic, but in reality you want to just patiently sync the chain)

zilliqa/src/consensus.rs
```diff
@@ -636,13 +778,13 @@ impl Consensus {
...
                 supermajority_reached,
+                total_weight = block.committee.total_weight(),
                 self.view,
                 vote_view = block_view + 1,
                 "storing vote"
             );
             if supermajority_reached {
-                self.block_store.request_block_by_view(block_view)?;
-                self.download_blocks_up_to(block_view)?;
+                self.block_store.request_block_by_view(block_view)?; // Is this required?
```
n-hutton, line 787: Not sure why this line was originally here?

zilliqa/src/consensus.rs
```diff
@@ -896,26 +1126,36 @@ impl Consensus {
...
         let Some(new_high_qc_block) = self.block_store.get_block(new_high_qc.block_hash)? else {
             // We don't set high_qc to a qc if we don't have its block.
+            warn!("Recieved potential high QC but didn't have the corresponding block");
             return Ok(());
         };
 
         let new_high_qc_block_view = new_high_qc_block.view();
 
         if self.high_qc.block_hash == Hash::ZERO {
+            // This seems like a potential bug???
```
n-hutton, line 1136: can people not send you a block with hash zero and you end up dumping your highQC?

zilliqa/src/consensus.rs
```diff
@@ -952,6 +1192,7 @@ impl Consensus {
 
     fn block_extends_from(&self, block: &Block, ancestor: &Block) -> Result<bool> {
         // todo: the block extends from another block through a chain of parent hashes and not qcs
+        // make ticket for this
```
n-hutton, line 1195: note: will create a single consensus ticket with all todos

zilliqa/src/consensus.rs
```diff
@@ -1128,59 +1380,69 @@ impl Consensus {
...
-        // This block's timestamp must be at most `self.allowed_timestamp_skew` away from the current time. Note this
+        // This block's timestamp should be at most `self.allowed_timestamp_skew` away from the current time. Note this
         // can be either forwards or backwards in time.
+        // Genesis is an exception for now since the timestamp can differ across nodes
         let difference = block
             .timestamp()
             .elapsed()
             .unwrap_or_else(|err| err.duration());
-        if difference > self.config.allowed_timestamp_skew {
+        if difference > self.config.allowed_timestamp_skew && parent.view() > 0 {
```
n-hutton, line 1402: IMO, the timestamp logic should just be 'is it more than the previous timestamp' - and maybe if it is a block proposal, is it less than now.

The problem you were seeing @JamesHinshelwood  about timestamp skew is that if you sync the chain, then blocks you check to be 'safe' fail this check since they were minted ages ago.

zilliqa/src/consensus.rs
```diff
@@ -1294,39 +1552,20 @@ impl Consensus {
...
-        let block = match block {
-            Some(block) => block,
-            None => self
-                .get_block_by_view(0)?
-                .ok_or_else(|| anyhow!("missing genesis block!"))?,
-        };
-
-        Ok(block.committee.leader(view))
+        // TODO: Implement batch verification - this will not work atm.
+        //verify_messages(agg.signature, &messages, &public_keys)
```
n-hutton, line 1564: aggregate QC would fail here, so it is commented out so that the unhappy path can be tested/used

zilliqa/src/message.rs
```diff
@@ -454,8 +474,10 @@ impl Block {
             .flat_map(|v| v.public_key.as_bytes())
             .collect();
 
+        // Could the hash be all zeroes for genesis perhaps?
```
n-hutton, line 477: I guess this might not work too well with shards in some way(?)

zilliqa/src/node.rs
```diff
@@ -166,17 +200,15 @@ impl Node {
...
-    pub fn handle_timeout(&mut self) -> Result<()> {
-        match self.consensus.timeout() {
-            Ok((leader, new_view)) => {
-                self.message_sender
-                    .send_external_message(leader, ExternalMessage::NewView(Box::new(new_view)))?;
-            }
-            Err(_) => {
-                warn!("timeout failed");
-            }
+    // handle timeout - true if something happened
```
n-hutton, line 203: the timeouts are changed - the node launcher just pings consensus every time there aren't messages for a little while. It is up to consensus to decide whether to send a message (newView) given that it knows what time the last timeout was.

This moves the consensus logic more into consensus which seems correct, and tests are able to ping the timeout until something happens, when testing network failures.

zilliqa/src/node.rs
```diff
@@ -379,7 +421,8 @@ impl Node {
     fn handle_block_request(&mut self, source: PeerId, request: BlockRequest) -> Result<()> {
         let block = match request.0 {
             crate::message::BlockRef::Hash(hash) => self.consensus.get_block(&hash),
-            crate::message::BlockRef::View(view) => self.consensus.get_block_by_view(view),
+            crate::message::BlockRef::View(view) => self.consensus.get_block_by_view(view), // todo: consider removing
```
n-hutton, line 424: I am not sure we need to be able to reference blocks by view externally

zilliqa/src/node.rs
```diff
@@ -399,4 +442,68 @@ impl Node {
...
+            "Received blocks response of length {}",
+            response.blocks.len()
+        );
+        let mut was_new = false;
+
+        for block in response.blocks {
+            was_new = self.consensus.receive_block(block)?;
+        }
+
+        if was_new {
```
n-hutton, line 499: effectively, every time you verify that you got new blocks when syncing from a peer, ask for more with your new head as the tip. This way it recurses until you should have them all.

zilliqa/src/p2p_node.rs
```diff
@@ -260,12 +259,6 @@ impl P2pNode {
 
                     let topic = Self::shard_id_to_topic(shard_id);
 
-                    // Push messages back into queue if there are no peers
```
n-hutton, line 263: this interferes with broadcasts in consensus that you want to also recieve when you are the only node in the network. So, take it or leave it

zilliqa/tests/it/main.rs
```diff
@@ -206,21 +217,23 @@ impl Network {
         // Pause time so we can control it.
         zilliqa::time::pause_at_epoch();
 
-        for node in &nodes[1..] {
```
n-hutton, line 209: nodes will now try to join when they see proposed blocks they are not in the committee for

zilliqa/src/consensus.rs
theo-zil: I think timestamp skew should prevent blocks being mined too far into the future (to allow for e.g. ~1-2s skew, at most, between system clocks in nodes), but I agree that it shouldn't restrict blocks in the past as long as they're later than their parent block.

zilliqa/src/message.rs
theo-zil: It used to be all zeroes but was changed to be the real hash to avoid special case handling - iirc it was done first so node joining and block syncing are simpler, but it's very useful for shards too.
And conceptually, I do think it's cleaner for genesis to have a proper hash just like any block.

zilliqa/src/node.rs
theo-zil: Nice!

zilliqa/src/node.rs
```diff
@@ -208,11 +240,12 @@ impl Node {
         Ok(hash)
     }
 
-    pub fn view(&self) -> u64 {
-        self.consensus.view()
+    pub fn number(&self) -> u64 {
```
theo-zil, line 243: This seems redundant with `get_chain_tip()` below. I wonder, is there a use for exposing an API to check the node's consensus view, i.e. leaving this function unchanged?
If not, then one of either this or `get_chain_tip()` should probably be deleted.

zilliqa/src/message.rs
```diff
@@ -116,10 +116,17 @@ impl NewView {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct BlockRequest(pub BlockRef);
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BlocksRequest(pub BlockRef);
```
theo-zil, line 120: I would rename this to something like `BlockBatchRequest` - just BlocksRequest looks confusingly similar to BlockRequest, and also at first made me think it would take some parameters specifying the range of blocks as opposed to just a predetermined batch.

Also maybe a doc comment specifying that it's a batch of 100

theo-zil (commented):
Looking forward to getting this in!

In think re: your comment regarding whether we need to expose `get_block_by_view` - I honestly don't think we need that, and I'm not sure we even need to save the `block.view`. Off the top of my head I haven't checked if we need it for consensus, but if not, I'd drop it entirely, and all the other related functions.
Even if we do, I think `canonical_block_views` is likely not necessary.

zilliqa/src/consensus.rs
```diff
@@ -167,17 +168,48 @@ impl PartialOrd for TxnOrder {
...
+/// - after threshold signatures are aggregated, a QC is formed which points to the block proposal
+///
+/// Happy path:
+/// - Start at genesis, there is only a block with a dummy QC which everyone sees (exceptional case).
+/// - everyone advances view to 1
+/// - validators vote on genesis
+/// - a high QC (QC pointing to the highest known hash) is formed from the validators votes on genesis
+/// - everyone advances view to 2
+/// - next leader proposes a block
+/// - validators vote on genesis -> new high QC... and so on.
```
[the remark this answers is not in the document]
n-hutton: I will change, but it is technically it is block 0 :)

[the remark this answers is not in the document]
n-hutton: Oh, I misread, yup well caught

zilliqa/src/consensus.rs
```diff
@@ -384,28 +421,127 @@ impl Consensus {
...
+        // Now consider whether we want to timeout - the timeout duration doubles every time, so it
+        // Should eventually have all nodes on the same view
+        let consensus_timeout_ms = self.config.consensus.consensus_timeout.as_millis() as u64;
+        let head_block = self.head_block();
+        let head_block_view = head_block.view();
+        trace!("Time since last change is: {:?}", self.last_timeout);
+        println!("Time since last change is: {:?}", self.last_timeout);
+        let time_since_last_view_change = SystemTime::now()
+            .duration_since(self.last_timeout)
+            .expect("last timeout seems to be in the future...")
```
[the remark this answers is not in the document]
n-hutton: Yes, great suggestions - how about `time_since_last_view_change` just defaults to 0s with a warn! if it fails? That way we don't touch/potentially race on `last_timeout`

zilliqa/src/consensus.rs
```diff
@@ -384,28 +421,127 @@ impl Consensus {
...
+    pub fn download_blocks_up_to_head(&mut self) -> Result<()> {
+        let head_block = self.head_block();
+        self.block_store.request_blocks(head_block.header.number)?;
 
         Ok(())
     }
 
-    pub fn timeout(&mut self) -> Result<(PeerId, NewView)> {
-        self.view += 1;
+    // Set view only here - that way the last_timeout should be set iff view is set
```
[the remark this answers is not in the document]
n-hutton: Yes, I had a similar thought and should have implemented it really!

zilliqa/src/consensus.rs
```diff
@@ -723,10 +880,17 @@ impl Consensus {
...
+                "(vote) supermajority already reached in this round {}",
+                self.view
+            );
             return Ok(None);
         }
 
+        if index > 5 {
+            panic!("we don't expect more than 5 validators currently");
+        }
+
```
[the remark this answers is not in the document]
n-hutton: It was just debug I guess, I will remove it

zilliqa/src/consensus.rs
```diff
@@ -802,38 +968,103 @@ impl Consensus {
...
+            info!(
+                "*** adding {} pending peers to committee",
+                self.pending_peers.len()
+            );
+        }
+
         committee.add_validators(self.pending_peers.drain(..));
-        Ok(committee)
+
+        if committee.len() <= 1 {
```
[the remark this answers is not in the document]
n-hutton: The check will still print when the length is 1 - this will happen if for example other nodes in the network can never catch up and so it's just the one node producing blocks (I think this is always undesirable outside of tests and bootstrapping the network).

zilliqa/src/consensus.rs
```diff
@@ -802,38 +968,103 @@ impl Consensus {
...
+            _ => {
+                warn!("parent not found during leader for view {}", view);
+                return None;
+            }
+        };
+
+        Some(parent.committee.leader(view).peer_id)
+    }
+
+    fn committee_for_hash(&mut self, parent_hash: Hash) -> Result<Committee> {
```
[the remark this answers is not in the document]
n-hutton: I would say that yes it is normal. Someone can send you a view in the future and your attempt to look it up can give nothing useful. Committee for hash is only really used to find the committee in the parent of the block you are evaluating, which is basically guaranteed to exist

zilliqa/src/message.rs
n-hutton: I will rename - but I would not specify 100 in a doc comment as it is up to the sending node and so could really be any number.

zilliqa-macros/src/test.rs
```diff
@@ -27,7 +27,7 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
                 vec![seed.to_str().unwrap().parse().unwrap()]
             } else {
                 let samples: usize = std::env::var_os("ZQ_TEST_SAMPLES")
-                    .map(|s| s.to_str().unwrap().parse().unwrap())
+                    .map(|s| s.to_str().unwrap_or("1").parse().unwrap_or(1))
```
n-hutton, line 30: this threw an error if it couldn't parse the string (trailing whitespace possibly), but you can't tell where in the macro it came from

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
theo-zil: My thinking was that if you adjust the system clock it might invalidate `last_timeout`. Say a machine was running ahead by 5-10 minutes (e.g. NTP misconfigured/not running/whatever), then the system clock gets abruptly fixed to the correct time: you would then wait 5-10 minutes for your next timeout. Whereas if you reset `last_timeout`, you will only wait up to, at the very most, double the normal timeout length.

However machines with such poorly configured system time probably can't properly run the chain anyway and don't need to be supported, so on second thought your more conservative solution of doing nothing (defaulting `time_since_last_view_change` to 0 and hoping the problem goes away eventually) might be just fine.

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
theo-zil: Makes sense, right.

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
theo-zil: Good points

zilliqa-macros/src/test.rs
theo-zil: A bit odd (trailing whitespace should never be part of an env variable in most shells... I would assume at least), also the same issue would probably happen with `ZQ_TEST_RNG_SEED`.

What context were you encountering the error in? I'd argue that it might be better to fail early, especially e.g. in CI we definitely don't want a parse error to be silently converted to 1 and thus believe we're running lots of tests when actually defaulting to a single sample.

zilliqa/tests/it/consensus.rs
```diff
@@ -1,11 +1,138 @@
...
+    network.restart();
+
+    // Panic if it can't progress to the target block
+    network
+        .run_until(
+            |n| {
+                let index = n.random_index();
+                n.get_node(index).get_finalized_height() >= finish_block
+            },
+            50000,
```
theo-zil, line 89: This surely could use a more reasonable timeout for network startup - maybe 1000 or even 500?

zilliqa/tests/it/main.rs
```diff
@@ -278,6 +277,110 @@ impl Network {
...
+        self.resend_message = resend_message;
+
+        // Now trigger a timeout in all of the nodes until we see network activity again
+        // this could of course spin forever, but the test itself should time out.
+        loop {
+            for node in &self.nodes {
+                if node.inner.lock().unwrap().handle_timeout() {
+                    return;
+                }
+                zilliqa::time::advance(Duration::from_millis(1));
```
theo-zil, line 352: This uses the normal timeout logic that defaults to 5 seconds, right? If so, maybe advancing more than 1ms at a time would still work fine without compromising the test, and reduce the busy spinloop a bit (instead of requiring 5000 iterations every time)?

zilliqa/tests/it/main.rs
```diff
@@ -278,6 +277,110 @@ impl Network {
...
+            }
+        }
+    }
+
+    // Drop the first message in each node queue with N% probability per tick
+    pub async fn randomly_drop_messages_then_tick(&mut self, failure_rate: f64) {
```
theo-zil, line 358: If block production still works with dropped messages, this makes me wonder whether we can just integrate this logic into `tick()` and run _all_ our tests on an unreliable network (rather than just having this special case method)?

Worth a try I think (possibly in a different PR though if you think it'll delay this one).

zilliqa/tests/it/consensus.rs
```diff
@@ -1,11 +1,138 @@
...
+                let index = n.random_index();
+                n.get_node(index).get_finalized_height() >= start_block
+            },
+            50,
+        )
+        .await
+        .unwrap();
+
+    // now, wait until block 15 has been produced, but dropping 10% of the messages.
+    for _i in 0..1000000 {
```
theo-zil, line 122:
```suggestion
    for _ in 0..1000000 {
```

Also, would a more reasonable timeout make sense here? If I'm reading this correctly the condition is just that the network advances a single block, I think a million messages is quite unreasonable even under unreliable conditions. I'd expect it should probably work within 50-100 messages comfortably, given that it works at all in the first place.

zilliqa/tests/it/consensus.rs
```diff
@@ -1,11 +1,138 @@
 use crate::CombinedJson;
+use crate::Network;
 use ethabi::Token;
-
+use ethers::{providers::Middleware, types::TransactionRequest};
+use tracing::*;
 use zilliqa::state::Address;
 
-use crate::Network;
+// Test that a node that joins later can sync up to the latest block and become a
```
theo-zil, line 8: These commented out blocks should probably be cleaned up before merging (including the entire commented test)

zilliqa/src/consensus.rs
```diff
@@ -1213,87 +1521,118 @@ impl Consensus {
...
-                    match e.0 {
-                        BlockRef::Hash(hash) => self.block_store.request_block(hash),
-                        BlockRef::View(view) => self.block_store.request_block_by_view(view),
-                    }?
-                } else {
-                    return Err(e);
-                }
+                //warn!("Error when receiving block {}", e);
+                warn!(?e, "invalid block received during sync!");
+                //return Err(e);
```
theo-zil, line 1630: (dev comments?)

zilliqa/src/consensus.rs
```diff
@@ -1065,13 +1362,18 @@ impl Consensus {
...
         let mut current = block.clone();
         while current.view() > ancestor.view() {
             let Some(next) = self.get_block(&current.parent_hash())? else {
+                warn!("Missing block when traversing to find ancestor! Current parent hash: {:?} {:?}", current.parent_hash(), current);
                 return Err(MissingBlockError::from(current.parent_hash()).into());
             };
             current = next;
         }
+
+        //warn!("Completing, but with hashes: {} and {} and view {} and view {}", current.hash(), ancestor.hash(), current.view(), ancestor.view());
```
theo-zil, line 1375: dev comment (or unintentionally disabled)

zilliqa-macros/src/test.rs
n-hutton: Yes, that is fair. The reason that I defaulted it to 1 was because the original code is:

```
                let samples: usize = std::env::var_os("ZQ_TEST_SAMPLES")
                    .map(|s| s.to_str().unwrap().parse().unwrap())
                    .unwrap_or(1);
```

And I just followed its intent - to default to 1. But I do agree, it should panic in a legible way if it cannot read it, I will change it.

The context is that my IDE (misconfigured by me) set this to a non-integer.

theo-zil: Yeah, the original unwraps the option (i.e. if the variable is not specified at all), but panics on the error.

zilliqa/tests/it/consensus.rs
n-hutton: I will set it lower, but the test exits when the condition is true or at tick amount, so having the amount be high doesn't increase the test time or anything (except when it fails).

zilliqa/tests/it/main.rs
n-hutton: Yes, that's right. I was worried that if you increase the time step too much you could get into weird situations, but that's probably unwarranted. I will bump it to 500ms, seems like a good change

zilliqa/tests/it/main.rs
n-hutton: I wouldn't favour doing that because it would make debugging the tests more annoying, plus it takes significantly longer for the network to stabilize.

On the other hand if done manually could potentially reveal stuff like forks created by flaky networking not being handled correctly I guess. 

I think on balance, your test should try and test only the functionality you want it to test, and not fail for some other kindof unrelated reason (for example you change networking and some other test now sometimes fails).

zilliqa/tests/it/consensus.rs
n-hutton: yes, you're right, I'll lower it

zilliqa/src/consensus.rs
n-hutton: ah yes, good catch. 

As a note: this is one of the bugs that was fixed - when the network is restarted from scratch, it is possible for a block to be sent to you which fails the block check. Rather than making an err, just return the block was bad.

zilliqa/src/consensus.rs
n-hutton: ah yes, good catch thanks

zilliqa/tests/it/main.rs
theo-zil: That's fair. I would argue that most, if not all, tests should ultimately be testing the network under non-ideal conditions - it's why we have the randomisation in the first place, for example.
But you make a good point that it will make it much harder to debug failures.

zilliqa/src/consensus.rs
theo-zil: Nice

theo-zil (approved):

PR #498, opened 2023-10-18
Updated shard contracts

theo-zil:
_Currently blocked on consensus stalling when multiple nodes get added to a shard at the same time; I'm investigating this._

Shuffles the system contracts such that all `Shard`s are deployed on the main shard (whose security is assumed trusted). They also contain the genesis hash of their shard, thus letting at least the shard state be rooted in the main shard's security.

This also implies that shards cannot be spun up "out of thin air" anymore: first you must manually spin up a network, obtaining at least the genesis hash, and then you can register it with the main shard at which point it becomes discoverable to all.


zilliqa/src/node.rs
```diff
@@ -96,6 +96,11 @@ impl Node {
         reset_timeout: UnboundedSender<()>,
     ) -> Result<Node> {
         let peer_id = secret_key.to_libp2p_keypair().public().to_peer_id();
+        println!(
+            "NODE SEKRET KEY: {}, PeerID: {}",
+            secret_key.to_hex(),
+            peer_id
+        );
```
JamesHinshelwood, line 103: Don't forget to remove :)

zilliqa/tests/it/consensus.rs
```diff
@@ -54,27 +58,116 @@ async fn launch_shard(mut network: Network) {
         .await
         .unwrap();
 
-    let abi = include_str!("../../src/contracts/shard_registry.json");
-    let abi = serde_json::from_str::<CombinedJson>(abi)
+    let child_shard_id = 80000u64;
+    // This is necessary to maintain a supermajority once the main shard nodes join.
+    // The size can be reduced once nodes stop joining the committee before they're
+    // fully caught up.
```
JamesHinshelwood, line 64: I wonder if @n-hutton 's latest PR changes this necessity?

JamesHinshelwood (dismissed):
LGTM :)

JamesHinshelwood (approved):

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #502, opened 2023-10-20
put fmt last

n-hutton:
Put fmt check last so you can find out if your code compiles and run tests even if the formatting isn't correct


JamesHinshelwood (approved):
Maybe we should run them in parallel by running format in a different job?

PR #511, opened 2023-10-24
start z2 deterministically

n-hutton:
With the consensus improvements, z2 could not restart from persistence as it gets a new set of private keys each time. This is one way to get around that (the other would be saving the private keys to disk which seems messier).

### DIFF
--- a/z2/src/setup.rs
+++ b/z2/src/setup.rs
@@ -42,7 +42,6 @@ impl Setup {
                 r#"
                     [[nodes]]
                     {}data_dir = "{DATADIR_PREFIX}{i}"
-                    eth_chain_id = 0x8001
                     consensus.genesis_committee = [ [ "{first_key}", "{first_peer_id}" ] ]
                     consensus.genesis_accounts = [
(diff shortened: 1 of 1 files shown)


JamesHinshelwood (commented):
Hmm, I recall it was a conscious choice of @rrw-zilliqa 's to generate random keys for nodes started by `z2`, so we could be sure that the software only worked on some specific configuration of keys.

I guess that is a fairly unlikely bug, but it would be quite unfortunate if we missed such a thing :)

How much of a pain would saving the keys to disk be? If you think its not worth it, I reckon it would be fine to merge this as-is and move back to randomly generated and persisted keys later (since storing private keys on disk without telling the user is a bit of a footgun of its own).

theo-zil (approved):
Probably the conceptually ideal solution would be for `z2` to implement its own separate jury-rigged persistence for keys, which would be synced up to the normal zq2 persistence (a simple way being to shoehorn a file into the nodes' data folders, that way it gets deleted if you wipe persistence, and saved and reloaded if you keep it).

But I'm not sure that'd be worth the trouble and extra complexity of adding it so happy to merge as-is.

PR #512, opened 2023-10-24
Support type 1 and type 2 Ethereum transactions

JamesHinshelwood:
I did some fairly major refactoring to make this support easier. We now have three different transaction types, where we used to have two:

* `Transaction` - The core information of a transaction. This is now an enum over the possible transaction types, rather than a struct which tries to unify the types. I've added accessor methods to the enum to obtain the common properties of all variants.
* `SignedTransaction` - A `Transaction` plus a signature. The details of the signature depend on the tranasction type. This is also an enum over the possible transaction types.
* `RecoveredTransaction` - A `SignedTransaction` plus the transaction's `signer` and its `hash`. This can be obtained by calling `SignedTransaction::recover_signer()`. We deliberately don't implement `serde` traits for this, to avoid it being sent over the network. This ensures nodes need to validate transaction signatures independently.


JamesHinshelwood:
I should add tests for the new transaction types...

zilliqa/src/exec.rs
```diff
@@ -158,7 +158,7 @@ impl State {
         &self,
         from_addr: Address,
         to_addr: Option<Address>,
-        gas_price: u64,
+        gas_price: u64, // TODO: Make this `u128`.
```
theo-zil, line 161: Perhaps native question but is there a reason we can't just change the signature now? Does it cause a lot more cascading refactoring?

zilliqa/src/transaction.rs
```diff
@@ -0,0 +1,542 @@
...
+            (self.y_is_odd as u64) + 27
+        };
+        rlp.append(&v)
+            .append(&strip_leading_zeroes(self.r.as_slice()))
+            .append(&strip_leading_zeroes(self.s.as_slice()));
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// A [SignedTransaction] which has had the signer recovered. The transaction's hash is also calculated and cached.
```
theo-zil, line 232: Maybe mention the fact that it's intentionally not serializable/deserializable here, for the future

theo-zil (approved):
I like this change. Some of the match statements seem verbose but I think it's worth it for the type organisation we gain from it.

zilliqa/src/api/eth.rs
```diff
@@ -622,41 +750,42 @@ fn net_listening(_: Params, _: &Arc<Mutex<Node>>) -> Result<bool> {
...
-                    .unwrap()
-            )
-        );
-        assert!(signed_tx.verify().is_ok());
+        let signed_tx = parse_transaction(&transaction).unwrap();
+        let recovered_tx = signed_tx.recover_signer().unwrap();
+        let expected = RecoveredTransaction {
+            tx: SignedTransaction::Legacy {
+                tx: TxLegacy {
+                    chain_id: Some(1),
```
n-hutton, line 771: am I missing something? Why does the legacy TX have a chain id, I thought the point was that it didn't?

n-hutton (changes requested):

zilliqa/src/api/eth.rs
n-hutton: oh, is it that the TXs can handle pre and post EIP155?

zilliqa/src/api/eth.rs
```diff
@@ -537,38 +560,143 @@ fn send_raw_transaction(params: Params, node: &Arc<Mutex<Node>>) -> Result<Strin
...
-    };
-
-    SignedTransaction::new(transaction, signing_info)
+fn parse_transaction(bytes: &[u8]) -> Result<SignedTransaction> {
+    // https://eips.ethereum.org/EIPS/eip-2718#backwards-compatibility
+    // "Clients can differentiate between the legacy transactions and typed transactions by looking at the first byte.
+    // If it starts with a value in the range [0, 0x7f] then it is a new transaction type, if it starts with a value in
+    // the range [0xc0, 0xfe] then it is a legacy transaction type."
+    match bytes[0] {
+        0xc0..=0xfe => {
```
n-hutton, line 582: seems like each of these could be split into its own function for readability?

zilliqa/src/api/eth.rs
n-hutton: ah yes, I see my folly. Legacy refers to both. Although, in all of our tests aren't we only using legacy, anyway?

The coverage of this test seems limited - doesn't cover txs with no chain id for example? I have found in the past that you can get edge cases with padding where most but not all txs get verified, just a thought.

zilliqa/src/consensus.rs
```diff
@@ -918,32 +916,34 @@ impl Consensus {
     }
 
     pub fn new_transaction(&mut self, txn: SignedTransaction) -> Result<()> {
+        let txn = txn.recover_signer()?;
```
n-hutton, line 919: Suggest change to `let verified_tx` or something, as it is not obvious on first reading that this changes the type of txn here.

zilliqa/src/exec.rs
JamesHinshelwood: Thanks for spotting, it was trivial to change this :)

zilliqa/tests/it/eth.rs
```diff
@@ -493,22 +498,128 @@ async fn get_storage_at(mut network: Network) {
...
+    assert_eq!(receipt.to.unwrap(), to);
+}
+
+/// Test which sends a legacy transaction, without the replay protection specified by EIP-155.
+#[zilliqa_macros::test]
+async fn send_legacy_transaction_without_chain_id(mut network: Network) {
+    let wallet = network.genesis_wallet().await;
+
+    let tx = TransactionRequest::pay(
+        H160::random_using(network.rng.lock().unwrap().deref_mut()),
```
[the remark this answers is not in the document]
JamesHinshelwood: Yeah. I thought it was particularly important here, rather than using a fixed address, so that we calculate and verify the expected hashes of multiple variants of each transaction type.

zilliqa/src/api/eth.rs
JamesHinshelwood: Yep, what you said :)

I just including this unit test here, because the test case was included in EIP-155. Legacy transactions without a chain ID should be tested by `send_legacy_transaction_without_chain_id` in `zilliqa/tests/it/eth.rs`.

zilliqa/src/api/eth.rs
JamesHinshelwood: Good idea, done

zilliqa/src/consensus.rs
```diff
@@ -918,32 +916,34 @@ impl Consensus {
...
         if let Some(txn) = self.new_transactions.get(&hash) {
             return Ok(Some(txn.clone()));
         }
 
         self.transactions
             .get(hash.0)?
-            .map(|encoded| Ok(bincode::deserialize::<SignedTransaction>(&encoded)?))
+            .map(|encoded| bincode::deserialize::<SignedTransaction>(&encoded))
+            .transpose()?
+            .map(|tx| tx.recover_signer())
```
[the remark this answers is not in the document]
JamesHinshelwood: Will resolve this and reply on the PR :)

zilliqa/src/consensus.rs
JamesHinshelwood: To be honest, I prefer it when we shadow the previous variable once there is no longer a use for it (and indeed we can't use it after this point because `.recover_signer()` takes `self`).

@theo-zil for a tiebreaker? :stuck_out_tongue:

JamesHinshelwood:
@n-hutton 

> -It seems odd on reading that you deseralize from new_transactions as a different type (and also wasted compute).
Sorry, but I don't understand what you mean by "deserialize from new_transactions as a different type" or why it wastes compute?
If you mean that `new_transactions` now holds a `RecoveredTransaction`, rather than a `SignedTransaction`, I guess that's saying that we recover the signer once when we first see and persist the transaction and keep that signer in persistence. It is necessary to recover the signer at this point so you can track the signer's nonce for the mempool logic.

> -Make RecoveredTransaction just be called Transaction, as it is 99% of the time what you want to be working with internally as you need the 'from' information. If not that, then consider a rename to FullTransaction maybe? To my eye, recovered is slightly confusing.
I guess I'd say I like `Recovered` because it implies to a programmer that a non-trivial operation was performed on the raw transaction in order to get it into this state. Maybe we need @theo-zil for another tie-breaker? :stuck_out_tongue_closed_eyes:

n-hutton:
> If you mean that `new_transactions` now holds a `RecoveredTransaction`, rather than a `SignedTransaction`, I guess that's saying that we recover the signer once when we first see and persist the transaction and keep that signer in persistence. It is necessary to recover the signer at this point so you can track the signer's nonce for the mempool logic.

What you have described here is how I would imagine it should go, but it doesn't seem so - for example in the function `get_transaction_by_hash` it has to perform a `tx.recover_signer()` every time, no?

Happy to get @theo-zil to tiebreak everything ofc (even though you guys gang up on me ;) )

zilliqa/src/consensus.rs
theo-zil: I honestly only have a very weak opinion here - especially with the borrow checker ensuring `txn: SignedTransaction` won't get reused no matter what, it really doesn't matter I think. Both shadowing and making the type change explicit have their merits and their drawbacks for readability.

I think if a tiebreaker is absolutely necessary I'd side with @n-hutton here, since otherwise `recover_signer()` almost makes it look like it could just be returning the signing info rather than consuming `self` and returning the full recovered transaction.

theo-zil (commented):
I would agree that `Recovered` is slightly confusing, it makes it sound as if the transaction itself was recovered somewhere.
Can I suggest `VerifiedTransaction`, as to my understanding `recover_signer()` also automatically performs signature verification? And in fact in most cases it seems that that's usually the main reason you'd want a transaction post-`recover_signer()`, and the fact that the signer information is also cached for convenience is almost incidental for the occasional case where you might want it (and why not, now that you've calculated it).

Maybe even change the method to something like `recover_signer_and_verify()`.

JamesHinshelwood:
Thanks both - I've renamed `RecoveredTransaction` to `VerifiedTransaction` and `recover_signer` to `verify`. Please re-review @n-hutton :)

n-hutton (approved):

PR #513, opened 2023-10-25
Simplify address type

JamesHinshelwood:
Instead of defining `Address` as a newtype of `Address(H160)`, we instead just alias `Address = H160`. This simplifies quite a bit of code and I'd argue it doesn't cost us any type safety, because we aren't using `H160` for anything other than addresses. The only thing this might make more difficult in the future is changing the width of addresses, but I'd argue that will be difficult anyway and its not worth paying the complexity cost now.


n-hutton (approved):

theo-zil (approved):
Meant to approve yesterday, whoops

PR #515, opened 2023-10-26
squash imports

n-hutton:
does this spark joy?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,12 +15,12 @@ env:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     steps:
-    - name: Clear some disk space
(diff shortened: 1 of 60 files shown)


JamesHinshelwood:
Would be nice to have `rustfmt` do this for us automatically...
https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=import#imports_granularity

JamesHinshelwood:
Oh, you added the option :) Do we need to do anything else in CI/locally to make sure it applies? I remember there being some caveat about the feature being unstable?

While we're here, should we also set `group_imports = StdExternalCrate`?

n-hutton:
Sure, if you want.

The features are unstable in a use case where you mix comments into your imports, I think it is unlikely to cause an issue, and even if it does, we can just go back to imports living their best life as they are.

JamesHinshelwood (approved):

PR #516, opened 2023-10-27
add view to vote and QC

### LINKED ISSUE #44: Include the view number in votes and QCs
This makes it easier to handle situations with missing blocks, e.g. when we receive votes for missing blocks or we receive a new view messages with a QC of a missing block

n-hutton:
closes #44


zilliqa/src/message.rs
```diff
@@ -53,24 +54,47 @@ impl Proposal {
...
+    }
+
     pub fn verify(&self) -> Result<()> {
-        self.public_key
-            .verify(self.block_hash.as_bytes(), self.signature)
+        let mut bytes = Vec::new();
+        bytes.extend_from_slice(self.block_hash.as_bytes());
+        bytes.extend_from_slice(&self.view.to_be_bytes());
+
+        warn!("Verify vote: {:?}", self);
```
theo-zil, line 95: This seems more like a trace, or info at most

theo-zil (changes requested):
LGTM if the log is changed

zilliqa/src/message.rs
n-hutton, line 95:
```suggestion
        trace!("Verify vote: {:?}", self);
```

theo-zil (approved):

PR #524, opened 2023-10-31
Forking tests

### LINKED ISSUE #12: Review consensus
I haphazardly copy/pasted the consensus algorithm from an outdated version of https://github.com/Zilliqa/zilliqa-experimental/tree/users/zoltan/consensus/fast-hotstuff .

We should double check that I didn’t miss anything.

After this ticket we can retire the experimental implementation and treat the zq2 implementation as the source of truth.

We may also take this opportunity to extract out the consensus algorithm from `node.rs`.

### LINKED ISSUE #13: Handle genesis blocks
Right now we generate mostly nonsense genesis blocks and skip validation on blocks 0, 1 and 2.

This feels unsatisfying and incorrect. We should do something sensible instead.

### LINKED ISSUE #236: Download missing blocks from the network when we join
Otherwise, nodes cannot join the network unless they were there when the genesis block was formed.

n-hutton:
closes #236  closes #13 I think and closes #12 

This PR explicitly tests forking by waiting until a propose message goes out, and then only sending it to one of the nodes.

The other nodes eventually move to a new view, and this node will have to revert the proposed blocks.

Actually though, because of the way it is designed, once the other nodes newView, they will also ask for and recieve this block, but nobody willl vote on it anymore. So everyone will have to fork.

Few other improvements. Comments are inline.

### DIFF
--- a/zilliqa-macros/src/test.rs
+++ b/zilliqa-macros/src/test.rs
@@ -22,6 +22,8 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
             // The original test function
             #input
 
+            let run_sequentially = std::env::var_os("ZQ_RUN_SEQUENTIALLY").is_some();
+
             // Work out what RNG seeds to run the test with.
             let seeds: Vec<u64> = if let Some(seed) = std::env::var_os("ZQ_TEST_RNG_SEED") {
                 vec![seed.to_str().unwrap().parse().unwrap()]
@@ -29,26 +31,36 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
                 let samples: usize = std::env::var_os("ZQ_TEST_SAMPLES")
                     .map(|s| s.to_str().unwrap().parse().expect(&format!("Failed to parse ZQ_TEST_SAMPLES env var: {:?}", s)))
                     .unwrap_or(1);
-                // Generate random seeds using the thread-local RNG.
-                rand::Rng::sample_iter(rand::thread_rng(), rand::distributions::Standard).take(samples).collect()
+                (0..samples).map(|x| x as u64).collect()
             };
 
             let mut set = tokio::task::JoinSet::new();
 
             for seed in seeds {
-                set.spawn(async move {
+                println!("Reproduce this test run by setting ZQ_TEST_RNG_SEED={seed}");
+                let mut rng = <rand_chacha::ChaCha8Rng as rand_core::SeedableRng>::seed_from_u64(seed);
+                let network = crate::Network::new(std::sync::Arc::new(std::sync::Mutex::new(rng)), 4, seed);
+
+                if run_sequentially {
                     // Set up a tracing subscriber, so we can see logs from failed test cases.
                     let subscriber = tracing_subscriber::fmt()
                         .with_ansi(false)
                         .with_env_filter(tracing_subscriber::EnvFilter::from_default_env());
                     let _guard = tracing_subscriber::util::SubscriberInitExt::set_default(subscriber);
 
-                    println!("Reproduce this test run by setting ZQ_TEST_RNG_SEED={seed}");
-                    let mut rng = <rand_chacha::ChaCha8Rng as rand_core::SeedableRng>::seed_from_u64(seed);
-                    let network = crate::Network::new(std::sync::Arc::new(std::sync::Mutex::new(rng)), 4, seed);
-                    // Call the original test function
                     #inner_name(network).await;
-                });
+                } else {
+                    set.spawn(async move {
+                        // Set up a tracing subscriber, so we can see logs from failed test cases.
+                        let subscriber = tracing_subscriber::fmt()
+                            .with_ansi(false)
+                            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env());
+                        let _guard = tracing_subscriber::util::SubscriberInitExt::set_default(subscriber);
+
+                        // Call the original test function
+                        #inner_name(network).await;
+                    });
+                }
             }
 
             while let Some(result) = set.join_next().await {
--- a/zilliqa-macros/src/test.rs
+++ b/zilliqa-macros/src/test.rs
@@ -22,8 +22,6 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
             // The original test function
             #input
 
-            let run_sequentially = std::env::var_os("ZQ_RUN_SEQUENTIALLY").is_some();
-
             // Work out what RNG seeds to run the test with.
             let seeds: Vec<u64> = if let Some(seed) = std::env::var_os("ZQ_TEST_RNG_SEED") {
                 vec![seed.to_str().unwrap().parse().unwrap()]
@@ -31,40 +29,66 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
                 let samples: usize = std::env::var_os("ZQ_TEST_SAMPLES")
                     .map(|s| s.to_str().unwrap().parse().expect(&format!("Failed to parse ZQ_TEST_SAMPLES env var: {:?}", s)))
                     .unwrap_or(1);
-                (0..samples).map(|x| x as u64).collect()
+                // Generate random seeds using the thread-local RNG.
+                rand::Rng::sample_iter(rand::thread_rng(), rand::distributions::Standard).take(samples).collect()
             };
 
             let mut set = tokio::task::JoinSet::new();
 
-            for seed in seeds {
-                println!("Reproduce this test run by setting ZQ_TEST_RNG_SEED={seed}");
-                let mut rng = <rand_chacha::ChaCha8Rng as rand_core::SeedableRng>::seed_from_u64(seed);
-                let network = crate::Network::new(std::sync::Arc::new(std::sync::Mutex::new(rng)), 4, seed);
+            // Keep a map of task IDs to test seeds, so we can print the seed of tasks that fail. We can't print the
+            // seed as soon as the test has failed, because other cases will be running in parallel and we don't want
+            // to interlace different cases' seeds and error messages. Instead, we wait until we resolve tasks from the
+            // `JoinSet` (via `.join_next_with_id()`), where we can guarantee to only process one case at a time.
+            let mut id_to_seed = std::collections::HashMap::new();
 
-                if run_sequentially {
+            // Silence the default panic hook.
+            std::panic::set_hook(Box::new(|_| {}));
+
+            for seed in seeds {
+                let handle = set.spawn(async move {
                     // Set up a tracing subscriber, so we can see logs from failed test cases.
                     let subscriber = tracing_subscriber::fmt()
                         .with_ansi(false)
                         .with_env_filter(tracing_subscriber::EnvFilter::from_default_env());
                     let _guard = tracing_subscriber::util::SubscriberInitExt::set_default(subscriber);
 
-                    #inner_name(network).await;
-                } else {
-                    set.spawn(async move {
-                        // Set up a tracing subscriber, so we can see logs from failed test cases.
-                        let subscriber = tracing_subscriber::fmt()
-                            .with_ansi(false)
-                            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env());
-                        let _guard = tracing_subscriber::util::SubscriberInitExt::set_default(subscriber);
+                    let mut rng = <rand_chacha::ChaCha8Rng as rand_core::SeedableRng>::seed_from_u64(seed);
+                    let network = crate::Network::new(std::sync::Arc::new(std::sync::Mutex::new(rng)), 4, seed);
 
-                        // Call the original test function
-                        #inner_name(network).await;
-                    });
-                }
+                    // Call the original test function, wrapped in `catch_unwind` so we can detect the panic.
+                    let result = futures::FutureExt::catch_unwind(std::panic::AssertUnwindSafe(#inner_name(network))).await;
+
+                    match result {
+                        Ok(()) => {},
+                        Err(e) => {
+                            std::panic::resume_unwind(e);
+                        }
+                    }
+                });
+                id_to_seed.insert(handle.id(), seed);
             }
+            // Restore the default panic hook, so the panic we actually care about gets printed.
+            let _ = std::panic::take_hook();
 
-            while let Some(result) = set.join_next().await {
-                let () = result.unwrap();
+            while let Some(result) = set.join_next_with_id().await {
+                match result {
+                    Ok((_, ())) => {},
+                    Err(e) => {
+                        let id = e.id();
+                        if let Ok(p) = e.try_into_panic() {
+                            // Silence the panic hook again. When we break from this loop and the `JoinSet` is dropped,
+                            // the remaining tasks will be aborted. If any of them have also failed, we don't want
+                            // their logs to be printed.
+                            std::panic::set_hook(Box::new(|_| {}));
+
+                            let seed = id_to_seed.get(&id).unwrap();
+                            println!("Reproduce this test run by setting ZQ_TEST_RNG_SEED={seed}");
+                            std::panic::resume_unwind(p);
+                        } else {
+                            panic!("task cancelled")
+                        }
+                    },
+                }
             }
         }
     }
--- a/zilliqa/src/block_store.rs
+++ b/zilliqa/src/block_store.rs
@@ -2,6 +2,7 @@ use crate::message::ExternalMessage;
 use std::{cell::RefCell, num::NonZeroUsize};
 
 use anyhow::Result;
+use libp2p::PeerId;
 use lru::LruCache;
 use sled::{Db, Tree};
 use tracing::*;
@@ -100,12 +101,25 @@ impl BlockStore {
         Ok(())
     }
 
-    pub fn request_blocks(&mut self, number: u64) -> Result<()> {
-        self.message_sender
-            .broadcast_external_message(ExternalMessage::BlockBatchRequest(BlockBatchRequest(
-                BlockRef::Number(number),
-            )))
-            .unwrap();
+    pub fn request_blocks(&mut self, peer: Option<PeerId>, number: u64) -> Result<()> {
+        let request = ExternalMessage::BlockBatchRequest(BlockBatchRequest(BlockRef::Number(number)));
+        match peer {
+            Some(peer) => {
+                self.message_sender
+                    .send_external_message(peer, request)
+                    .unwrap();
+            }
+            None => {
+                self.message_sender
+                    .broadcast_external_message(request)
+                    .unwrap();
+            }
+        }
+        //self.message_sender
+        //    .send_external_message(peer, ExternalMessage::BlockBatchRequest(BlockBatchRequest(
+        //        BlockRef::Number(number),
+        //    )))
+        //    .unwrap();
 
         Ok(())
     }
--- a/zilliqa/src/block_store.rs
+++ b/zilliqa/src/block_store.rs
@@ -102,7 +102,10 @@ impl BlockStore {
     }
 
     pub fn request_blocks(&mut self, peer: Option<PeerId>, number: u64) -> Result<()> {
-        let request = ExternalMessage::BlockBatchRequest(BlockBatchRequest(BlockRef::Number(number)));
+        let request =
+            ExternalMessage::BlockBatchRequest(BlockBatchRequest(BlockRef::Number(number)));
+
+        // We can request blocks from a single peer or from all peers.
         match peer {
             Some(peer) => {
                 self.message_sender
@@ -115,11 +118,6 @@ impl BlockStore {
                     .unwrap();
             }
         }
-        //self.message_sender
-        //    .send_external_message(peer, ExternalMessage::BlockBatchRequest(BlockBatchRequest(
-        //        BlockRef::Number(number),
-        //    )))
-        //    .unwrap();
 
         Ok(())
     }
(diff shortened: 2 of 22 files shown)


zilliqa/src/block_store.rs
```diff
@@ -100,12 +101,25 @@ impl BlockStore {
...
 
-    pub fn request_blocks(&mut self, number: u64) -> Result<()> {
-        self.message_sender
-            .broadcast_external_message(ExternalMessage::BlockBatchRequest(BlockBatchRequest(
-                BlockRef::Number(number),
-            )))
-            .unwrap();
+    pub fn request_blocks(&mut self, peer: Option<PeerId>, number: u64) -> Result<()> {
+        let request = ExternalMessage::BlockBatchRequest(BlockBatchRequest(BlockRef::Number(number)));
+        match peer {
```
n-hutton, line 106: reduces the amount of unneeded messages which makes the tests faster etc.

zilliqa/src/consensus.rs
```diff
@@ -600,6 +618,11 @@ impl Consensus {
             return Ok(None);
         }
 
+        if block.view() <= head_block.header.view {
```
n-hutton, line 621: Some of the lower down logic relies on making sure the block is definetly from a higher view (reverting when forking).

zilliqa/src/consensus.rs
```diff
@@ -635,6 +658,21 @@ impl Consensus {
             );
 
             let mut block_receipts: Vec<TransactionReceipt> = Vec::new();
+
+            if transactions.len() > 0 {
+                trace!("applying {} transactions to state", transactions.len());
+            }
+
+            // Must make sure state root hash is set to the parent's state root hash before applying transactions
```
n-hutton, line 666: Previously forks were tolerated becaue most likely you would switch to a different branch and all the TXs would fail due to being executed before but the state hash would line up anyway.

zilliqa/src/consensus.rs
```diff
@@ -1777,4 +1821,50 @@ impl Consensus {
...
+
+            // State is easily set - must be to the parent block, though
+            self.state.set_to_root(H256(parent_block.state_root_hash().0));
+
+            // Persistence - only need to update head block pointer as it should be impossible
+            // to change finalized height
+            let new_highest = head_block.header.number.saturating_sub(1);
+            self.db.insert(HIGHEST_BLOCK_NUMBER, &new_highest.to_be_bytes())?;
+
+            // touched address index: TODO
```
n-hutton, line 1865: ticket #525

zilliqa/src/message.rs
```diff
@@ -278,6 +281,24 @@ pub struct BlockHeader {
...
+        write!(f, "Block Number: {} ", self.number)?;
+        write!(f, "Block Hash: {} ", self.hash)?;
+        write!(f, "Parent Hash: {} ", self.parent_hash)?;
+        write!(f, "State Root Hash: {} ", self.state_root_hash)?;
+        write!(f, "Timestamp: {} ", format_system_time(self.timestamp))?;
+        Ok(())
+    }
+}
+
+// Helper function to format SystemTime as a string
```
n-hutton, line 296: Now blocks are printed in display mode like:

```
Header: View: 9 Block Number: 9 Block Hash: 743703b84dcacae8336e5222635941b935ca25e19d0bf8594782861d8bca5579 Parent Hash: 6e66f4b922d29fccb34fbc2e8575eca7ea626682c22b7da5f25dbc10f051373c State Root Hash: 18781bbe4d066138ad97028ff30b6145f70508b554d90b9441d5c7b5d18f08b9 Timestamp: 1970-01-01T00:00:00+00:00  QC hash: 6e66f4b922d29fccb34fbc2e8575eca7ea626682c22b7da5f25dbc10f051373c Transactions: [bf244a927d89ebe2f81a6e2bb11f1a82dfe7deabf0ba536528a946ca4355ce4a] Committee: [PeerId("12D3KooWDa88EfYGyVs5KqZsJdDQuejdF7s4bEoYZB6THvD9aSX9"), PeerId("12D3KooWF2a6JqjePGskZ25BCxauQb6hGWEARi7FqWAVFc2nNW5Q"), PeerId("12D3KooWGqg8XegNFJT7qeHBpDkyCBME97W8WbUutx9J7k6pqRjw"), PeerId("12D3KooWKCRdm8jkthXcPyH6ZjJibe1epddbS3jkHpaGuF4CMQvY")]
```

zilliqa/src/message.rs
```diff
@@ -439,6 +460,16 @@ impl Committee {
     pub fn add_validators(&mut self, validators: impl IntoIterator<Item = Validator>) {
         self.0.extend(validators);
     }
+
+    pub fn remove_by_peer_id(&mut self, peer_id: PeerId) {
+        self.0.retain(|v| v.peer_id != peer_id);
+    }
+
+    pub fn choose_random(&mut self) -> Validator {
```
n-hutton, line 468: ❗ - I really like that in the tests everything is deterministic. This potentially makes it not so - when you are requesting someone to update you with the chain, you request another peer with true random probability.

Is there a way to do this nicely with libp2p? Even so I suspect that it would be random, still. Alternatives are to request based on an xor of your peer id with chain height or something, seems hacky though.

zilliqa/src/node.rs
```diff
@@ -495,12 +495,13 @@ impl Node {
...
         );
         let mut was_new = false;
+        let length_recvd = response.blocks.len();
 
         for block in response.blocks {
             was_new = self.consensus.receive_block(block)?;
         }
 
-        if was_new {
+        if was_new && length_recvd > 1 {
```
n-hutton, line 504: Essentially, when you only get a single block back from a request that should be many blocks, you can assume there is nothing more to request.

zilliqa/src/state.rs
```diff
@@ -36,6 +36,12 @@ use crate::{
 };
 
 #[derive(Debug)]
+/// The state of the blockchain, consisting of:
+/// -  state - a database of Map<Address, Map<key,value>>
+/// -  accounts, Map<Address, Account>
+/// where an address is a 20-byte array representing a user.
+/// where Account is (nonce, code, storage_root)
+/// the storage root is used to index into the state
```
n-hutton, line 44: double check?

zilliqa/tests/it/consensus.rs
```diff
@@ -219,3 +214,86 @@ async fn launch_shard(mut network: Network) {
...
+    let original_receipt = first.unwrap();
+
+    trace!("nathan.");
+    trace!("Running until the network has reverted the block");
+    // Now we should be able to run the network until we get a different tx receipt from the first
+    // node, which indicates that it has reverted the block
+    network
+        .run_until(
+            |n| {
+                let receipt = n.get_node(0).get_transaction_receipt(hash.into());
```
n-hutton, line 288: specifically, the receipt here must exist and have a different block hash (not just stop existing)

zilliqa-macros/src/test.rs
```diff
@@ -22,33 +22,45 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
             // The original test function
             #input
 
+            let run_sequentially = std::env::var_os("ZQ_RUN_SEQUENTIALLY").is_some();
```
n-hutton, line 25: convenient for when it will fail when running many seeds.

I actually did have it failing on multiple seed, but I have yet to encounter the situation where it always fails on the same seed, so there is some non-determanism in the test I believe.

n-hutton: Also, the println always prints at the end, its a todo to interleave it with the tracing framework somehow

zilliqa/src/consensus.rs
```diff
@@ -496,11 +497,25 @@ impl Consensus {
             .unwrap()
     }
 
-    // This function is called when we suspect that we are out of sync with the network/need to catchup
+    /// This function is called when we suspect that we are out of sync with the network/need to catchup.
+    /// We ask peers for their chain above our head, if the network is syncronised there should be nothing
+    /// to return.
```
n-hutton, line 502: open to debate on this solution - it sort of assumes that validators are the only people who exist on the network which seems incorrect/bad

zilliqa/src/consensus.rs
```diff
@@ -730,7 +784,8 @@ impl Consensus {
 
         // If we haven't applied the transaction yet, do so. This ensures we don't execute the transaction twice if we
         // already executed it in the process of proposing this block.
```
n-hutton, line 786: OK, so I do not think it makes sense for this function to not execute if it thinks the TX has been executed before. It also represented a bug where when you are proposing a block, it might not be from your HEAD, you can change the state root, but the TX will not execute due to this check.

If the tx has been executed before, it should fail the nonce check.

zilliqa/src/consensus.rs
n-hutton: Yeah, that is fair, the check can go after the fork unwinding, then

zilliqa/src/message.rs
n-hutton: Yup seems great

zilliqa/src/message.rs
n-hutton: I think it's too convoluted atm. Maybe lets see what James thinks when get gets back. Now that I think of it, another way you could do it would be to override the random import, much like we do with systemtime. That way you can use random freely in the code.

zilliqa/src/consensus.rs
n-hutton: Yes, I agree with this, but it is def a TODO which will have to be addressed at some point.

zilliqa-macros/src/test.rs
```diff
@@ -22,33 +22,45 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
...
             // Work out what RNG seeds to run the test with.
             let seeds: Vec<u64> = if let Some(seed) = std::env::var_os("ZQ_TEST_RNG_SEED") {
                 vec![seed.to_str().unwrap().parse().unwrap()]
             } else {
                 let samples: usize = std::env::var_os("ZQ_TEST_SAMPLES")
                     .map(|s| s.to_str().unwrap().parse().expect(&format!("Failed to parse ZQ_TEST_SAMPLES env var: {:?}", s)))
                     .unwrap_or(1);
-                // Generate random seeds using the thread-local RNG.
-                rand::Rng::sample_iter(rand::thread_rng(), rand::distributions::Standard).take(samples).collect()
+                (0..samples).map(|x| x as u64).collect()
```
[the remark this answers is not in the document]
n-hutton: That is correct. I think it is cleaner to have it like this because:
A) When you look at the seed it failed on you get an idea of how many seeds it took until failure.
B) You are guaranteed that you are running the same seeds as the CI.

The slight issue is that the way it is currently, the CI will spit out that you failed and replicate this test by running seed X.

However, since all of the tests are running async, I believe it is the case that the last seed printed out doesn't neccessarily correspond to the failing test. This makes it a bit easier to replicate the failing CI env locally.

The downside is of course that the tests will be run against seeds 0-100 forever so there is some loss of coverage as it were.

zilliqa/src/consensus.rs
```diff
@@ -635,6 +661,25 @@ impl Consensus {
...
+            if !transactions.is_empty() {
+                trace!("applying {} transactions to state", transactions.len());
+            }
+
+            // Must make sure state root hash is set to the parent's state root hash before applying transactions
+            if self.state.root_hash()? != parent.state_root_hash() {
+                warn!("state root hash prior to block execution mismatch, expected: {:?}, actual: {:?}", parent.state_root_hash(), self.state.root_hash()?);
+            }
+
+            if head_block.hash() != parent.hash() || block.number() != head_block.header.number + 1
```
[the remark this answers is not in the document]
n-hutton: No, the parent hash mismatch is not good enough because if there are no transactions in the network, both branches will have the same hash for every block. Part of dealing with the fork is making sure you re-point head_block, which I have written it such that head_block is the absolute reference point for the chain (finalized block being more of a side detail).

WRT should the check be `block.number() < head_block.header.number` - I see where you are getting at but I don't think that `deal_with_fork` should be able to unwind to genesis. Consider the loop:

```
        while self.head_block().header.number >= block.header.number {
```

It only loops while head block is >= proposal. So the final head_block height should just be proposed block - 1 if it is working correctly.

As for the case where the block height is >> head_block - I believe that this should be impossible in a healthy system, in any case there is a check for it in `check_block`: 

```
        // Blocks must be in sequential order
        if block.header.number != parent.header.number + 1 {
```

Perhaps it can occur while you are syncronizing and you approach HEAD.

zilliqa/src/consensus.rs
```diff
@@ -1777,4 +1841,53 @@ impl Consensus {
...
+    fn deal_with_fork(&mut self, block: &Block) -> Result<()> {
+        // Start at the head block and recursively unwind the changes that have been made
+        // until the head block is the parent of this block.
+        while self.head_block().header.number >= block.header.number {
+            let head_block = self.head_block();
+            let parent_block = self
+                .get_block(&head_block.parent_hash())?
+                .ok_or_else(|| anyhow!("missing block parent when reverting blocks!"))?;
+
+            if head_block.header.view == 0 {
```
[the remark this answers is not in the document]
n-hutton: I don't think genesis should ever come into the equation here - proposals should always be height >= 1 so the parent is genesis (for the case of a fork at genesis), but if logic is correct it will never need to request genesis parent.

The check is just for database corruption/race conditions. If you start a chain unwind and you have no HEAD parent.

I guess maybe it could happen on a light node too now I think of it.

zilliqa/tests/it/main.rs
```diff
@@ -420,8 +512,12 @@ impl Network {
             trace!("Messages were empty - advance time and trigger timeout in all nodes!");
             zilliqa::time::advance(Duration::from_millis(500));
 
-            for node in &self.nodes {
-                node.inner.lock().unwrap().handle_timeout();
+            for (index, node) in self.nodes.iter().enumerate() {
+                let span = tracing::span!(tracing::Level::INFO, "handle_timeout", index);
```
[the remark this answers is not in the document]
n-hutton: Yeah I like this one - much easier to debug what is going on with each node!

zilliqa/tests/it/main.rs
```diff
@@ -354,13 +354,125 @@ impl Network {
...
+            }
+        }
+        messages
+    }
+
+    // Take all the currently ready messages from the stream,
+    // remove N-1 propose messages we see where network size = N and the remaining one is
+    // the first node in the vector
+    // *** Only perform this when the propose message contains one or more txs.
+    pub async fn drop_propose_messages_except_one(&mut self) {
```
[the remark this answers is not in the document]
n-hutton: yeah good point. I guess in future for consensus tests the way to do it is maybe to have something like `network.run_until(|network, messages| { messages.filter(predicate); network.height() > N })` - essentially, the closure also lets you view the message queue so you can inject the message manipulation.

Not sure how clean that would end up looking though.

zilliqa/src/consensus.rs
```diff
@@ -1777,4 +1841,53 @@ impl Consensus {
...
+
+            // TX receipts are indexed by block
+            let _ = self.transaction_receipts.remove(head_block.hash().0);
+
+            // State is easily set - must be to the parent block, though
+            trace!("Setting state to: {}", parent_block.state_root_hash());
+            self.state
+                .set_to_root(H256(parent_block.state_root_hash().0));
+
+            // Persistence - only need to update head block pointer as it should be impossible
```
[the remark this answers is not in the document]
n-hutton: Oh yeah... wait, this logic is way more dangerous than I thought now you point it out.

Because quite early on in the flow you add the proposed block to the database. This changes the canonical_block_numbers mapping, if you were to accidentally refer by number rather than hash lower down. I think I have to put it after the safe block check at least.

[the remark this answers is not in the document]
n-hutton: ah, it isn't incorrect technically, but for whatever reason, moving the add block until after the safe check breaks consensus votes. I will ideally fix it in a follow-up PR.

zilliqa-macros/src/test.rs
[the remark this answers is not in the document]
theo-zil: I'm not a big fan, mainly because I'm worried about the loss of coverage. Now in theory the CI should have enough samples to uncover failures in any case, but given that we're randomising the messages anyway this will always be probabilistic and therefore a tradeoff. If there's some message ordering that just happens to never come up on the seeds 0-100, it's guaranteed to never come up in CI now, whereas currently any given CI run _might_ miss an error but would then likely catch whenever it gets re-ran (which means very soon, considering CI tends to get ran multiple times per PR, and we open new PRs regularly).

Locally too, when you're running a single sample (or sometimes I run with say 3-5, but rarely more), this makes it more annoying because mid-development you're likely to "overfit" to seed 0 and if there's any consensus bugs, you either have to run the full CI test, or manually randomise the seed, or you won't discover them until you push and wait for CI. Whereas currently ever `cargo test` is randomised so you gain confidence that your changes are probably mostly robust when tests pass.

> The slight issue is that the way it is currently, the CI will spit out that you failed and replicate this test by running seed X.
> However, since all of the tests are running async, I believe it is the case that the last seed printed out doesn't neccessarily correspond to the failing test.

Does this change actually improve that, though? If the latest number printed doesn't necessarily correspond to the failure, I don't see how making the numbers sequential would necessarily fix things.
And overall, while I agree that it's slightly annoying to isolate a failing seed, it usually only takes me a couple of minutes to bisect the likely candidates and find the one(s) responsible, and I'd rather keep the randomised coverage than make this step slightly easier at the expense of only ever testing the same 100 seeds henceforth.

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
theo-zil: You're probably right but then I don't think the `block.view == 0` check is necessary at all.
I think we are currently assuming that the database isn't going to be corrupted, because if it is then anything goes and every part of consensus is liable to failing. I'd remove this if-statement to simplify the code a bit and shorten the function.

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
theo-zil: Technically, canonical block numbers aren't finalized block numbers - it's just what's _currently_ considered to be the main chain. By definition, the highest two blocks in that mapping are non-finalized and therefore subject to change.

At least that's how it's currently implemented. We could change it to a finalized block mapping only. But I think a lot of our chain architecture uses a best-effort model where we pipeline block creation and optimistically assume that the latest blocks will be fine even before finalization. In any case it's not too different to querying for `Latest` or just generally for a non-finalized block by hash, in all these cases it might happen that that blocks ends up being reverted.

zilliqa/src/message.rs
JamesHinshelwood: I agree that keeping the tests deterministic is important.

I would probably reach for adding an `Rng` to `Consensus` as the simplest solution, since doing the import-mocking trick requires quite a bit more fiddling with features, etc. and is less immediately understandable.

zilliqa-macros/src/test.rs
JamesHinshelwood: I've seen a similar issue, where it is hard to match the `ZQ_TEST_SEED` messages with the runs that actually failed. This is annoying if only one seed failed and you get 100 possible seeds that caused it.

e.g.
https://github.com/Zilliqa/zq2/actions/runs/6670165505/job/18129486734#step:9:156

I will see if we can fix this more elegantly by controlling how we print things in the `JoinSet` stuff of the test harness...

JamesHinshelwood: #534 , sorry it took a while.

zilliqa-macros/src/test.rs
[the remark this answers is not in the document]
JamesHinshelwood: I agree with @theo-zil here.

Right now, we explore a different area of the large search space of possible computations every time anything runs the test suite. In my opinion, this can give us a high degree of confidence that our code is correct under most conditions.

If we wanted to test the same set of computations every time, I would argue we need to dramatically increase the number of samples we test, because we don't have the benefit of every subsequent PR contributing to further tests of previous merges.

However, I think #534 might address some the issues you mention in a different way @n-hutton?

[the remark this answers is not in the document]
n-hutton: Okay - with James' PR it makes it much more usable so I will revert this

zilliqa/src/message.rs
n-hutton: ^ will perform this in a follow up PR

n-hutton:
> I'd like to double-check I understand the high-level logic. Say a node has a finalized block `N` at height `n`, thus followed by (non-final) proposals `N1` and `N2`. So its blockchain looks like this:
>...
> 
> I still haven't read the consensus paper properly so maybe I'm missing some details, too. Since you just wrote this logic it'd be great to hear your explanation directly though, to help wrap my head around this.

Mostly correct, responded in slack, but importantly the way this has been designed, nodes immediately execute a block proposal, so it has NOT been confirmed by a supermajority yet, and could be forked.

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
n-hutton: Yeah, makes sense 👍

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
n-hutton: Just a note: the current implementation is, I believe, incorrect in that it considers blocks finalized so long as it is not one of the top two. This isn't actually the correct definition of finalized, which is that a block has a two blocks higher than it, where the views increment sequentially (so it is all happy path). I think (not too sure tho) that there can be arbitrary chain lengths that have no finalized blocks.

I don't see a lot of merit in referring to or thinking about finalized blocks at this stage, though. 

And, yeah I agree, don't think it should be changed to finalized block mapping only.

theo-zil (approved):

PR #538, opened 2023-11-10
Improve how contracts are pre-compiled and tested

JamesHinshelwood:
Instead of a separate `.json` file per contract, we instead have a single `compiled.json` which contains the output artifacts of all contracts.

We've stopped using the deprecated `--combined-json` option of `solc` and are instead using the `--standard-json` interface, which is what gives us the combined JSON output. External tooling is likely to have better support for this format.

We've also changed how the compilation assertion tests work. Instead of requiring the developer to manually run `solc` and check in the results, you are now expected to run the assertion test with the `ZQ_CONTRACT_TEST_BLESS` environment variable set. In this case, the test will recompile the contracts and update the `compiled.json` file on disk. If the variable is not set, the test compares the compiled value with the file on disk.

The final change is that we are using the `solc` support in `ethers-rs` to compile in the test. Under the hood, this still shells out to `solc`, but it uses `svm-rs` to download and manage the correct version. This means the test should now work on Windows and Mac, as well as Linux.


zilliqa/src/contracts/mod.rs
```diff
@@ -39,220 +36,119 @@ pub mod native_token {
...
-            .arg("--evm-version=paris")
-            .output()
-            .unwrap();
+        let output = solc.compile_exact(&input).unwrap();
+        let output_file = root.join("src").join("contracts").join("compiled.json");
 
-        eprintln!("{}", std::str::from_utf8(&output.stderr).unwrap());
-        let combined_json: Value = serde_json::from_slice(&output.stdout).unwrap();
-        let contract = combined_json["contracts"][&json_key].clone();
+        if std::env::var_os("ZQ_CONTRACT_TEST_BLESS").is_some() {
```
n-hutton, line 142: Bless? What is it supposed to mean? I might suggest something like `...TEST_ONEOFF_CONTRACT_COMPILE` or something as it is non-obvious.

zilliqa/src/contracts/mod.rs
```diff
@@ -39,220 +36,119 @@ pub mod native_token {
...
+            let file = File::create(output_file).unwrap();
+            serde_json::to_writer(file, &output).unwrap();
 
-        let bin = contract["bin-runtime"].as_str().unwrap();
-        let expected_code = hex::decode(bin).unwrap();
-        assert_eq!(code.len(), expected_code.len()); // fail fast on a mismatch rather than dumping
-                                                     // full blobs into stdout
-        assert_eq!(code, expected_code);
+            panic!("compiled contracts updated, run this test again with `ZQ_CONTRACT_TEST_BLESS` unset");
```
n-hutton, line 146: I read the PR description, but I was still initally confused the tests failed on the first run - I might put something like 'this is not a test failure, please rerun with...'

n-hutton (changes requested):
minor comments, looks great, though. Works on macos (although I already had solc so the download is untested)

zilliqa/src/contracts/mod.rs
JamesHinshelwood: I used "bless" to mean "bless the current output as the new expected output". I think I subconsciously stole the term from rustc: https://rustc-dev-guide.rust-lang.org/tests/ui.html#output-comparison

> UI tests store the expected output from the compiler in .stderr and .stdout files next to the test. You normally generate these files with the --bless CLI option, and then inspect them manually to verify they contain what you expect.

zilliqa/src/contracts/mod.rs
JamesHinshelwood: Thinking about it more, there's no reason we should force you to run once with the env var set and then again with it unset to see a green tick locally. I'll just print this message and let the test pass in this branch.

n-hutton (approved):

n-hutton (approved):

PR #562, opened 2023-11-20
Forking tests fix

n-hutton:
Fixes a few things:
- Includes proper BLS pub key aggregation (disabled due to increasing test time)
- Fixes an issue where when syncing the chain, you didn't execute blocks correctly

Comments are inline. The main fix is to request blocks with TXs also when block syncing, so that you can execute them properly. This uses the same call path as the fn that recieves block proposals, this way there isn't a discontinuity.

There are a few other changes which should fix the flakiness of the tests (only occurred when the proposer failed twice in a row).


zilliqa-macros/src/test.rs
```diff
@@ -17,13 +17,18 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
     input.sig.ident = inner_name.clone();
 
     quote! {
+        use ntest::*;
```
n-hutton, line 20: Use a * here to avoid complaints about re-importing multiple times. Now, tests have a timeout before they fail with a message about the test taking too long.

zilliqa-macros/src/test.rs
```diff
@@ -56,14 +61,18 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
                     let network = crate::Network::new(std::sync::Arc::new(std::sync::Mutex::new(rng)), 4, seed);
 
                     // Call the original test function, wrapped in `catch_unwind` so we can detect the panic.
-                    let result = futures::FutureExt::catch_unwind(std::panic::AssertUnwindSafe(#inner_name(network))).await;
+                    let span = tracing::span!(tracing::Level::INFO, "tst");
```
n-hutton, line 64: Useful if you are finding that your test only fails when you are running multiple at once (such as when time is interfering). Can add the seed or task id to the span when debugging.

zilliqa/src/consensus.rs
```diff
@@ -202,6 +204,8 @@ pub struct Consensus {
     db: Arc<Db>,
     /// Transactions ordered by priority, map of address of TXn (from account) to ordered TXns to be executed.
     new_transactions_priority: BTreeMap<Address, BinaryHeap<TxnOrder>>,
+    /// PRNG
+    rng: ChaCha8Rng,
```
n-hutton, line 208: Consensus has a PRNG that is seeded from 0. This makes the tests fully deterministic again, and since they are using the prng modulo the committee minus themselves, their behaviour shouldn't be identical every time. Open to seeding it with the secret key too.

zilliqa/src/consensus.rs
```diff
@@ -588,7 +604,7 @@ impl Consensus {
         trace!("checking if block view {} is safe", block.view());
 
         // If the proposed block is safe, vote for it and advance to the next round.
-        if self.check_safe_block(&block)? {
+        if self.check_safe_block(&block, during_sync)? {
```
n-hutton, line 607: So basically one of the issues is that when one of the nodes is behind, it needs to receive blocks and execute them.

The logic it needs to execute them (and all checks) is identical for the proposal logic, with the exception that in certain circumstances, the node's own view can be ahead of the block (but the block is still valid).

One way this can happen is if the node is the only one to get a propose message. It will increment its view after executing the propose message, but the rest of the network will be on a lower view and will eventually propose an alternate block.

So, if you are executing the block during sync, you can ignore specifically the view check.

zilliqa/src/consensus.rs
```diff
@@ -642,6 +658,11 @@ impl Consensus {
                     .insert_transaction_receipt(&block.hash(), &block_receipts)?;
             }
 
+            // Important - only add blocks we are going to execute because they can potentially
+            // overwrite the mapping of block height to block, which there should only be one of.
+            // for example, this HAS to be after the deal with fork call
+            self.add_block(block.clone())?;
```
n-hutton, line 664: This is the original bug - the block was added to the db before the fork was dealt with. BUT, during the fork resolution, when it was getting head_block, this was the new block at that height, and not the proper head block.

The fix is to not add blocks to the db unless they are about to be executed. There is only one mapping of block height to block. Loose blocks are discarded.

zilliqa/src/consensus.rs
```diff
@@ -1319,11 +1344,7 @@ impl Consensus {
         cosigned: BitVec,
     ) -> QuorumCertificate {
         // we've already verified the signatures upon receipt of the responses so there's no need to do it again
-        QuorumCertificate {
-            signature: NodeSignature::aggregate(signatures).unwrap(),
-            cosigned,
-            block_hash,
-        }
+        QuorumCertificate::new(signatures, cosigned, block_hash)
```
n-hutton, line 1347: It seems poor practice to put the signature aggregation outside of the QC constructor

zilliqa/src/consensus.rs
```diff
@@ -1394,42 +1415,42 @@ impl Consensus {
         }
     }
 
-    fn check_and_commit(&mut self, proposal_hash: Hash) -> Result<()> {
-        let Some(proposal) = self.get_block(&proposal_hash)? else {
-            trace!("block not found: {proposal_hash}");
-            return Ok(());
+    fn check_and_commit(&mut self, proposal: &Block) -> Result<()> {
```
n-hutton, line 1418: The finalized logic here is rewritten to be actually correct. It is important for the block verification logic to work correctly (discarding/not voting on blocks that QC points before finalized). Note that the finalized is not HEAD-3 neccessarily if the views are not contigous.

zilliqa/src/consensus.rs
```diff
@@ -1579,25 +1607,45 @@ impl Consensus {
...
                 trace!(
-                    "updating high QC and view, blocks seems good! {} {} {}",
+                    "updating high QC and view, blocks seems good! hash: {} number: {} view: {}",
                     block.hash(),
                     block.number(),
                     block.view()
                 );
-                self.update_high_qc_and_view(block.agg.is_some(), block.qc.clone())?;
-                self.add_block(block)?;
+                self.proposal(Proposal::from_parts(block, transactions), true)?;
```
n-hutton, line 1644: we call the proposal logic here. Perhaps it should be renamed to 'execute block' since here, we are not treating it entirely like a proposal.

zilliqa/src/consensus.rs
```diff
@@ -1694,9 +1742,22 @@ impl Consensus {
...
-    fn verify_qc_signature(&self, _: &QuorumCertificate) -> Result<()> {
-        // TODO: Build aggregate signature from public keys and validate `qc.block_hash` against `qc.signature`.
-        Ok(())
+    fn verify_qc_signature(
+        &self,
+        qc: &QuorumCertificate,
+        public_keys: Vec<NodePublicKey>,
+    ) -> Result<()> {
+        let len = public_keys.len();
+        match qc.verify(public_keys) {
```
n-hutton, line 1751: the bitvec wasn't quite right when enabling the QC on the first iteration but is now

zilliqa/src/node.rs
```diff
@@ -433,20 +433,50 @@ impl Node {
...
+                    .get_transaction_by_hash(*tx_hash)
+                    .unwrap()
+                    .unwrap()
+            })
+            .collect::<Vec<_>>();
+
         self.message_sender.send_external_message(
             source,
-            ExternalMessage::BlockResponse(BlockResponse { block }),
+            ExternalMessage::BlockResponse(BlockResponse {
```
n-hutton, line 450: the block sync mechanism now needs to also include TXs since you need to execute the block. So it now returns proposals. I chose this because it was cleaner/faster, but the most optimal would be to rely on the TX sync mechanism being so good you never don't have one.

zilliqa/src/time.rs
```diff
@@ -23,13 +23,15 @@ pub use time_impl::*;
 #[cfg(feature = "fake_time")]
 mod time_impl {
     use serde::{Deserialize, Serialize};
+    use std::collections::HashMap;
```
n-hutton, line 26: I was getting an issue where running `cargo test --all-targets --all-features` would cause a single test to fail, but when I ran it with that seed, or just individually that test, it would pass.

It seems like this was caused by other tests advancing time concurrently to the failing test, making it advance views when the test wasn't expecting that.

So, if we use tokio task id to map to the time since epoch, we can isolate tests from each other.

This does mean that sometimes the test id could be reused by tokio and a test will start not at epoch zero... but I don't believe that's a particular problem.

zilliqa/tests/it/main.rs
```diff
@@ -601,7 +605,14 @@ impl Network {
                 } else {
                     self.nodes.iter().collect()
                 };
-                for (index, node) in nodes.iter().enumerate() {
+
+                for node in &nodes {
```
n-hutton, line 609: Btw, this was incorrect in the last PR to master; you can't use the iterated index because then it always seems like the message is going to node 0, when it isn't.

zilliqa/tests/it/main.rs
```diff
@@ -545,7 +549,7 @@ impl Network {
 
         if messages.is_empty() {
             trace!("Messages were empty - advance time and trigger timeout in all nodes!");
-            zilliqa::time::advance(Duration::from_millis(500));
+            zilliqa::time::advance(Duration::from_millis(1000));
```
n-hutton, line 552: Now that time isn't being advanced from other tests...

.github/workflows/ci.yaml
```diff
@@ -11,7 +11,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   # Run a decent number of samples for our randomized
-  ZQ_TEST_SAMPLES: 100
+  ZQ_TEST_SAMPLES: 50
```
JamesHinshelwood, line 14: I'd argue against this. The fact that we managed to break the tests suggests we didn't have enough samples, so reducing them further makes me worry. What's the motivation for this?

zilliqa/Cargo.toml
```diff
@@ -55,6 +56,7 @@ paste = "1.0.14"
 primitive-types = { version = "0.12.2", features = ["serde"] }
 prost = "0.12.2"
 rand = "0.8.5"
+rand_chacha = "0.3.1"
```
JamesHinshelwood, line 59: Can you remove this from `dev-dependencies` if we're adding it here?

zilliqa/src/crypto.rs
```diff
@@ -38,6 +38,25 @@ impl NodeSignature {
...
 
+    // Verify that the aggregated signature is valid for the given public keys and message.
+    // That is, each public key has signed the message, and the aggregated signature is the
+    // aggregation of those signatures.
+    pub fn verify_aggregate(
+        signature: &NodeSignature,
+        message: &[u8],
+        public_keys: Vec<NodePublicKey>,
+    ) -> Result<()> {
+        // first verify that the signature itself has been composed from the public keys and the message
```
JamesHinshelwood, line 49: Can we avoid forking `bls_signatures` by constructing the public key ourselves here?
e.g. something like
```rust
let aggregate_key: bls_signatures::PublicKey = public_keys.iter().map(|p| G1Projecttive::from(p.0)).sum().into();
aggregate_key.verify(signature.0, message);
```

Creating and maintaining a fork seems like quite a bit of effort, when I think we could just get away with implementing what we want directly here.

zilliqa/src/message.rs
```diff
@@ -125,11 +133,11 @@ pub struct BlockBatchRequest(pub BlockRef);
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct BlockResponse {
-    pub block: Block,
+    pub proposal: Proposal,
```
JamesHinshelwood, line 136: What's the reason for this change? If I understand correctly, this will now mean blocks and the inner transactions will all be included in a response?

zilliqa/tests/it/consensus.rs
```diff
@@ -112,7 +113,7 @@ async fn block_production(mut network: Network) {
                     .map_or(0, |b| b.number())
                     >= 10
             },
-            50000,
+            500000,
```
JamesHinshelwood, line 116: Was this meant to be included?

zilliqa-macros/src/test.rs
```diff
@@ -17,13 +17,18 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
     input.sig.ident = inner_name.clone();
 
     quote! {
+        use tracing::*;
+        use ntest::*;
+
         #[tokio::test(flavor = "multi_thread")]
+        #[timeout(300000_00)]
```
JamesHinshelwood, line 24: IMO it doesn't make sense to time tests out based on wall-clock time, but instead based on the number of ticks of our simulated network. This will remain deterministic across machines, rather than having the situation where if a machine is too slow, its impossible to run our test suite.

JamesHinshelwood (changes requested):
There are quite a few distinct changes in this PR, which makes it difficult to review because:

* I need to context switch between different logical changes and their motivations as I read the linear list of changes.
* Its difficult to understand which changes correspond to which comments in the PR description.
* I have to give an aggregate opinion on the whole PR, so I can't say lets merge some changes, while having a discussion on others. This is particularly important, because one of the changes fixes the flaky tests, which we'd obviously like to merge ASAP.

How sad would you be if I asked you to split these changes up into separate PRs so they can be reviewed and merged independently?

.github/workflows/ci.yaml
n-hutton: The CI struggles to run the tests/dies and it also takes a long time. This is a significant burden. I would argue that whatever is needed in the short term to make reduce overhead on PRs to a managable level is worth it

zilliqa/src/crypto.rs
n-hutton: No, there is a guard inside their library against verifying with multiple public keys signing the same message (preventing what we want to do), as it is a security vulnerability if you do it incorrectly.

zilliqa/src/message.rs
n-hutton: yes, see my comment above:

image

zilliqa/tests/it/consensus.rs
n-hutton: no, will revert

zilliqa-macros/src/test.rs
n-hutton: I will split this into a different PR for discussion.

zilliqa/src/consensus.rs
```diff
@@ -566,8 +582,11 @@ impl Consensus {
             Ok(()) => {}
             Err(e) => {
                 if let Some(e) = e.downcast_ref::<MissingBlockError>() {
-                    warn!("missing finalized block1");
-                    info!(?e, "missing block when checking block proposal");
+                    info!(?e, "missing block when checking block proposal - try and request the parent from the network: {}", block.header.number.saturating_sub(1));
+
+                    let random_peer = self.get_random_other_peer();
+                    self.block_store
+                        .request_blocks(random_peer, block.header.number.saturating_sub(1))?;
```
theo-zil, line 589: Would there be any harm in requesting up to the proposal head? Given network latency, if it happens that the proposal passes before the request is processed, then you'll be a block behind still and might have to repeat this process, whereas if you request it then you automatically catch up.

If it doesn't, then worst case the response will just have one less block than expected which I believe block_store handles without issue.

zilliqa/src/consensus.rs
```diff
@@ -1341,22 +1364,21 @@ impl Consensus {
...
-    fn check_safe_block(&mut self, proposal: &Block) -> Result<bool> {
+    fn check_safe_block(&mut self, proposal: &Block, during_sync: bool) -> Result<bool> {
         let Some(qc_block) = self.get_block(&proposal.qc.block_hash)? else {
             trace!("could not get qc for block: {}", proposal.qc.block_hash);
             return Ok(false);
         };
         // We don't vote on blocks older than our view
         let outdated = proposal.view() < self.view.get_view();
-        let proposal_hash = proposal.hash();
+        let _proposal_hash = proposal.hash();
```
theo-zil, line 1374: Is this unused - should it be removed?

zilliqa/src/consensus.rs
```diff
@@ -1341,22 +1364,21 @@ impl Consensus {
...
         match proposal.agg {
             // we check elsewhere that qc is the highest among the qcs in the agg
             Some(_) => match self.block_extends_from(proposal, &qc_block) {
                 Ok(true) => {
-                    let block_hash = proposal_hash;
-                    self.check_and_commit(block_hash)?;
+                    self.check_and_commit(proposal)?;
                     trace!("check block aggregate is outdated? {}", outdated);
```
theo-zil, line 1380: Very minor but I suspect this trace might be misleading for synced nodes - could either delete it, or update it with something like
```suggestion
                    trace!("check block aggregate is outdated? {outdated} {}", if during_sync { "but it doesn't matter because we're syncing" } else { "" });
```

zilliqa/src/consensus.rs
```diff
@@ -1394,42 +1416,42 @@ impl Consensus {
...
+        // have views N+1, N+2 (the final one being proposal block).
+
+        let Some(qc_block) = self.get_block(&proposal.qc.block_hash)? else {
+            warn!("missing qc block when checking whether to finalize!");
+            return Err(MissingBlockError::from(proposal.qc.block_hash).into());
         };
-        let Some(prev_1) = self.get_block(&proposal.qc.block_hash)? else {
-            trace!("parent not found: {}", proposal.qc.block_hash);
+
+        // At genesis it could be fine not to have a qc block, so don't error.
```
theo-zil, line 1435: Iff genesis is the only reason we don't error then I'd honestly make an explicit check for that, and still error if `qc_block.view() > 0`, only returning Ok(()) if it's actually genesis.

zilliqa/src/consensus.rs
```diff
@@ -1579,25 +1607,49 @@ impl Consensus {
...
+        // Check whether it is loose or not - we do not store loose blocks.
+        if !self.block_store.contains_block(block.parent_hash())? {
+            trace!("received block is loose: {}", block.hash());
+
+            warn!(
+                "missing received block the parent! Lets request the parent, then: {}",
+                block.parent_hash()
+            );
+            self.block_store
+                .request_blocks(None, block.header.number.saturating_sub(1))?;
```
theo-zil, line 1636: Likewise here: we've just received a block; we're requesting its parent now, but we are throwing away the block itself. Why not request up to the block's height, so if the parent does exist, we then store it _and_ the block?

zilliqa/src/consensus.rs
theo-zil: I wouldn't object to this

zilliqa/src/node.rs
```diff
@@ -433,20 +433,50 @@ impl Node {
             return Ok(());
         };
 
+        // Need to get and send corresponding TXs for the block
+        let txs: Vec<SignedTransaction> = block
```
theo-zil, line 437: Is there a reason not to use the new `block_to_proposal()` function added just below? It seems to duplicate this logic exactly (unless I've missed some detail!)

zilliqa/src/time.rs
```diff
@@ -72,7 +86,9 @@ mod time_impl {
 
     static PAUSED: AtomicBool = AtomicBool::new(false);
     /// Stores the duration between the currently set fake time time and the `UNIX_EPOCH`.
-    static CURRENT_TIME: OnceLock<Mutex<Duration>> = OnceLock::new();
+    /// This is on a tokio task basis, to avoid tests interfering with each other.
+    //static CURRENT_TIME: OnceLock<Mutex<Duration>> = OnceLock::new();
```
theo-zil, line 90: dev comment?

zilliqa/tests/it/consensus.rs
```diff
@@ -219,6 +220,7 @@ async fn launch_shard(mut network: Network) {
 // and progresses.
 #[zilliqa_macros::test]
 async fn handle_forking_correctly(mut network: Network) {
+    trace!("Handle forking correctly test starting....");
```
theo-zil, line 223: Should this trace be left in? I don't think the rest of our tests come with any logging by default outside of the network's message log.
On the other hand I guess it doesn't hurt, if you're debugging some other test you'll probably only run that one so this won't get in the way.

zilliqa/src/message.rs
```diff
@@ -77,7 +85,7 @@ impl Vote {
     ) -> Self {
         let mut bytes = Vec::new();
         bytes.extend_from_slice(block_hash.as_bytes());
-        bytes.extend_from_slice(&view.to_be_bytes());
+        //bytes.extend_from_slice(&view.to_be_bytes());
```
n-hutton, line 88: the votes and the QC are signing different things (not both include view) - address in a split-off PR

zilliqa/src/node.rs
n-hutton: ah, that is an oversight! thanks.

zilliqa/tests/it/consensus.rs
n-hutton: I didn't intend on it being there, thanks :)

I don't think traces in the tests are so bad, though.

zilliqa/src/consensus.rs
n-hutton: Yes, thanks!

zilliqa/src/consensus.rs
```diff
@@ -722,6 +729,10 @@ impl Consensus {
         if let Some(priority_txs) = self.new_transactions_priority.get_mut(&removed.signer) {
             priority_txs.retain(|tx| tx.hash != *tx_hash);
         }
+
+        // Finally, insert tx into the db, so there is no discontinuity where the TX can't
+        // be found anywhere
+        let _ = self.db.insert_transaction(tx_hash, &removed.tx);
```
theo-zil, line 735: This discards the Result, right - any particular reason rather than bubbling it up?

zilliqa/src/consensus.rs
theo-zil: As Zoltan clarified the correct condition is actually more lax (the two-chain has to have consecutive views `n-k`, `n-k+1` but the latest block at view `n` doesn't have to be consecutive, i.e. `k >= 2`). I'm fine with fixing that in a subsequent PR if you prefer, up to you

theo-zil (approved):
Couple minor comments but they're non-blocking on my part, so approving

zilliqa/src/consensus.rs
n-hutton: My auto formatter inserted it! I will fix in subsequent, thanks

PR #563, opened 2023-11-20
Use task-local values for fake time

JamesHinshelwood:
This means each test instance gets its own independent and deterministic value of fake time.


theo-zil (approved):
As mentioned in slack, I think I prefer this to the hashmap-based solution

PR #567, opened 2023-11-21
disable flaky e2e tests

n-hutton:


JamesHinshelwood (dismissed):
Should we just delete this file instead (and add it again when we make the tests less flaky)?

Also you might need to delete the workflow in `.github/`

JamesHinshelwood (approved):

PR #568, opened 2023-11-22
Force tests to have upper bound timeout

n-hutton:
image

I would argue that we're not trying to run our tests on a raspberry pi here. Having a high upper bound on timeout will allow the test to bomb out and indicate you are writing unperformant code/increasing the CI time too much in your PR. Also, if there are tests which sometimes can loop forever, the way the CI is now is that the pod eventually dies, leaving you no clue as to why. This would in that scenario at least provide you information.


JamesHinshelwood (commented):
I still feel quite strongly that this is a bad idea. If the motivation is to inform us if a test is written non performantly, I think its better to do that by counting the number of ticks of the simulated network, since that gives you a deterministic answer. If the problem is infinite looping in CI, then we should add a timeout at the level of CI (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes).

n-hutton:
> I still feel quite strongly that this is a bad idea. If the motivation is to inform us if a test is written non performantly, I think its better to do that by counting the number of ticks of the simulated network, since that gives you a deterministic answer. If the problem is infinite looping in CI, then we should add a timeout at the level of CI (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes).

Relevant: #579 

My LOGAF is : `middling`.

The motivation is to provide a gate to inform you that you are increasing the test times for whatever reason, in practice nobody looks at the time the test they have written takes to execute before they push it to CI, and so this is a form of early warning system. 

I don't think it is so controversial to say 'hey, each test should take no longer than X minutes on any reasonable machine'. You could also, I guess, also have it such that its just a warning when its not on the CI.

WRT counting the ticks on the network, I don't think that it is very useful. The number is mostly static and it is only really relevant to block sync etc., when you are sending messages that might not be strictly neccessary. Unperformant code (for example if you read from disk unneccessarily during a vote) causes the time each message takes to process to increase, which would not be captured by counting the ticks.

Interested to hear @theo-zil opinion. I don't mind closing the PR though.

theo-zil:
I think time taken is actually a functional requirement for us. If we're targeting, say, ~1 second block times, a test which requires the network to advance ten blocks should never ever take 10 minutes. The benefits of discouraging very long-running tests are obvious, but it's not just for CI ergonomics: if a PR starts making a test timeout, that can be considered an actual failure. (Provided that the limits set are generous enough to not count as false negatives. Again, for example, I cannot imagine a circumstance where a test that creates ten blocks would need more than 10 minutes to complete, on a blockchain targeting sub-second block times.)

As @n-hutton mentioned, ticks don't capture that because a) if we add non-performant code that won't necessarily increase the number of ticks, and b) the number of ticks is actually rather independent from the time taken, because for instance every `Vote` sent by every separate node to the current leader is a separate tick, whereas in a real network they would all get sent simultaneously.

JamesHinshelwood:
Fair enough, I can see the argument. Though in that case shouldn't the timeout be at the granularity of a single run with a seed, rather than for the entire test. Otherwise, setting too great a number of `ZQ_TEST_SAMPLES` is guaranteed to make the test fail, even though that is a useful operation (e.g. we might want to run a nightly job which tests `main` with a million samples or something)

theo-zil:
Good point, agree there. This would probably need to be done manually with a `select!` then.

PR #569, opened 2023-11-22
fix span messages in tests

n-hutton:
The comment should make it clear, but as I noted in my main forking PR, the index has to be recalculated otherwise it incorrectly appears as if the messages are all or mostly going to node 0.

### DIFF
--- a/zilliqa/tests/it/main.rs
+++ b/zilliqa/tests/it/main.rs
@@ -601,7 +601,15 @@ impl Network {
                 } else {
                     self.nodes.iter().collect()
                 };
-                for (index, node) in nodes.iter().enumerate() {
+
+                for node in &nodes {
+                    // Need to recalculate index against the original vec, not the filtered one.
+                    let index = self
+                        .nodes
+                        .iter()
+                        .position(|n| n.peer_id == node.peer_id)
+                        .unwrap();
+
                     let span = tracing::span!(tracing::Level::INFO, "handle_message", index);
                     span.in_scope(|| {
                         node.inner


JamesHinshelwood (approved):
My LOGAF is low, but I'd probably just handle the broadcast vs p2p messages separately, i.e.

```rust
AnyMessage::External(external_message) => {
    if let Some(destination) = message.1 {
        let node = self.nodes.iter().find(|n| n.peer_id == destination).unwrap();
        let span = tracing::span!(tracing::Level::INFO, "handle_message");
        span.in_scope(|| {
            node.inner.lock().unwrap().handle_network_message(message.0, external_message).unwrap();
        })
    } else {
        for (index, node) in nodes.iter().enumerate() {
            let span = tracing::span!(tracing::Level::INFO, "handle_broadcast", index);
            span.in_scope(|| {
                node.inner.lock().unwrap().handle_network_message(message.0, external_message.clone()).unwrap();
            });
        }
    }
}
```

n-hutton:
> My LOGAF is low, but I'd probably just handle the broadcast vs p2p messages separately, i.e.
> 
> ```rust
> ```

LOGAF, I like that...

Well, it doesn't make sense to do it like that IMO because for example there is a test currently I wrote that does something like 'send propose message only to node 0 and not anybody else'. So in the tests, the node index is distinct and important, especially when the behaviour is divergent. Other parts of the test framework correctly have the node index as part of the span, so it is misleading when you read later in the logs that node 0 got a message that was actually for node 2 according to a different span etc.

PR #570, opened 2023-11-22
Reduce iterations in tests

n-hutton:
image

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,12 +15,12 @@ env:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, gcp]
     steps:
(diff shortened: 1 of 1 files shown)


JamesHinshelwood:
Now that the time for `ZQ2 CI / check` is down to `~16m`, is the motivation for this change weaker?

n-hutton:
> Now that the time for `ZQ2 CI / check` is down to `~16m`, is the motivation for this change weaker?

Only slightly - the CI still bombs out I think if there are more than 2 PRs running in parallel?

JamesHinshelwood (changes requested):
I think we should resolve the CI problems, not reduce the number of tests to work around them.

n-hutton:
> I think we should resolve the CI problems, not reduce the number of tests to work around them.

Yes, that's fair. Closing

PR #572, opened 2023-11-22
Switch to using deterministic prbs in consensus

n-hutton:
When requesting blocks from the network during sync/consensus, it doesn't make sense to do this as a broadcast, as this is inefficient in both the test framework and IRL.

Rather, we pick a random validator to request blocks from. In order to keep the tests deterministic, we use a seeded PRNG.

image


zilliqa/src/time.rs
```diff
@@ -8,6 +8,7 @@ pub type SystemTime = std::time::SystemTime;
 
 pub type OffsetDateTime = time::OffsetDateTime;
 
+#[cfg(feature = "fake_time")]
```
n-hutton, line 11: This is needed in order to compile on my system sometimes. Not entirely sure why

JamesHinshelwood: It should be unnecessary now that #563 is merged.

zilliqa/src/consensus.rs
```diff
@@ -351,6 +355,7 @@ impl Consensus {
             state,
             db,
             new_transactions_priority: BTreeMap::new(),
+            rng: <rand_chacha::ChaCha8Rng as rand_core::SeedableRng>::seed_from_u64(0),
```
JamesHinshelwood, line 358: I worry that we'll use this rng to generate something sensitive down the line and regret having a very predictable seed. Seeding it from the secret key sounds like a good idea to me. Can we do that instead?

Also this should probably be `ChaCha20Rng` for the same reason?

zilliqa/src/consensus.rs
```diff
@@ -439,23 +444,30 @@ impl Consensus {
...
         let head_block = self.head_block();
 
-        let random_validator_peer = {
-            // Remove self from potential people to send to
-            let mut other_peers = head_block.committee.clone();
-            other_peers.remove_by_peer_id(self.peer_id());
-
-            match other_peers.len() {
-                0 => None,
-                _ => Some(other_peers.choose_random().peer_id),
```
JamesHinshelwood, line 449: Remove `choose_random` from `Committee`?

zilliqa/src/consensus.rs
```diff
@@ -439,23 +444,30 @@ impl Consensus {
...
+        let random_peer = self.get_random_other_peer();
         self.block_store
-            .request_blocks(random_validator_peer, head_block.header.number + 1)?;
+            .request_blocks(random_peer, head_block.header.number + 1)?;
 
         Ok(())
     }
 
+    pub fn get_random_other_peer(&mut self) -> Option<PeerId> {
+        // Remove self from potential people to send to
```
JamesHinshelwood, line 455: Can you do something like

```rust
use rand::seq::IteratorRandom;
...
self.head_block().committee.iter().filter(|v| v.peer_id != self.peer_id()).choose(&mut self.rng);
```
which will avoid cloning the committee?

The length check should also be unnecessary since `choose` will return `None` if the iterator is empty.

zilliqa/src/consensus.rs
```diff
@@ -568,6 +580,11 @@ impl Consensus {
                 if let Some(e) = e.downcast_ref::<MissingBlockError>() {
                     warn!("missing finalized block1");
                     info!(?e, "missing block when checking block proposal");
+
+                    let random_peer = self.get_random_other_peer();
+                    self.block_store
+                        .request_blocks(random_peer, block.header.number.saturating_sub(1))?;
```
JamesHinshelwood, line 586: Should we be requesting the block from the `BlockRef` in the `MissingBlockError`? Why are we sure that `block.header.number.saturating_sub(1)` is the missing one?

JamesHinshelwood (changes requested):

zilliqa/src/consensus.rs
n-hutton: Great suggestion, I have implemented it :

```
    pub fn get_random_other_peer(&mut self) -> Option<PeerId> {
        let our_id = self.peer_id();
        self.head_block()
            .committee
            .iter()
            .filter(|v| v.peer_id != our_id)
            .choose(&mut self.rng)
            .map(|v| v.peer_id)
    }
```

`our_id` is captured outside the filter to please the borrow checker.

zilliqa/src/consensus.rs
n-hutton: Most of the time this is correct, with edge cases being you are just joined/syncing. `request_blocks` also performs a check that the number is sane, and lowers it in the case it's above HEAD, so you should never request a block you clearly cannot attach to your chain.

zilliqa/src/consensus.rs
```diff
@@ -353,6 +358,10 @@ impl Consensus {
             state,
             db,
             new_transactions_priority: BTreeMap::new(),
+            // Seed the rng with the node's public key
+            rng: <rand_chacha::ChaCha8Rng as rand_core::SeedableRng>::seed_from_u64(
+                BigEndian::read_u64(secret_key.node_public_key().as_bytes().as_slice()),
```
JamesHinshelwood, line 363:
```suggestion
                u64::from_be_bytes(secret_key.node_public_key().as_bytes()[..8].try_into().unwrap()),
```
to avoid the extra dependency?

(copied from the implementation of `BigEndian::read_u64` :stuck_out_tongue:)

zilliqa/src/consensus.rs
```diff
@@ -204,6 +207,8 @@ pub struct Consensus {
     db: Arc<Db>,
     /// Transactions ordered by priority, map of address of TXn (from account) to ordered TXns to be executed.
     new_transactions_priority: BTreeMap<Address, BinaryHeap<TxnOrder>>,
+    // PRNG
+    rng: ChaCha8Rng,
```
JamesHinshelwood, line 211:
```suggestion
    rng: ChaCha20Rng,
```

zilliqa/src/consensus.rs
JamesHinshelwood: But `fn check_block` sometimes returns a `MissingBlockError` when the missing block isn't the `block`s parent. So doesn't this mean we'll end up unnecessarily requesting a block we already have and not requesting the block we're actually missing.

If we instead take the `BlockRef` out of `e`, then we can know exactly what was missing and request that?

zilliqa/src/consensus.rs
JamesHinshelwood: Sorry I didn't read this comment before making my suggestion. I was basing the suggestion of 20 rounds on:
> With the ChaCha algorithm it is possible to choose the number of rounds the core algorithm should run. The number of rounds is a tradeoff between performance and security, where 8 rounds is the minimum potentially secure configuration, and 20 rounds is widely used as a conservative choice.

from https://docs.rs/rand_chacha/latest/rand_chacha/struct.ChaCha8Rng.html

JamesHinshelwood: Ah sorry I didn't register the public key point. I don't think using the public key is sensible if our argument is that we want these random numbers to be unpredictable. What's wrong with using the private key?

zilliqa/src/consensus.rs
n-hutton: I didn't do it that way because if you allow someone to send you a block and you generate a network request on the specific missing information you could get into situations where you have two blocks that have missing refs to each other on the network and everything explodes into network requests. Or everyone requesting genesis's parent ref all the time or something.

Also, in the normal case where you join the network and start to sync, you will recieve a proposal from HEAD which may be very far in the future. You request the missing ref for HEAD-1 which makes you request the missing ref for HEAD-2... and so on, zipping down. This is super inefficient.

To make things as simplistic/minimalist as possible, I arranged block sync like follows:
- Only make a request if you have cause to believe you are missing information
- Only request from a single random peer
- This request ONLY refers to block number
  - block number should always reference a valid block (unlike view), unless it's above HEAD
  - the limited blocks number that are relevant to you to request are `> finalized block number`, and `<= HEAD + 1`
- The peer has the discresion to respond with any amount of blocks upward from this height
- If all the blocks you recieve from a request are valid and new, submit another request (zipping upwards)

Happy to discuss with you and @theo-zil whether you agree with these axioms or not - but it seems beyond the scope of this PR so I would suggest it's addressed in a follow up one

zilliqa/src/consensus.rs
n-hutton: Yes great, I should really have thought of this myself...

zilliqa/src/consensus.rs
theo-zil: I agree with you @n-hutton - specifically, the check in `block_store` to never request a block that's detached from your local chain ensures that the chain of requests for HEAD-1, HEAD-2... etc never happens.

But this raises another question: when would you ever want to request blocks at anything other than your current local highest_block_number? As I understand it, we're currently operating on the assumption that nodes will never need to request blocks earlier than its head. (For example, because every node holds a full database from genesis.) But at the same time, for aforementioned reasons you never want to request blocks above your head+1, either. So could we make `block_store.request_blocks()` take zero parameters, and have it effectively always just request a new batch of blocks starting from your head+1?

Happy to address this in a follow-up PR if you prefer. Or even migrate this discussion to slack if you want to merge this first.

n-hutton: Hi Theo, it is not correct that nodes will not request blocks earlier than head.

In the happy case, there will be two blocks at height N+1 and N+2, which are not finalized, while N is finalized. If you recieve a block proposal which has a missing parent at height N+2 (HEAD), you should request from other nodes the block N+1. In the non-happy case, the number of non-finalized blocks can be > 2 and so I believe you could be in a very extreme case requesting far before HEAD.

There is also another case, which is when you recieve a valid block proposal that you personally reject because the view is less than your current view, but it later gets accepted by the rest of the network. This happens because it is impossible to avoid view desync in certain scenarios. This is tested in the forking test.

n-hutton: I guess, now that I think about it, if we wanted to, it could be simplified in the way that you suggest by just requesting `finalized + 1`.

JamesHinshelwood (approved):

PR #574, opened 2023-11-22
Display for QC

n-hutton:
As in title

### DIFF
--- a/zilliqa/src/message.rs
+++ b/zilliqa/src/message.rs
@@ -245,6 +245,15 @@ impl QuorumCertificate {
     }
 }
 
+impl Display for QuorumCertificate {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "QC hash: {}, ", self.compute_hash())?;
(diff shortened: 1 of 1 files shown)


zilliqa/src/message.rs
```diff
@@ -245,6 +245,15 @@ impl QuorumCertificate {
     }
 }
 
+impl Display for QuorumCertificate {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "QC hash: {}, ", self.compute_hash())?;
+        write!(f, "QC signature: [..], ")?;
```
JamesHinshelwood, line 251: The signature is a fixed length - Is there a reason we don't display it?

JamesHinshelwood (approved):

zilliqa/src/message.rs
n-hutton: It doesn't really offer useful information IMO. And if you do display it, it fills your terminal up with stuff you likely don't care about. Debug will display it though ofc

PR #593, opened 2023-12-01
Add gas charging on revert, and gas_used tracking to receipts. Add tests for reverted transactions.

theo-zil:
One thing I'm not sure about is whether event logs should be kept on revert - I haven't checked what Eth clients do, and I'm not sure what ZQ1 does either in these cases. So I haven't touched them here, keeping the previous behaviour of keeping them available.

### DIFF
--- a/zilliqa/src/exec.rs
+++ b/zilliqa/src/exec.rs
@@ -101,6 +101,8 @@ pub struct TransactionApplyResult {
     pub logs: Vec<EvmProto::Log>,
     /// The traces from the EVM, if enabled
     pub traces: Arc<Mutex<LoggingEventListener>>,
+    /// The gas paid by the transaction
+    pub gas_used: u64,
 }
 
 impl State {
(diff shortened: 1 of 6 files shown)


JamesHinshelwood:
This is on my review list for tomorrow :) Just wanted to mention re logs: There is a `removed` field in the `logs` response of `eth_getTranasctionReceipt`, documented as 
> true when the log was removed, due to a chain reorganization. false if its a valid log.

zilliqa/src/exec.rs
```diff
@@ -444,11 +446,23 @@ impl State {
             }
         }
 
+        // If this was contract creation, apply this to the deltas for the convenience of
```
JamesHinshelwood, line 449: I'm confused how this would only be for the convenience of the caller? Would something later result in this happening anyway?

zilliqa/tests/it/eth.rs
```diff
@@ -686,6 +687,116 @@ async fn eth_call(mut network: Network) {
...
+    let contract_address = receipt.contract_address.unwrap();
+    let setter = abi.function("revertable").unwrap();
+    let getter = abi.function("value").unwrap();
+
+    // First ensure contract works
+    let success_call = TransactionRequest::new()
+        .to(contract_address)
+        .data(setter.encode_input(&[Token::Bool(true)]).unwrap());
+    let (_, receipt) = send_transaction(&mut network, success_call.into()).await;
+    assert_eq!(receipt.status.unwrap().as_u32(), 1);
```
JamesHinshelwood, line 712: Maybe we can assert that `receipt.gas_used` is non-zero here and below too?

JamesHinshelwood: Oh ignore this, I just read the second test :)

JamesHinshelwood (approved):

zilliqa/src/exec.rs
theo-zil: Good question, this comment was already there in the code so I'm not sure. My best guess is that it reduces the need to special case contract creation in the layer above, by letting most of it be handled by `apply_deltas()` rather than needing dedicated logic?
@n-hutton might know more about this.

PR #594, opened 2023-12-04
Add flag to run test suite in a 'summary' mode

JamesHinshelwood:
When running the test suite with `ZQ_TEST_SEED_SUMMARY` set, the test suite will not fail fast and keep running if a seed fails. The failing seeds will not be printed - Instead, a summary with the total number of passes and failures will be printed.

My intention was that this could be run with a high seed count, outside of the normal PR process. This would let us tell if our suite is flaky and, if so, how flaky.


theo-zil (commented):
Very neat idea!

What about displaying the actual failed seeds if there's few enough of them? Especially for something like the current forking test where 0.19% of seeds fail, sometimes it might be useful to have more than one failing seed to debug and compare what happens in different failure cases (to try to find a common cause, for example, or conversely to ensure you've caught every failure mode if there's multiple).

I think it'd be useful to store either, say, the first 0.5% of failures, or some constant like the first 20 or 50 failures, and print out their seeds. This way it doesn't flood the output with dozens or hundreds of seeds, but still gives you a selection of test cases to investigate; and if a test is failing much more often than that then finding more failing seeds should be relatively easy.

n-hutton (changes requested):
I think this is good, but how does the progress bar display on your system? Mine seems to display test iterations as a time which seems incorrect

JamesHinshelwood:
@n-hutton Well spotted, I messed up the style format. It should be better now. On my system it looks like:
```
[00:02:32/00:04:49] #######################################################----------------------------------------------- 5295/10000
```

JamesHinshelwood:
Great idea @theo-zil , thank you. We now print the first 16 failed seeds e.g.:
```
---- consensus::handle_forking_correctly stdout ----
Total seeds tested: 100000
Success: 99875
Failure: 125 (seeds: 2309785970321532039,1864881025440800072,5480424727184238143,2041668194461047195,14864211872815470318,17315898059188977203,1187597600032925948,10413544769819499832,608473593712743793,1005440416005636688,7485729435067507308,14156993530512487444,14092886448668307691,10435364861596307405,3055291945807100457,16014147235183226002...)
Pass rate: 99.875%
```

n-hutton (approved):

PR #597, opened 2023-12-04
Reduce test times

n-hutton:


JamesHinshelwood (commented):
I think it makes sense to add `--release` to `cargo build` and `cargo clippy` too. We're building everything in release mode anyway, so there's no point building in debug mode for those two commands earlier.

n-hutton:
> I think it makes sense to add `--release` to `cargo build` and `cargo clippy` too. We're building everything in release mode anyway, so there's no point building in debug mode for those two commands earlier.

I think that my LOGAF level is low, but the changes by building in release are: less space used on disk, takes significantly longer to compile. 

I would say that longer compilation times/longer CI is less optimal.

JamesHinshelwood:
I think `--release` everywhere should be objectively faster.

Running `cargo test --release` requires you to build everything in release mode anyway. Running `cargo build` (without `--release`) before this means you will also spend time building everything in debug mode.

In other words (`<` means is quicker than):

- `cargo build` < `cargo build --release`
- `cargo test` < `cargo test --release`
- `cargo build` + `cargo test --release` > `cargo build --release` + `cargo test --release`

(at least as far as I understand)

JamesHinshelwood (approved):

PR #601, opened 2023-12-05
allow seed in spans

n-hutton:
example `ZQ_TEST_SEED_SPANS=1 RUST_LOG=zilliqa=trace,it=trace cargo test  --release --test it -- consensus::launch_shard`

Produces:

```
2023-12-05T20:01:53.776561Z TRACE {seed=16741818491380055426}:handle_message{index=12}: zilliqa::consensus: head block request: highest block number: 4
2023-12-05T20:01:53.776567Z TRACE {seed=16741818491380055426}:handle_message{index=12}: zilliqa::consensus: handling block proposal de38f21300ab3f2d97787efabdc41468c08840b035d2a8362df0d59c6a87fb36 block_view=5 block_number=5
2023-12-05T20:01:53.777132Z TRACE {seed=16741818491380055426}:handle_message{index=12}: zilliqa::consensus: (check block) I think the block proposer is: 12D3KooWKnS5uGQ1X3KAp95PA65CtEvkNdTWVBtrxrwLhhJUtFYr, we are 12D3KooWDp2EZLhPeH58WebRCFkd8mN5EGFsH1SU2jrtv3zyMG77
```

Useful to turn on when you have a test which only fails when multiple are running at the same time (race conditions).

### DIFF
--- a/zilliqa-macros/src/test.rs
+++ b/zilliqa-macros/src/test.rs
@@ -55,22 +55,33 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
                     let subscriber = tracing_subscriber::fmt()
                         .with_ansi(false)
                         .with_env_filter(tracing_subscriber::EnvFilter::from_default_env());
+
                     let _guard = tracing_subscriber::util::SubscriberInitExt::set_default(subscriber);
 
-                    let mut rng = <rand_chacha::ChaCha8Rng as rand_core::SeedableRng>::seed_from_u64(seed);
-                    let network = crate::Network::new(std::sync::Arc::new(std::sync::Mutex::new(rng)), 4, seed);
+                    // Optionally print the seed in each tests logging so it can be filtered out later.
+                    use tracing::Instrument;
+                    let span = if std::env::var_os("ZQ_TEST_SEED_SPANS").is_some() {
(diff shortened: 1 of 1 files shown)


zilliqa-macros/src/test.rs
```diff
@@ -55,22 +55,33 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
...
-                        Ok(()) => {},
-                        Err(e) => {
-                            std::panic::resume_unwind(e);
+                    // Optionally print the seed in each tests logging so it can be filtered out later.
+                    use tracing::Instrument;
+                    let span = if std::env::var_os("ZQ_TEST_SEED_SPANS").is_some() {
+                        tracing::span!(tracing::Level::INFO, "", seed)
+                    } else {
+                        tracing::span!(tracing::Level::INFO, "")
+                    };
```
JamesHinshelwood, line 67: Any reason not to just print the seed unconditionally? If I understand correctly, this will only be printed when you have `RUST_LOG=it=info` anyway?

Also should we give the span a name? e.g. `"test"`

n-hutton: The seeds are long and ugly so I don't think you would usually want them. Likewise for naming the test span - I did have this before but it doesn't add anything. If you put the network construction in the span too then every single trace message would have 'test' - which adds 0 net information.

PR #602, opened 2023-12-05
Flush db to disk before copying it over in test

n-hutton:
The test `consensus::network_can_die_restart` had a race condition where when it kills the nodes, their DBs are in a state of writing. This caused it to sometimes randomly fail. This forcibly flushes them so that the tests are deterministic and pass once again.

### DIFF
--- a/zilliqa/tests/it/main.rs
+++ b/zilliqa/tests/it/main.rs
@@ -322,9 +322,9 @@ impl Network {
         );
         let genesis_committee = vec![validator];
 
-        //for nodes in &mut self.nodes {
-        //    nodes.inner.lock().unwrap().db.flush();
-        //}
+        for nodes in &mut self.nodes {
+            nodes.inner.lock().unwrap().db.flush();
+        }
 
         let (nodes, external_receivers, local_receivers): (Vec<_>, Vec<_>, Vec<_>) = keys
             .into_iter()
--- a/zilliqa/tests/it/main.rs
+++ b/zilliqa/tests/it/main.rs
@@ -5,6 +5,7 @@ mod web3;
 mod zil;
 use std::{env, ops::DerefMut};
 
+//use crate::time::SystemTime;
 use ethers::solc::SHANGHAI_SOLC;
 use itertools::Itertools;
 use serde::Deserialize;
@@ -321,6 +322,10 @@ impl Network {
             keys[0].to_libp2p_keypair().public().to_peer_id(),
         );
         let genesis_committee = vec![validator];
+
+        for nodes in &mut self.nodes {
+            nodes.inner.lock().unwrap().flush();
+        }
 
         let (nodes, external_receivers, local_receivers): (Vec<_>, Vec<_>, Vec<_>) = keys
             .into_iter()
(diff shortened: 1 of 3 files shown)


zilliqa/tests/it/main.rs
```diff
@@ -322,6 +323,10 @@ impl Network {
         );
         let genesis_committee = vec![validator];
 
+        for nodes in &mut self.nodes {
+            nodes.inner.lock().unwrap().flush();
+        }
```
JamesHinshelwood, line 328: Hmm, having to remember to flush the DB manually seems somewhat wrong IMO. `Db` will flush to disk in its `Drop` impl already, so I think the right thing to do is drop the old nodes, before we copy the data and create the new ones?

Especially because I don't think this implementation is correct. `fn flush` returns the number of bytes it managed to flush and you need to call it in a loop:
https://docs.rs/sled/latest/src/sled/tree.rs.html#98

JamesHinshelwood: cc @theo-zil since I remember there were some caveats around flushing/dropping with sled that you encountered before.

n-hutton: I don't think deleting the nodes first will work well. The db will indeed flush to disk, but the underlying tempdir object that the node holds will also get released, creating a race condition where the OS is likely to clean the dir up as you are trying to copy it over.

Wrt the drop - that's a good spot, it does seem(?) this use is incorrect. The documentation for flush is super confusing, it could well be interpreted that this is correct which ever way I read it:

```
    /// ...If this succeeds, it is guaranteed that all
    /// previous writes will be recovered if the system
    /// crashes.
 ```

https://docs.rs/sled/latest/src/sled/tree.rs.html#788

Comment (theo-zil) [reply zilliqa/tests/it/main.rs:328]: You can just grab the tempdir alone, see for example tests/it/persistence.rs:
```
#[allow(clippy::redundant_closure_call)]
let dir = (|mut node: TestNode| node.dir.take())(node).unwrap(); // move dir out and drop the rest of node
```

The problem with dropping that I was having with the persistence test was that for some reason the Arc<Node> wasn't releasing the underlying Node object when dropped, so the `Drop` wasn't getting called in the first place. I never saw any issues with the db flushing and closing properly _if_ it gets actually dropped.

JamesHinshelwood: I think that change makes more sense because it keeps the changes specific to the test suite, rather than requiring the `flush` methods in our code, which shouldn't really be needed. :+1:

n-hutton: I spent some time trying to replicate this but you simply cannot rely on drop in this current codebase. At the point in the code here that you try to drop it, `node` has 8 strong references, and `db` has 3. I am very confused as to what is holding these references... in any case, I think the implementation as I have it currently is the best it could get - having a flush method on a db class is nothing extreme.

theo-zil: That sounds similar to the issue that was causing the persistence test to fail... Odd because I had no idea either where the references were coming from (and still don't).

Flushing is nothing extreme but with RAII, in an ideal world, it would never have to be called manually. But given this weirdness it's a fair workaround

theo-zil (dismissed):

JamesHinshelwood (approved):

PR #603, opened 2023-12-05
time bound the tests

n-hutton:
as discussed in a prev PR - it simply panics at the end if it took too long all in all.

### DIFF
--- a/zilliqa-macros/src/test.rs
+++ b/zilliqa-macros/src/test.rs
@@ -142,7 +142,7 @@ pub(crate) fn test_macro(_args: TokenStream, item: TokenStream) -> TokenStream {
             }
 
             let duration = start.elapsed();
-            let mut time_allowed_ms = Duration::from_millis(1000).as_millis() * (seeds_number as u128);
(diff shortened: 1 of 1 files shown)


JamesHinshelwood (commented):
If I'm reading correctly, this gives 10 seconds to each seed (or 100 seconds in debug mode)?

How long does the average seed take currently? My gut feeling is that 10 seconds is multiple orders of magnitude too slow for a single test run.

n-hutton:
> If I'm reading correctly, this gives 10 seconds to each seed (or 100 seconds in debug mode)?
> 
> How long does the average seed take currently? My gut feeling is that 10 seconds is multiple orders of magnitude too slow for a single test run.

Hi, I didn't set it to anything that would stress it. I have now set it to 1s per seed - that passes reliably on my system, but lower starts to cause it to fail. Wdyt?

JamesHinshelwood (approved):
1s/10s still seems horribly slow to me if you aren't running these tests on a toaster. But I guess we can look into that later.

PR #604, opened 2023-12-05
Launch shard fix

n-hutton:
In the launch shard test, a bug could occasionally happen. This was a race condition where you would verify your proposal before the tx contained within it ever got gossiped to you.

Later, if someone asks you to sync your chain, you cannot find the corresponding TX for that block since you didn't commit it.

This makes the TX check and add part of the same function so that proposals will always make sure to commit the TX.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,7 @@ jobs:
     - name: Clippy
       run: cargo clippy --all-targets --all-features -- -D warnings
     - name: Test
-      run: cargo test --release --all-targets --all-features
+      run: cargo test --all-targets --all-features
       timeout-minutes: 30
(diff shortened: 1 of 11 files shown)


JamesHinshelwood:
> This was a race condition where you would verify your proposal before the tx contained within it ever got gossiped to you.

How is this possible? The proposal contains the transactions in the proposed block?

n-hutton:
> > This was a race condition where you would verify your proposal before the tx contained within it ever got gossiped to you.
> 
> How is this possible? The proposal contains the transactions in the proposed block?

This is possible because the logic was flawed in that the way TXs enter the db was by calling `remove_tx_from_mempool` - if the tx was/is not in the mempool then it would not write it to disk.

PR #606, opened 2023-12-05
enable bls

n-hutton:


zilliqa/src/message.rs
```diff
@@ -87,7 +87,7 @@ impl Vote {
     ) -> Self {
         let mut bytes = Vec::new();
         bytes.extend_from_slice(block_hash.as_bytes());
-        bytes.extend_from_slice(&view.to_be_bytes());
+        //bytes.extend_from_slice(&view.to_be_bytes());
```
JamesHinshelwood, line 90: Why did we remove the view from vote signatures?

n-hutton: when you enable BLS everything needs to be signed on the same stuff essentially (which it isn't in main). This PR contains the full fix: https://github.com/Zilliqa/zq2/pull/607

PR #607, opened 2023-12-05
Enable bls, but also sign on the view

n-hutton:
enable bls, but make sure to also sign against the view

### DIFF
--- a/zilliqa/src/message.rs
+++ b/zilliqa/src/message.rs
@@ -249,15 +249,19 @@ impl QuorumCertificate {
     // against the aggregated signature
     pub fn verify(&self, _public_keys: Vec<NodePublicKey>) -> bool {
         // Select which public keys have gone into creating the signature
-        let public_keys = _public_keys.into_iter().zip(self.cosigned.iter()).filter_map(
-            |(public_key, cosigned)| {
-                if *cosigned {
-                    Some(public_key)
-                } else {
-                    None
-                }
-            },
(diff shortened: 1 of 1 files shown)


zilliqa/src/message.rs
```diff
@@ -248,7 +248,26 @@ impl QuorumCertificate {
     // Verifying an aggregated signature is a case of verifying the aggregated public key
     // against the aggregated signature
     pub fn verify(&self, _public_keys: Vec<NodePublicKey>) -> bool {
-        true
+        // Select which public keys have gone into creating the signature
+        let public_keys = _public_keys
```
JamesHinshelwood, line 252: nit: We can rename `_public_keys` to `public_keys` now it is used.

zilliqa/src/message.rs
```diff
@@ -248,7 +248,26 @@ impl QuorumCertificate {
...
+            .filter_map(
+                |(public_key, cosigned)| {
+                    if *cosigned {
+                        Some(public_key)
+                    } else {
+                        None
+                    }
+                },
+            )
+            .collect::<Vec<_>>();
```
JamesHinshelwood, line 264: nit: We can avoid some allocations by passing the `impl Iterator` directly to `verify_aggregate` (maybe in a later PR)

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #608, opened 2023-12-09
make subscriber respect RUST_LOG

n-hutton:
`tracing_subscriber::fmt::init();` and `tracing_subscriber::fmt().init();` do not behave the same - the first will respect RUST_LOG and the second will not


JamesHinshelwood (commented):
Oh good spot, thanks. Though we should also respect `RUST_LOG` when `log_json` is `true`...

JamesHinshelwood (changes requested):

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #647, opened 2024-01-08
clear memory in consensus

n-hutton:
The borrow checker did not like the snippet as it originally was:

```
        self.votes
            .retain(|k, _| {
                match self.get_block(k) {
                    Ok(Some(block)) => block.view() >= self.view.get_view(),
                    _ => { warn!("Missing block for vote (this shouldn't happen), removing from memory"); false },
                }
            });
```

Because the use of self inside the closure borrows it when it is already being borrowed for `self.votes`. So I rewrote it slightly longer.


zilliqa/src/consensus.rs
```diff
@@ -878,6 +879,26 @@ impl Consensus {
         ret
     }
 
+    /// Clear up anything in memory that is no longer required. This is to avoid memory leaks.
```
theo-zil, line 882: I wouldn't really call it a "memory leak" - we were just explicitly and intentionally storing things that we didn't need any longer. (In my mind, a leak is when memory is kept allocated accidentally or unintentionally.)

Correspondingly I'd also name it something like `cleanup_votes` rather than memory.

theo-zil (changes requested):
I think this matches what we want to do - minor comment on the naming though I don't consider it a blocker.

Only change requested is that the comment on lines 1004-1005 is made obsolete by this change and should be removed. (At the end of `fn vote()`.)

zilliqa/src/consensus.rs
n-hutton: Ah yes, it is not technically a memory leak :) - okay, I will rename it!

PR #656, opened 2024-01-10
Scilla integration

n-hutton:
It is commented extensively inline, I will put the high level here. This is a big pr and I am happy to try to split it up. When reviewing, just do the broad strokes for now because there is still some cleanup to do.

There is a PR here which puts scilla inside a docker container, and opens three TCP ports. These are used for sending the scilla commands, for the scilla server to read/write state, and for writing the input files.

I chose to do it this way because historically the IPC has been buggy, especially on mac, with random bugfixes being required to the jsonrpccpp library. With tcp it is also a lot easier to capture the traffic when debugging.

The scilla process will get spawned every time the tcp connection for commands is opened, and it gets killed when the connection closes. This means each call is sandboxed to some degree.

https://github.com/Zilliqa/scilla/pull/1244

Some work needs to be done to make a proper PR to master that includes the B64 changes plus these scripts. If there is interest I will explain how I did this.

- Transaction payloads are now (data, init_data) to accomodate scilla contract creation
- Some structures now additionally carry scilla related info, like events
- The execution of a scilla tx takes the same backend structure as the evm
- The return deltas of a scilla is the same as the evm
- In order to fit the scilla data which is String, Vec<u8> for key value to U256 U256 a scheme is used where they key is hashed, and the value length is written to that location. Subsequent hash locations contain the value bytes
- The current code successfully deployes and executes a scilla transaction
- generally the scilla invokation should panic if something unexpected happens like being unable to connect to the server, and error if it looks like something about the TX was malformed
- Both base64 mode for the scilla state read/writes and non are supported since I initially was working without base64

Comments on the details are inline.


scilla/src/backend_collector.rs
```diff
@@ -0,0 +1,212 @@
...
+
+use evm::backend::Backend;
+use evm_ds::protos::evm_proto::{Apply, EvmResult, Storage};
+use primitive_types::{H160, H256, U256};
+use sha3::{Digest, Keccak256};
+use tracing::{*};
+
+pub type Address = H160;
+
+#[derive(Debug, Clone)]
```
n-hutton, line 11: This has been copied from the main repo to avoid the scilla library depending on the zq2 repo, thus creating a circular dependency.

evm-ds/src/protos/evm_proto.rs
```diff
@@ -65,6 +65,7 @@ pub struct EvmCallArgs {
     pub enable_cps: bool,
     pub tx_trace_enabled: bool,
     pub tx_trace: Arc<Mutex<LoggingEventListener>>,
+    pub is_scilla: bool,
```
n-hutton, line 68: In terms of imports, scilla refers in limited cases to evm or evm-ds, but not to zilliqa. This is to avoid circular dependencies

scilla/src/call_scilla_server.rs
```diff
@@ -0,0 +1,291 @@
...
+    fn new(method: &str, params: Params, id: u32) -> Self {
+        JsonRpcRequest {
+            jsonrpc: "2.0".to_string(),
+            method: method.to_string(),
+            params,
+            id,
+        }
+    }
+}
+#[derive(Serialize, Deserialize, Debug)]
```
n-hutton, line 63: This file generally handles making the json rpc calls to the scilla server. Unf the jsonrpc rust library did not support bare tcp connections, so I did it manually. The scilla server is a bit odd in that the result is sometimes a string which contains a json object?

It was also, as far as I could tell, impossible to cleanly use this pattern below for attaching methods. This is because the backend passed to scilla has to be a reference (it is constructed in exec.rs as a reference to self). You can't even wrap it in an Arc because the compiler knows that the reference inside could potentially outlive the closure you are passing to add_method. I would be interested to hear if it is actually possible other than doing something like using channels

```
        let mut io = IoHandler::new();
        io.add_method("fetchStateValueB64", {
            let inner = Arc::clone(&inner);
            move |params| inner.lock().unwrap().fetch_state_value_b64(params)
        });
```

scilla/src/scilla_tcp_server.rs
```diff
@@ -0,0 +1,318 @@
...
+}
+
+impl<'a, B: evm::backend::Backend> ScillaServer<'a, B> {
+    pub fn handle_request(
+        &mut self,
+        request: &JsonRpcRequest,
+    ) -> Result<Value, jsonrpc_core::Error> {
+
+        match request.method.as_str() {
+            //"fetchStateValueB64" => {
```
n-hutton, line 40: Todo

zilliqa/Cargo.toml
```diff
@@ -33,10 +33,11 @@ bs58 = "0.5.0"
 bytes = "1.5.0"
 chrono = "0.4"
 clap = { version = "4.4.13", features = ["derive"] }
-crypto-bigint = "=0.5.5"
+crypto-bigint = "=0.5.0"
```
n-hutton, line 36: by the way, version 0.5.5 of crypto-bigint breaks my editor which is why I keep reverting it. Since it is a minor version it would be nice if it could be frozen for now

zilliqa/src/consensus.rs
```diff
@@ -1029,7 +1026,7 @@ impl Consensus {
 
         committee.add_validators(self.pending_peers.drain(..));
 
-        if committee.len() <= 1 {
+        if false {
```
n-hutton, line 1029: will revert for merge to main

zilliqa/src/consensus.rs
```diff
@@ -645,6 +645,7 @@ impl Consensus {
                         contract_address: result.contract_address,
                         logs: result.logs,
                         gas_used: result.gas_used,
+                        scilla_events: serde_json::to_string(&result.scilla_events).unwrap(),
```
n-hutton, line 648: Unfortunetly, the TransactionReceipt needs to have a concrete type since it gets written down to disk. Using a serde_json::Value doesn't work as it doesn't know how much space to allocate it essentially. So it gets set to a string here. There might well be a way to get around this more cleanly

zilliqa/src/exec.rs
```diff
@@ -553,17 +588,35 @@ impl State {
...
+            _ => {}
+        }
 
         let txn = txn.tx.into_transaction();
 
         // fail early on nonce mismatch
         let mut acct = self.get_account(from_addr).unwrap();
-        if acct.nonce != txn.nonce() {
+
+        let desired_account_nonce = if is_scilla {
```
n-hutton, line 606: The js framework seems to be sending a tx starting from nonce 0 for some reason, need to investigate...

zilliqa/src/api/zil.rs
```diff
@@ -103,10 +112,70 @@ fn create_transaction(params: Params, node: &Arc<Mutex<Node>>) -> Result<serde_j
...
+}
+
+fn get_transaction(params: Params, node: &Arc<Mutex<Node>>) -> Result<Option<GetTxResponse>> {
+    let hash: H256 = params.one()?;
+    let hash: Hash = Hash(hash.0);
+
+    let ret = get_scilla_transaction_inner(hash, &node.lock().unwrap())?;
+    let mut receipt = node.lock().unwrap().get_transaction_receipt(hash)?;
+
+    // Spin up to 1s while waiting for the receipt to become available
```
n-hutton, line 138: There is some issue with the JS framework/scilla plugin where it attempts to immediately query the TX receipt after submitting the TX. Responding null does not work. This causes the test to erronerously fail - I am looking into it.

scilla/src/call_scilla_server.rs
JamesHinshelwood: Do you have a version of this in git somewhere? I'd be curious to know the error message when you use the `Arc` as in your snippet above.

scilla/src/call_scilla_server.rs
```diff
@@ -0,0 +1,275 @@
...
+            }
+            // It is valid if there is no data to read for either connection because it can still be
+            // processing the request.
+            Err(_) => {
+                //debug!("Scilla backend error: {:?}", e);
+            }
+        }
+
+        // Check if final character read was a newline, which indicates the end of the response
+        if bytes_read_backend > 0 && response_backend[bytes_read_backend - 1] == b'\n' {
```
JamesHinshelwood, line 149: This is all quite scary and manual to me, but maybe we can live with it until we reach "Scilla interface optimisation".. :)

zilliqa/src/api/types/zil.rs
```diff
@@ -67,6 +72,74 @@ struct TxBlockHeader {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    event_logs: Option<Vec<Value>>,
+}
+
+impl GetTxResponse {
+    pub fn new(verified_tx: VerifiedTransaction, receipt: TransactionReceipt) -> Option<Self> {
+        match verified_tx.tx {
+            SignedTransaction::Zilliqa { ref tx, .. } => Some(GetTxResponse {
+                id: verified_tx.hash.to_string(),
+                version: "65537".to_string(),
```
JamesHinshelwood, line 117: Doesn't this need to consider the chain ID?

zilliqa/src/api/types/zil.rs
```diff
@@ -67,6 +72,74 @@ struct TxBlockHeader {
...
+                    epoch_num: "1".to_string(), // todo here
+                    success: receipt.success,
+                    event_logs: Some(serde_json::from_str(&receipt.scilla_events).unwrap()),
+                },
+                gas_price: "2000000000".to_string(),
+                gas_limit: "50000".to_string(),
+                code: tx.code.to_string(),
+                data: tx.data.to_string(),
+            }),
+            _ => None,
```
JamesHinshelwood, line 138: IIRC, this API does work in ZQ1 for non-Zilliqa transactions?

zilliqa/src/api/zil.rs
```diff
@@ -103,10 +112,83 @@ fn create_transaction(params: Params, node: &Arc<Mutex<Node>>) -> Result<serde_j
...
     };
 
-    let transaction_hash = node.create_transaction(transaction)?;
-    let transaction_hash = hex::encode(transaction_hash.0);
+    let transaction_hash = node.create_transaction(transaction.clone())?;
+    let transaction_verified: VerifiedTransaction = transaction.verify()?;
 
-    Ok(json!({"TranID": transaction_hash}))
+    match transaction_verified.tx {
+        SignedTransaction::Zilliqa { ref tx, .. } => {
```
JamesHinshelwood, line 119: This smells a bit to me. You create a `TxZilliqa` above, wrap it into a `SignedTransaction`, then unconditionally match on it to get the `TxZilliqa` back again later.

Can we move the `contract_address` calculation to before the call to `node.create_transaction` and then remove any matching on the transaction here?

zilliqa/src/api/zil.rs
JamesHinshelwood: I'd feel quite uncomfortable adding this code to `main` since it seems like a serious performance problem waiting to happen.

We should check the behaviour of the ZQ1 API when this method is called for a transaction that hasn't committed yet.

zilliqa/src/consensus.rs
JamesHinshelwood: I don't understand why `serde_json::Value` doesn't work here? It represents things internally with dynamic allocation so it shouldn't need to know how much space it allocates in advance.

zilliqa/src/consensus.rs
JamesHinshelwood: I'd be in favour of just removing this `if` block. There's no reason a network with 1 validator shouldn't work fine (in fact it does, with this log message being the only annoyance).

zilliqa/src/exec.rs
```diff
@@ -103,6 +105,8 @@ pub struct TransactionApplyResult {
     pub traces: Arc<Mutex<LoggingEventListener>>,
     /// The gas paid by the transaction
     pub gas_used: u64,
+    /// The scilla events, if any
+    pub scilla_events: Value,
```
JamesHinshelwood, line 109: Should we attempt to decode the scilla events, rather than just leaving them as a JSON `Value`? IIRC we need to transform each one to a special representation when returning them from `eth_getLogs` so it might be easier to store something more structured internally?

zilliqa/src/exec.rs
```diff
@@ -180,20 +186,25 @@ impl State {
...
+        let mut data: Vec<u8> = payload.clone();
         let mut traces: Arc<Mutex<LoggingEventListener>> =
             Arc::new(Mutex::new(LoggingEventListener::new(tracing)));
 
         // The backend is provided to the evm as a way to read accounts and state during execution
         let mut backend = EvmBackend::new(self, U256::zero(), caller, chain_id, current_block);
 
         // if this is none, it is contract creation
-        if to_addr.is_none() {
+        if to_addr.is_none() || to_addr.is_some_and(|x| x == Address::zero()) {
```
JamesHinshelwood, line 197: This isn't correct. In Ethereum-land you are allowed to send a transaction to the zero address and it means something different to a transaction without an address.

JamesHinshelwood (commented):
I haven't done a complete review, but there are probably enough comments worthy of discussion that its worth submitting now :)

zilliqa/src/exec.rs
n-hutton: The correct way to create a contract in eth is to send a zero length byte array for the 'to'. Thus it should not be Address::zero(), that is correct. However, I believe that certain clients will send the zero address to indicate contract creation (including scilla). Would you agree?

This isn't right though as you point out, because someone might be trying to burn tokens, and there is not neccessarily data with the tx.

Do you think that when the scilla TX is received, we should map the zero address to Option<None>?

zilliqa/src/exec.rs
n-hutton: I could map them into eth `Log` object if that is what you mean? Is this for otterscan? I had assumed that scilla TXs are assumed not to have logs via the eth api

zilliqa/src/api/zil.rs
n-hutton: yes, it is smelly - I will rewrite it. The issue is that you need the nonce and the from address, so you neccessarily requre constructing a verified transaction, but it requires a match on that to get the nonce. However, I could just copy the nonce from earlier in the code.

zilliqa/src/api/types/zil.rs
n-hutton: I don't recall - I will add support for it in any case. On an unrelated note, the return structs from z1 api are a random mix of camel snake and pascal case...

zilliqa/src/api/types/zil.rs
n-hutton: That is a good point

scilla/src/call_scilla_server.rs
n-hutton: Yes, I agree with you on that. It can definetly be improved, but for the moment most of the development work is not on the interface itself (but rather feature completeness).

I realised later I should have upgraded the scilla connection to http rather than tcp, then the jsonrpc framework would support it.

scilla/src/backend_collector.rs
```diff
@@ -0,0 +1,365 @@
...
+pub struct Account {
+    pub nonce: u64,
+    pub code: Vec<u8>,
+    pub storage_root: Option<H256>,
+    pub is_scilla: bool,
+}
+
+// Structure that answers queries about the state by using the backend, while also collecting
+// the state changes so it can generate an evm result
+pub struct BackendCollector<'a, B: evm::backend::Backend> {
```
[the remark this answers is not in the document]
n-hutton: They do share the same backend in exec.rs

```
            result = if call_args.is_scilla {
                run_scilla_impl_direct(call_args.clone(), &backend)
            } else {
                run_evm_impl_direct(call_args.clone(), &backend)
            };

            if print_enabled {
                debug!("Evm invocation complete - applying result {:?}", result);
            }

            // Apply the results to the backend so they can be used in the next continuation
            backend.apply_update(result.take_apply());
```

Perhaps it would help to rename the scilla one, its purpose is more for caching and compress/decompress. Even if you used the same structure, the scilla one needs additional functionality on top (plus it would be a circular reference). I have tried to keep any ugliness in the scilla side...

scilla/src/backend_collector.rs
```diff
@@ -0,0 +1,365 @@
...
+            // Now we can write the key using the compression scheme
+            self.write_compressed(address, key_pointer, key.as_bytes());
+        }
+
+        // Write the key normally using the compression scheme
+        self.write_compressed(address, key_as_hash, value);
+    }
+
+    /// Internal function to write a compressed value to the database, according
+    /// to the scheme described in update_account_storage_scilla
```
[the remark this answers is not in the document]
n-hutton: It will take more space, that is true. What would you prefer? `translate`, `map`?

scilla/src/backend_collector.rs
```diff
@@ -0,0 +1,365 @@
...
+
+use evm::backend::Backend;
+use evm_ds::protos::evm_proto::{Apply, EvmResult, Storage};
+use primitive_types::{H160, H256, U256};
+use serde_json::Value;
+use sha3::{Digest, Keccak256};
+use tracing::*;
+
+pub type Address = H160;
+type AccountStorage = HashMap<H256, H256>;
```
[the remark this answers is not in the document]
n-hutton: I would argue (but certainly up for debate), that you can essentially avoid doing this by using the mapping algorithm here (previously called `compression`) and essentially point it at storage to provide an abstraction where the caller sees it as `(String, String)`. You know whether you need to do this because the `Account` flags whether it is scilla or not.

WRT key value pairs, the tests I am running require them, and so there is a scheme to recover the original keys, too. it is called in `node.rs` as `get_scilla_kv_pairs` (not sure if that's the best name for it).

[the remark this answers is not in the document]
n-hutton: I don't know what your solution is for the z1 storage, but you should also be able to map the z1 storage to this storage using this scheme and everything should work as before.

scilla/src/backend_collector.rs
```diff
@@ -0,0 +1,365 @@
...
+}
+
+// Structure that answers queries about the state by using the backend, while also collecting
+// the state changes so it can generate an evm result
+pub struct BackendCollector<'a, B: evm::backend::Backend> {
+    // Reference to the original backend
+    pub backend: &'a B,
+    // Map of cached (execution in progress) address to account and any dirty storage.
+    // If the value is None, this means a deletion of that account and storage
+    pub account_storage_cached: HashMap<Address, Option<(Account, AccountStorage)>>,
```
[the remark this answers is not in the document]
n-hutton: Ah yes I guess it is acting as a buffer and a cache... I will change it to buffer.

scilla/src/backend_collector.rs
```diff
@@ -0,0 +1,365 @@
...
+        if !self.account_storage_cached.contains_key(&address) {
+            trace!("Account not in cache, checking backend");
+            self.backend.storage(address, key)
+        } else {
+            let entry = self.account_storage_cached.get_mut(&address).unwrap();
+
+            match entry {
+                Some((_, storage)) => storage.get(&key).cloned().unwrap_or(H256::zero()),
+                None => {
+                    error!("Queried storage in cache is None: {:?}", address);
```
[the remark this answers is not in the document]
n-hutton: I have never yet seen evm or scilla make a read to a deleted entry, so I just put this log message in to see if it ever even happens.

scilla/src/backend_collector.rs
```diff
@@ -0,0 +1,365 @@
...
+                    H256::zero()
+                }
+            }
+        }
+    }
+
+    pub fn update_account_storage(&mut self, address: Address, key: H256, value: H256) {
```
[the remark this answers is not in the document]
n-hutton: good point

scilla/src/backend_collector.rs
```diff
@@ -0,0 +1,365 @@
...
+        trace!(
+            "Updating account storage for scilla: KEY: {:?} VALUE: {:?} Lens: {} {}",
+            key,
+            value,
+            key.len(),
+            value.len()
+        );
+        let key_as_hash = H256::from_slice(&Keccak256::digest(key.as_bytes()));
+
+        // To be able to recover the original keys, we will use the following scheme:
```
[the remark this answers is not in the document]
n-hutton: I don't believe its inevitable - but in the case that it was, how would you do it? Make the `db` take arbitrary lenght key value pairs, or have two databases, one fixed width and one not?

scilla/src/call_scilla_server.rs
n-hutton: I can re-create this for you if you like but IIRC the error was something like `Arc<T> is invariant over parameter T`. Basically, the io.add_method wants to take a `fn` (not fnOnce or fnMut which are inferred based on whats moved or referred to in the closure) - this is inferred using scoping and lifetime rules.

I don't understand the mechanism too well, but I believe that `Arc` understands the properties of its generic type (its lifetime, mutability etc). This would be through subtyping and variance. If you create an Arc<&T> it understands that you could potentially get a memory violation if the closure holding the Arc clone outlives the reference to T.

zilliqa/Cargo.toml
n-hutton: It is referenced here: https://youtrack.jetbrains.com/issue/RUST-11841/Can-not-run-nested-type-inference-types-and-highlighting-in-the-code-are-not-displayed

The behaviour in my editor is that it randomly corrupts hotkeys, highlighting comes on only sometimes, and it randomly deletes when you want to insert text. So, it makes it highly unusable which is quite surprising for a random library.

zilliqa/src/consensus.rs
n-hutton: It is only useful in a dev environment, although I think it's impossible currently to start the network as more than a single node :)

I'd approve a PR if you want to add that flag :)

zilliqa/src/consensus.rs
n-hutton: So I re-ran it, it will generate this panic at runtime:

```
2024-01-24T16:43:03.827129Z ERROR zilliqa::api::eth: e=Bincode does not support the serde::Deserializer::deserialize_any method

Stack backtrace:
   0: std::backtrace_rs::backtrace::libunwind::trace
             at /rustc/eeff92ad32c2627876112ccfe812e19d38494087/library/std/src/../../backtrace/src/backtrace/libunwind.rs:104:5
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/eeff92ad32c2627876112ccfe812e19d38494087/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2: std::backtrace::Backtrace::create
             at /rustc/eeff92ad32c2627876112ccfe812e19d38494087/library/std/src/backtrace.rs:331:13
   3: std::backtrace::Backtrace::capture
             at /rustc/eeff92ad32c2627876112ccfe812e19d38494087/library/std/src/backtrace.rs:296:9
   4: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at /Users/nhutton/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.79/src/error.rs:565:25
   5: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at /rustc/eeff92ad32c2627876112ccfe812e19d38494087/library/core/src/result.rs:1959:27
   6: zilliqa::db::Db::get_transaction_receipts
             at ./zilliqa/src/db.rs:227:5
```

What I wrote before isn't actually correct, it is on the deserialization, it doesn't know what type to deser into, as mentioned here: https://stackoverflow.com/questions/77315666/how-can-i-deserialize-a-bincode-serialised-data-into-something-like-serde-json

So it seems like it is something like, one of bincode or the struct it is getting deseralized into need to know the type/size, but neither of them know in the case of `serde_json::Value`

scilla/src/backend_collector.rs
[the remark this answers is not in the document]
JamesHinshelwood: Yeah probably two DBs, but I agree with you that lets try it like this for now.

zilliqa/src/api/types/zil.rs
JamesHinshelwood: Should we add a `TODO` or raise an issue to fix this?

zilliqa/src/exec.rs
JamesHinshelwood: Scilla events need to be encoded in EVM receipts as described in Section 5 of [RFC075](https://docs.google.com/document/d/196FM_IWCeKqBxKubii3b0af7veVPMV_r_i2K9q5myh8/edit?usp=sharing). ZQ1 implements this.

zilliqa/src/exec.rs
JamesHinshelwood: > However, I believe that certain clients will send the zero address to indicate contract creation (including scilla). Would you agree?

I think its important to consider the Ethereum and Zilliqa APIs separately.

In Ethereum, `to_addr = None` and `to_addr = Some(0x0)` are two different things - The transaction payloads will be different, have different signatures and different semantics. Clients that confuse these will definitely break on other EVM chains.

In Zilliqa, `to_addr` is not `Option`al and contract creations are specified by setting `to_addr = 0x0`.

I think the correct place for this mapping would be in the `fn to_addr` method of `SignedTransaction`. I think we should change the `Zilliqa` variant from:
```rust
            Transaction::Zilliqa(TxZilliqa { to_addr, .. }) => Some(*to_addr),
```
to
```rust
            Transaction::Zilliqa(TxZilliqa { to_addr, .. }) if to_addr.is_zero() => None,
            Transaction::Zilliqa(TxZilliqa { to_addr, .. }) => Some(*to_addr),
```

JamesHinshelwood (dismissed):
LGTM, minus the point about `None` addresses in `exec.rs` - Please can we fix that one before merging?

There are a few other open comments for which I'll let you decide whether you'd rather fix now or later.

JamesHinshelwood (approved):

zilliqa/src/api/types/zil.rs
n-hutton: I will create some tickets off the back of this PR

PR #666, opened 2024-01-18
fix h2

n-hutton:
Fix dependency


JamesHinshelwood (approved):
What was broken out of interest?

PR #686, opened 2024-01-29
pos with fix

n-hutton:


zilliqa/src/consensus.rs
```diff
@@ -564,58 +557,7 @@ impl Consensus {
...
-                    block_state_root,
-                    block
-                );
-                return Err(anyhow!(
-                    "state root hash mismatch, expected: {:?}, actual: {:?}",
-                    block_state_root,
-                    self.state.root_hash()
-                ));
-            }
+            self.execute_block(&block, transactions)?;
```
n-hutton, line 560: move the execution of a block into its own function because we potentially need to execute blocks when dealing with a fork

PR #688, opened 2024-01-31
Proper special-case handling of intershard TXs

theo-zil:
The two main features of this PR are:
1. Intershard TXs are not included as part of `Proposals` being broadcast during block creation, which ensures the block only gets voted on if all validators have locally seen in using their source shard (light) clients. As a consequence, Proposals now have an additional `opaque_transactions` field containing only hashes.
2. Nonce handling in intershard TXs is fixed: there's no account nonce associated with them, meaning they are effectively valid at all times. They do have a `bridge_nonce` field, which exists to ensure hash uniqueness and is set on the source chain, and does not have any semantic meaning on the target shard.

As a result `Transaction`s now have an `Option`al nonce. Main consequence is for Ethereum RPCs which must return a u64 - I've currently hardcoded it to `u64::MAX`.

Also, a new environment variable is added to the test suite, `ZQ_TEST_TIMEOUT`, because ever since the timeout check was added both tests dealing with shard creation would consistently take just slightly over the timeout on my local machine. So I'm letting myself bump it up for convenience. The default value is the same as the previous hardcoded one.


zilliqa/src/contracts/intershard_bridge.sol
```diff
@@ -4,35 +4,27 @@ pragma solidity ^0.8.4;
...
-    mapping(uint64 => mapping(uint64 => bool)) public dispatched;
-    uint64 public nonce;
+    uint64 nonce; // internal value used to guarantee transaction hash uniqueness
 
     function bridge(
         uint64 targetShard,
-        bool contract_creation,
-        address target,
+        bool contractCreation,
+        address target, // not used if this is contractCreation = true
```
WuBruno, line 22: can we use target == 0x0 for contract creation?

zilliqa/src/message.rs
```diff
@@ -170,10 +189,11 @@ pub struct BlockBatchResponse {
 pub struct IntershardCall {
     pub source_address: H160,
     pub target_address: Option<H160>,
+    pub source_chain_id: u64,
+    pub bridge_nonce: u64,
```
DrZoltanFazekas, line 193: To avoid confusion between UCCB and x-bridge, which is a specific application built on UCCB, we've renamed UCCB to Universal Cross-Chain Broker in the whitepaper - up to you if you want to rename `bridge_nonce` to `uccb_nonce`...

zilliqa/src/pool.rs
```diff
@@ -1,33 +1,83 @@
...
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+struct NoncedTxIndex {
+    pub address: Address,
+    pub nonce: u64,
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+struct IntershardTxIndex {
+    pub source_shard: u64,
+    pub link_nonce: u64,
```
DrZoltanFazekas, line 21: Does `link_nonce` refer to the `bridge_nonce` as it's called elsewhere in the code?

zilliqa/src/pool.rs
```diff
@@ -176,7 +263,30 @@ mod tests {
                 },
             },
             signer: from_addr,
-            hash: Hash::ZERO,
+            hash: Hash::compute([from_addr.as_bytes(), &[nonce]]),
```
DrZoltanFazekas, line 266: Is this the "official" txn hash? If it is, shouldn't it be keccak256 of all txn fields of the signed transaction?

zilliqa/tests/it/main.rs
```diff
@@ -821,15 +815,11 @@ impl Network {
             .unwrap()
     }
 
-    pub async fn genesis_wallet(
-        &mut self,
-    ) -> SignerMiddleware<Provider<LocalRpcClient>, LocalWallet> {
+    pub async fn genesis_wallet(&mut self) -> Wallet {
```
DrZoltanFazekas, line 818: Does the genesis wallet have the same semantics as the coinbase wallet on ZQ1 i.e. "holding" all not yet minted native tokens and receiving burned txn fees? On ZQ1 the coinbase address is 0x0, but according to the wallet creation in the next line, x-shards will have a non-zero genesis wallet address, right?

DrZoltanFazekas (approved):
Just added a few questions to ensure alignment between the whitepaper and the implementation.

zilliqa/tests/it/main.rs
theo-zil: The genesis wallets are pre-funded in the initial state according to the config passed to the genesis nodes (see [https://github.com/Zilliqa/zq2/blob/766617811202715b9b3d8ca9d8d15be82c60ee5e/zilliqa/src/state.rs#L73](src/state.rs)). There may be more than one such wallet. Other than minting some tokens to their balance, they are not special in any way.
Currently their main usecase is as easy pre-funded wallets to use during testing.

I'm not sure how we'll handle minting on mainnet - possibly it'll be allocated to genesis wallets, but e.g. validator fees can be minted on the go and don't have to be "held" by the genesis wallet.

zilliqa/src/pool.rs
DrZoltanFazekas: The main reason for keeping this nonce in the intershard case was to retain compatibility with UCCB and use the gateway contract for requesting cross-shard calls (the gateway contract injects the nonce in the encoded message structure), so calling it `uccb_nonce` makes sense to me

zilliqa/tests/it/main.rs
DrZoltanFazekas: Regarding mainnet ZIL supply we had a short discussion in whitepaper comments: https://docs.google.com/document/d/1wkYv7GV_-2X9Ha7Di56vjcNhS_H1o6n5rFbKz4xviYQ/edit?disco=AAABD6HQ3pQ

I guess we will import the null account from the ZQ1 persistence and from that wallet we will mint the mainnet validator rewards.

zilliqa/src/consensus.rs
```diff
@@ -1471,7 +1501,19 @@ impl Consensus {
 
                 let current_head = self.head_block();
 
-                self.proposal(Proposal::from_parts(block, transactions), true)?;
+                self.proposal(
+                    Proposal::from_parts_and_hashes(
```
JamesHinshelwood, line 1505: I would assume instead here you want to reconstruct the `Proposal` you had at the start of this call to `recieve_block`? Won't this drop any `opaque_transactions` that were sent?

zilliqa/src/pool.rs
```diff
@@ -1,33 +1,83 @@
...
+}
 
-use crate::{crypto::Hash, state::Address, transaction::VerifiedTransaction};
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+enum TxIndex {
+    /// from_address, nonce (unique for that address)
+    Nonced(Address, u64),
+    /// source_shard, nonce (unique for the bridge from that shard)
+    Intershard(u64, u64),
+}
```
JamesHinshelwood, line 30: Should we use the structs defined above here?
```suggestion
enum TxIndex {
    /// from_address, nonce (unique for that address)
    Nonced(NoncedTxIndex),
    /// source_shard, nonce (unique for the bridge from that shard)
    Intershard(IntershardTxIndex),
}
```

zilliqa/src/pool.rs
```diff
@@ -67,75 +116,105 @@ impl TransactionPool {
...
+                //
                 // We loop until we find a transaction that hasn't been made invalid.
                 continue;
             };
-            self.hash_to_index.remove(&transaction.hash);
 
-            if let Some(next_txn) = self.transactions.get(&(from_addr, nonce + 1)) {
-                self.ready.push(next_txn.into());
+            // If we've popped a nonced transaction, that may have made a subsequent one valid
+            if let TxIndex::Nonced(address, nonce) = tx_index {
```
JamesHinshelwood, line 133: I might be overthinking it, but a slightly cleaner way could be something like
```rust
impl TxIndex {
  fn next(&self) -> Option<TxIndex> {
    match self {
      TxIndex::Nonced(address, nonce) => Some(TxIndex::Nonced(address, nonce + 1)),
      _ => None,
    }
  }
}
```
and using `if let Some(next) = tx_index.next() {` here?

zilliqa/src/pool.rs
```diff
@@ -67,75 +116,105 @@ impl TransactionPool {
...
+                    return false;
+                } else {
+                    // Remove the existing transaction from `hash_to_index` if we're about to replace it.
+                    self.hash_to_index.remove(&existing_txn.hash);
+                }
+            }
+        } else {
+            // now we've confirmed it's nonceless, so ensure its index isn't duplicate
+            if let Some(existing_nonceless_txn) = self.transactions.get(&txn.mempool_index()) {
+                warn!(tx = ?existing_nonceless_txn, "Duplicate-indexed nonceless transactions encountered, this shouldn't happen. Ignoring the new one.");
```
JamesHinshelwood, line 169: How do we ensure nonces aren't duplicated later on, once transactions have left the mempool and been committed to the chain?

zilliqa/src/pool.rs
```diff
@@ -67,75 +116,105 @@ impl TransactionPool {
...
                 return false;
-            } else {
-                // Remove the existing transaction from `hash_to_index` if we're about to replace it.
-                self.hash_to_index.remove(&existing_txn.hash);
             }
         }
 
-        if txn.tx.nonce() == account_nonce {
-            // This transaction has a nonce equal to the account's current nonce, so it is ready to be executed.
+        // If this transaction ether has a nonce equal to the account's current nonce,
```
JamesHinshelwood, line 174:
```suggestion
        // If this transaction either has a nonce equal to the account's current nonce,
```

zilliqa/src/consensus.rs
```diff
@@ -566,8 +568,26 @@ impl Consensus {
...
+
+            // We re-inject any missing Intershard transactions (or really, any missing
+            // transactions) from our mempool. If any txs are unavailable either in the
+            // message or locally, the proposal cannot be applied
+            for (idx, tx_hash) in block.transactions.iter().enumerate() {
+                if transactions.get(idx).is_some_and(|tx| tx.hash == *tx_hash) {
+                    // all good
+                } else {
+                    let Some(local_tx) = self.transaction_pool.pop_transaction(*tx_hash) else {
+                        warn!("Proposal {} at view {} referenced a transaction that was neither included in the broadcast nor found locally - cannot apply block", block.hash(), block.view());
```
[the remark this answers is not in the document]
theo-zil: Not a silly question at all - this is down to the security model of intershard links. Basically intershard messaging is specified to be local-only, i.e. _never p2p_, and as such every validator of a given shard _A_ must have a (light or full) client of any source shards for those shards to be able to send messages to _A_. The validator's sister client will then message the validator node with the `IntershardCall`, populating the mempool.

So a precondition for an intershard tx to be included in a block is that 2/3rds of the validators, including the current leader, have had local clients observe the source call and relay a local message.

This is what allows us to skip all the logic from the UCCB and just accept these transactions as trusted, because we know the validators of _A_ have all locally observed the source transaction, and have verified its authenticity according to the source shard's consensus. Hence why p2p broadcasting is disabled - as p2p messages are inherently untrusted, e.g. a single node would be able to claim an intershard event happened.

zilliqa/src/consensus.rs
```diff
@@ -804,17 +829,20 @@ impl Consensus {
 
                     let transactions = self.get_txns_to_execute();
 
-                    let applied_transactions: Vec<_> = transactions
+                    let mut had_ixshd = false;
```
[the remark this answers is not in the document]
theo-zil: Whoops, left from debugging. Good catch thanks!

zilliqa/src/message.rs
```diff
@@ -32,18 +32,36 @@ pub struct Proposal {
...
             header: block.header,
             qc: block.qc,
             agg: block.agg,
             committee: block.committee,
-            transactions,
+            transactions: transactions.0,
+            opaque_transactions: block
+                .transactions
+                .into_iter()
+                .filter(|hash| !provided_tx_hashes.contains(hash))
```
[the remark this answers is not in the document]
theo-zil: It filters out all the `Hash`es that are provided alongside `Transaction`s, thus leaving, as the field name suggests, only the opaque transactions that are listed (as hashes) in `block.transactions` but without the body being provided.

However good thing you asked this, because upon re-reading, I'm finding that my naming is quite terrible here. I'll try to clarify the code to be more self-explanatory.

zilliqa/src/consensus.rs
theo-zil: The `opaque_transactions` will still be present as hashes on `block.transactions` - the `transactions` local variable here will only contain the non-opaque ones, whereas the `block` always contains _all_ hashes that should go into the block.

This `from_parts_and_hashes()` should be the exact inverse of `into_parts()`. As I've mentioned though the parameter naming in `from_parts_and_hashes()` is mediocre so I'll rename that.

zilliqa/src/pool.rs
theo-zil: Good catch - I went back and forth a few times on how I wanted to do this, and it seems this got accidentally left in.

I'm leaning towards actually just deleting the structs, since they're not used anywhere else, need no methods or other structure, it's only two fields which is easy to keep track of, and if we do use them then constructing a `TxIndex` inline becomes a lot more verbose and boilerplaty.

I'm gonna do that for now, let me know if you prefer otherwise.

zilliqa/src/pool.rs
theo-zil: I like this!

zilliqa/src/pool.rs
theo-zil: We don't - in theory we assume that the local sister node sending us the local `IntershardCall` notifications should never double-send the same one. This check is more of a sanity check for transient local errors - `"this shouldn't happen"` is not simply wishful, it's meant to be literal.
Otherwise the bridge nonce is not really meant to provide replay protection. The source shard is responsible for that.

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
DrZoltanFazekas: Maybe it was also James question and if not then it's mine: Is the payload of `Intershard` transactions always included in the block? If the block contained only hashes of `Intershard` transactions, which are not propagated via p2p, then non-validator nodes would never receive their payload and couldn't execute them...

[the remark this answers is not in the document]
theo-zil: Very good point.
Currently, only the hashes are included in Proposals for the current round, but the full body is included in responses to BlockRequest messages. So non-validator nodes would need to request the block explicitly, and wouldn't be able to follow the currently voted-on proposal.

We don't explicitly support non-validator nodes right now but a fix would be for them to simply request the block if they see a Proposal they can't execute. Nodes that want to e.g. run block explorers that want to display the currently voted-on tip, could then spin up sister clients for connected source shard(s) to get the same information that validators do, and everyone else would have to wait until the block vote is accepted and they can get it by requesting from a node.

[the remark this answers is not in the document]
DrZoltanFazekas: Got it, thanks. Maybe it's worth consireding that all non-validator nodes would request those blocks from the validators because they are the only ones that have the intershard transactions' body due to their sister clients, which might cause unnecessary workload on the validators. Why can't validators actually broadcast intershard transactions to everyone in the network, but rely on intershard transactions they have seen themselves when it comes to voting on a block?

JamesHinshelwood (approved):

PR #690, opened 2024-02-01
Pr feedback on scilla integration

n-hutton:


scilla/src/backend_collector.rs
```diff
@@ -61,27 +61,26 @@ impl<'a, B: Backend> BackendCollector<'a, B> {
...
+            let account = Account {
+                nonce: self.backend.basic(address).nonce.as_u64(),
+                code: self.backend.code(address),
+                storage_root: None,
+                is_scilla: true,
+            };
+
+            e.insert(Some((account, HashMap::from([(key, value)]))));
+        } else {
+            let entry = self.account_storage_buffered.get_mut(&address).unwrap();
```
JamesHinshelwood, line 76: Why did you change from `match` to `if let`? In my opinion, it was clearer before because you handle both cases explicitly rather than defaulting into the `else` and then having to `.unwrap()` once you know the key is there.

JamesHinshelwood (dismissed):

scilla/src/backend_collector.rs
n-hutton: Perhaps I misunderstood your original comment - I thought this was what you specified. I will switch it to the match again

JamesHinshelwood (dismissed):

PR #691, opened 2024-02-02
Fix peer discovery

JamesHinshelwood:
* Dial the configured bootstrap node straight away to speed up subsequent requests.
* Run the Kademlia bootstrap process every 5 minutes to ensure we discover new nodes periodically. Also run an early bootstrap if a point-to-point message fails because we sent it to a peer that we didn't know the address of.
* Ignore advertised loopback addresses


n-hutton (approved):
lgtm!

theo-zil (approved):

PR #718, opened 2024-02-13
Buffer votes for blocks we don't know about

JamesHinshelwood:
Votes can arrive for blocks we aren't yet aware of in (at least) two circumstances:
* If we disconnect and reconnect, we will download the blocks we missed, but nodes might send us their votes for the next block before we've received the missed blocks.
* Due to network latency, a vote for a block proposal could arrive before the block proposal itself.

Before this commit, the node ignores these votes and the network eventually recovers via a new view. However, this slows things down so we should recover faster if possible. Instead, we store votes for unknown blocks in memory and replay them if the block later becomes known to us.

The implementation is fairly straight forward but there are a few caveats and TODOs:
* The return type of `Consensus::proposal` is now more complicated, as it doesn't just return a `Message::Vote` any more. If the proposal results in some buffered votes being replayed and those votes form a supermajority, then the node can immediately propose the next block.
* There's nothing which limits the memory usage of buffered votes. A malicious node is perfectly able to send us loads of votes with non-existant block hashes, which we will store forever. I've raised #719 to resolve this.
* When applying buffered votes as a result of a proposal, they take priority over our own vote. This means we lose out on the cosigner reward. #720
* The unreliability test could be improved to be more efficient and to assert a stricter condition on the network - #721.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -26,7 +26,7 @@ use crate::{
         AggregateQc, BitSlice, BitVec, Block, BlockHeader, BlockRef, Committee, ExternalMessage,
         InternalMessage, NewView, Proposal, QuorumCertificate, Vote,
     },
-    node::MessageSender,
+    node::{MessageSender, NetworkMessage},
     pool::TransactionPool,
     state::State,
     time::SystemTime,
@@ -128,6 +128,9 @@ pub struct Consensus {
     message_sender: MessageSender,
     pub block_store: BlockStore,
     votes: BTreeMap<Hash, (Vec<NodeSignature>, BitVec, u128, bool)>,
+    /// Votes for a block we don't have stored. They are retained in case we recieve the block later.
+    // TODO(#719): Consider how to limit the size of this.
+    buffered_votes: BTreeMap<Hash, Vec<Vote>>,
     new_views: BTreeMap<u64, NewViewVote>,
     high_qc: QuorumCertificate,
     view: View,
@@ -286,6 +289,7 @@ impl Consensus {
             block_store,
             message_sender,
             votes: BTreeMap::new(),
+            buffered_votes: BTreeMap::new(),
             new_views: BTreeMap::new(),
             high_qc,
             view: View::new(start_view),
@@ -328,7 +332,7 @@ impl Consensus {
         &mut self,
         peer_id: PeerId,
         public_key: NodePublicKey,
-    ) -> Result<Option<(Option<PeerId>, ExternalMessage)>> {
+    ) -> Result<Option<NetworkMessage>> {
         if self.state.get_stake(public_key)?.is_none() {
             info!(%peer_id, "peer does not have sufficient stake");
             return Ok(None);
@@ -487,7 +491,7 @@ impl Consensus {
         &mut self,
         proposal: Proposal,
         during_sync: bool,
-    ) -> Result<Option<(PeerId, Vote)>> {
+    ) -> Result<Option<NetworkMessage>> {
         self.cleanup_votes();
         let (block, transactions) = proposal.into_parts();
         let head_block = self.head_block();
@@ -575,6 +579,27 @@ impl Consensus {
                 );
             }
 
+            if let Some(buffered_votes) = self.buffered_votes.remove(&block.hash()) {
+                // If we've buffered votes for this block, process them now.
+                let count = buffered_votes.len();
+                for (i, vote) in buffered_votes.into_iter().enumerate() {
+                    trace!("applying buffered vote {} of {count}", i + 1);
+                    if let Some((block, transactions)) = self.vote(vote)? {
+                        // If we reached the supermajority while processing this vote, send the next block proposal.
+                        // Further votes are ignored (including our own).
+                        // TODO(#720): We should prioritise our own vote.
+                        trace!("supermajority reached, sending next proposal");
+                        return Ok(Some((
+                            None,
+                            ExternalMessage::Proposal(Proposal::from_parts(block, transactions)),
+                        )));
+                    }
+                }
+
+                // If we reach this point, we had some buffered votes but they were not sufficient to reach a
+                // supermajority.
+            }
+
             if !block.committee.iter().any(|v| v.peer_id == self.peer_id()) {
                 trace!(
                     "can't vote for block proposal, we aren't in the committee of length {:?}",
@@ -587,7 +612,7 @@ impl Consensus {
 
                 if !during_sync {
                     trace!(proposal_view, ?next_leader, "voting for block");
-                    return Ok(Some((next_leader, vote)));
+                    return Ok(Some((Some(next_leader), ExternalMessage::Vote(vote))));
                 }
             }
         } else {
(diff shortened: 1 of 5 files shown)


zilliqa/src/block_store.rs
```diff
@@ -88,10 +88,12 @@ impl BlockStore {
     }
 
     pub fn request_blocks(&mut self, peer: Option<PeerId>, number: u64) -> Result<()> {
-        // If the request is higher than our head, lower it to our head, as we don't store
-        // loose blocks
-        let number = std::cmp::min(number, self.db.get_highest_block_number().unwrap().unwrap());
-        let number = std::cmp::max(number, 1); // avoid requesting genesis unnceccessarily
+        let highest_number = self.db.get_highest_block_number().unwrap().unwrap();
+
+        // Clamp the requested block to between 1 (because we already know the genesis block at 0) and the current
```
n-hutton, line 93: 👍

zilliqa/src/consensus.rs
```diff
@@ -123,6 +123,9 @@ pub struct Consensus {
     message_sender: MessageSender,
     pub block_store: BlockStore,
     votes: BTreeMap<Hash, (Vec<NodeSignature>, BitVec, u128, bool)>,
+    /// Votes for a block we don't have stored. They are retained in case we recieve the block later.
+    // TODO: Consider how to limit the size of this.
```
n-hutton, line 127: this todo could be a ticket unless its not resolved in the scope of this PR

zilliqa/src/consensus.rs
```diff
@@ -625,6 +629,27 @@ impl Consensus {
 
             self.save_highest_view(block.hash(), block.number(), proposal_view)?;
 
+            if let Some(buffered_votes) = self.buffered_votes.remove(&block.hash()) {
```
n-hutton, line 632: It seems unfortunate that this edge case is injected here, where processing a proposal can respond with another proposal - it might be worth thinking of a way to make sure the system will just 'settle' without having to buffer neccessarily

JamesHinshelwood: I think it makes sense for a proposal to result in another proposal if you accept that the node might be buffering sufficient votes for the first proposal to immediately form a supermajority.

I guess another way of representing this could be to buffer the votes in `Node`, rather than `Consensus`, then the return type of `proposal` wouldn't change. I'm not sure that would be much clearer though?

zilliqa/src/consensus.rs
```diff
@@ -1512,12 +1553,14 @@ impl Consensus {
                 )?;
 
                 // Return whether the head block hash changed as to whether it was new
-                Ok(self.head_block().hash() != current_head.hash())
+                let was_new = self.head_block().hash() != current_head.hash();
+
+                Ok((was_new, response))
```
theo-zil, line 1558: Also, is it really worth sending `Vote` messages? Any non-leader node in the committee will spam stale `Vote`s for every single block it syncs to, I think.

But really we're only ever interested in sending messages for the latest pending proposal (if any), and those messages can be either a Vote if we aren't the leader, or a Proposal iff we are the leader and buffered enough votes to accept the pending one.
So I think `Proposal`s returned will always be worth broadcasting, but `Vote`s will usually be stale.
Dropping the Vote seems like it shouldn't harm consensus too much usually (unless you're the last vote needed for a 2/3 majority and happened to be offline, which seems rare) and there's no easy way to detect whether we're at the head of the chain (and thus our vote matters) or not (and then it'd be stale). I think avoiding spamming the network with messages for every single block during on sync is probably worth dropping an occasional very rare meaningful vote.

zilliqa/tests/it/main.rs
```diff
@@ -672,14 +677,23 @@ impl Network {
...
                 let nodes: Vec<(usize, &TestNode)> = if let Some(destination) = destination {
-                    vec![self
+                    let (index, node) = self
                         .nodes
                         .iter()
                         .enumerate()
                         .find(|(_, n)| n.peer_id == destination)
-                        .unwrap()]
+                        .unwrap();
+                    if self.disconnected.contains(&index) {
```
theo-zil, line 686: Maybe removing this `if` here (and the `filter` below) and then just adding `nodes.retain(|(i, _)| !self.disconnected.contains(i));` to cover both cases at once would be more concise - down to preference

zilliqa/tests/it/unreliable.rs
```diff
@@ -0,0 +1,35 @@
...
+        let i = network.random_index();
+        if i != restarted_node {
+            break i;
+        }
+    };
+    let other_wallet = network.wallet_of_node(other_wallet).await;
+
+    // Produce a few blocks to start with. Enough for everyone to join the consensus committee.
+    // TODO(#721): Once the committee is visible in the API, we can avoid waiting as long.
+    network.run_until_block(&wallet, 8.into(), 400).await;
```
theo-zil, line 19: Is 8 blocks actually necessary in your experience? That seems like quite a long time to wait just for consensus to stabilise on genesis.

theo-zil (changes requested):
Mainly one question about messages sent after processing blocks during sync, and a couple of non-blocking minor nits/questions, otherwise all looks good

zilliqa/tests/it/main.rs
JamesHinshelwood: I still need a case to filter by the message's destination, if the node is not disconnected, but I think you're right that I can combine them.

I changed it to this, let me know if you think its worse :) 

```rust
                let nodes = self.nodes
                    .iter()
                    .enumerate()
                    .filter(|(i, n)| {
                        let disconnected = self.disconnected.contains(i);
                        // Send to this node if it is the intended destination or if the message is a broadcast.
                        let is_destination = destination.map(|d| n.peer_id == d).unwrap_or(true);

                        !disconnected && is_destination
                    });
```

zilliqa/tests/it/unreliable.rs
JamesHinshelwood: The test would failllll ona  few random test cases with a lower block number. What was happening was that only a single node was in the committee when we passed this point and that was also the node that got disconnect. This obviously meant no blocks were produced while the node was disconnected.

So I just raised this to a number that makes it very likely there will be at least 2 nodes in the committee.

I raised #721 to improve this though - We should really be doing "run until the committee contains at least 2 nodes".

zilliqa/src/consensus.rs
```diff
@@ -1504,12 +1547,14 @@ impl Consensus {
                 )?;
 
                 // Return whether the head block hash changed as to whether it was new
-                Ok(self.head_block().hash() != current_head.hash())
+                let was_new = self.head_block().hash() != current_head.hash();
+
+                Ok((was_new, response.filter(|_| was_new)))
```
theo-zil, line 1552: Re: my comment about sending stale votes: I still don't think that's sufficient, because `self.head_block()` is only our local view and may be changing all the time while we're syncing. In fact IIRC `was_new` is used to decide whether to request more blocks to continue syncing, so if `was_new == true` it's pretty much an indicator that we're still syncing and therefore still out of date, if I remember the fetching logic correctly.

I still think that unconditionally dropping all `Vote`s (and potentially still broadcasting `Proposal`s just in case) will be worth it, as the very niche edge case where this would cause a `NewView` being far outweighed in my opinion by the upside of significantly cutting down on stale traffic.

JamesHinshelwood: Right yeah, sorry I misunderstood you originally. I agree.

To restate your argument and confirm my understanding:

If we've requested a pre-existing block and received it, we should assume that we're already too late to vote on that block. In 99% of cases this assumption will be true. In the 1% of cases, where we've requested the most recent block and the next block hasn't been proposed, missing our vote will not bring down the network. In fact, if we wanted to handle this case in the future, I'd suggest we also keep a buffer of the most recent `Proposal` we received, if it didn't form a chain with our existing blocks. However, I don't think that is very useful or worth the effort at the moment.

So I will restore this function back to the previous version :)

theo-zil (approved):

PR #730, opened 2024-02-19
Fix revert test cases

bzawisto:
This PR fixes issues related to gas estimation - where the client expects well formatted response having potentially revert message or panic code. 
I also enabled line numbers to be printed which helps with log analysis - I found that more convenient that doing grep in the code for log prefixes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,9 +39,9 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.8.3"
+version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac1f845298e95f983ff1944b728ae08b8cebab80d684f0a832ed0fc74dfa27e2"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
@@ -112,9 +112,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.11"
+version = "0.6.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e2e1ebcb11de5c03c67de28a7df593d32191b44939c482e97702baaaa6ab6a5"
+checksum = "96b09b5178381e0874812a9b157f7fe84982617e48f71f4e3235482775e5b540"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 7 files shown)


zilliqa/src/eth_helpers.rs
```diff
@@ -0,0 +1,68 @@
...
+    const REVERT_SELECTOR: [u8; 4] = [0x08, 0xc3, 0x79, 0xa0];
+    const PANIC_SELECTOR: [u8; 4] = [0x4e, 0x48, 0x7b, 0x71];
+
+    let generic_error = String::from("execution reverted");
+    // At least 4 elemenets + anything to decode
+    if encoded.len() < REVERT_SELECTOR.len() + 1 {
+        return generic_error;
+    }
+    // It is safe since we check the size above
+    let prefix: [u8; 4] = unsafe { *encoded.as_ptr().cast::<[u8; 4]>() };
```
JamesHinshelwood, line 17: Can we avoid the `unsafe` here? I doubt its necessary for performance.

You could do something like
```rust
let decoded_error = if encoded.starts_with(&REVERT_SELECTOR) {
  ethabi::decode(&[ParamType::String], &encoded[4..])
} else if encoded.starts_with(&PANIC_SELECTOR) {
  ethabi::decode(&[ParamType::Uint(256)], &encoded[4..])
} else {
  return generic_error;
};
let Ok(decoded_error) = decoded_error else { return generic_error; };
// etc.
```
?

JamesHinshelwood: Also should we consider `ethabi::decode_whole` instead of `ethabi::decode`?

zilliqa/src/exec.rs
```diff
@@ -915,7 +916,19 @@ impl State {
...
                         format!("Estimate gas failed with error: {:?}", result.exit_reason);
                     warn!(error_str);
-                    return Err(anyhow!(error_str));
+
+                    let decoded_revert_msg = extract_revert_msg(&result.return_value);
+
+                    // See: https://github.com/ethereum/go-ethereum/blob/9b9a1b677d894db951dc4714ea1a46a2e7b74ffc/internal/ethapi/api.go#L1026
+                    const REVERT_ERROR_CODE: i32 = 3;
+
+                    let response = jsonrpsee::types::ErrorObjectOwned::owned(
```
JamesHinshelwood, line 925: As a minor point to consider for the future - I don't like that we're adding API-specific concerns at this level. Ideally this function would be returning an "I reverted" error and we map that to the error code at the API-layer. I'd say a similar thing for the special "Txn Hash not present" error which we have to later suppress since its not a 'real' error.

We can fix it in the future though :)

zilliqa/src/eth_helpers.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    let generic_error = String::from("execution reverted");
+    // At least 4 elemenets + anything to decode
+    if encoded.len() < REVERT_SELECTOR.len() + 1 {
+        return generic_error;
+    }
+
+    let payload = &encoded[4..];
+
+    let Ok(vec) = (if encoded.starts_with(&REVERT_SELECTOR) {
```
JamesHinshelwood, line 19: Small nit but I'd prefer to split this over two separate lines to avoid the large bracketed expression. My mind sees the `(`, then needs to scan for the closing `)` and see the `else` after it before I understand what this line does.
i.e.
```rust
let vec = if encoded.starts_with(...) {
  ...
} else {
  ...
};
let Ok(vec) = vec else { return generic_error; };
```

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #733, opened 2024-02-19
Log swarm events

JamesHinshelwood:
At debug level, we log all libp2p swarm events. As it stands, lots of useful events such as errors are lost by our `_ => {}` branch.

In future, we should consider handling more of of these cases explicitly and log `error!`s.


JamesHinshelwood:
FWIW the diff is me adding `debug!(?event, "swarm event")` before the `match` and then indenting everything inside one level deeper. Sorry for the messy diff.

PR #738, opened 2024-02-22
Fixes to static call and transfer issues

bzawisto:
This PR fixes issues related with static calls, transfers and selfdestructs. In a nutshell, there's a temporary state kept in `EvmBackend` tracking current balance updates - which if succeeded - are applied in `apply_delta` call.


JamesHinshelwood (approved):
LGTM, thanks

PR #754, opened 2024-03-01
Tests and mempool fixes

### LINKED ISSUE #755: Block proposers fail to execute their own blocks when an intershard transaction is present
Intershard transactions do not have nonces that can protect against replay attacks, so instead we only insert them into the mempool based on trusted local messages from a sister-node (and do not accept them from any other sources, such as broadcasted block proposals), and rely on correct mempool accounting to ensure they're included only once - namely, any transaction is removed from the mempool upon execution (as one might expect).

The issue is that block leaders execute their own proposal twice: firstly, when constructing it as a result of the last `Vote` that brings the previous block to a supermajority; and the second time, after they broadcast the resulting `Proposal`, it does a round-trip through the message channels and is sent back locally to the leader from the `p2p_node` same way as every other node receives the message, causing the leader to execute the block again.

As a result, a leader that tries to include an intershard transaction will execute it (and thus purge it from the mempool) upon construction, and then when receiving its own proposal for voting, will fail to do so as it no longer has the transaction in its mempool.

The two solutions I can think of are
1. Add special accounting in the mempool to allow leaders to "fake-execute" the transaction during construction, leaving it to linger in the mempool; or
2. Act on the `TODO` in the code to optimise leaders by having them vote on their own proposals immediately, rather than having them execute the block twice.

theo-zil:
This branches off of #644. Until it's merged the diff includes all those commits, the last two are the only relevant changes here: https://github.com/Zilliqa/zq2/pull/754/files/4a6c77c90cb1785023edcf9dbc7dbf3f8221f1ec..df1a3fed9a18228c8a550a2d2adbfb1c630f47ff

Four fixes:
 - In the `cross_shard_contract_creation` test, fix a rare issue where, while we're waiting to detect the cross-shard transaction, the node our wallet is connected to could finalize two blocks within the same tick (this happens if the node didn't manage to vote on the block, and then has to request a batch of blocks to catch up, and processes at least two new blocks upon receiving the reply message). The test would then never check a block, never find the transaction and fail even though it ran correctly. A workaround is included by explicitly keeping track of the latest checked block with a mutex, which is a bit ugly but allows it to work within the async closure.

That bug causing tests to fail was what prompted the initial investigation, but in the process three other bugs were identified and fixed:

 - In the mempool, rename `update_nonce()` to `mark_executed()` to reflect its intended functionality. This now correctly purges the newly-executed transaction from the pool; for nonced transaction it also marks the next-higher-nonce one as ready, if one exists (as per previous functionality). This avoids a bug where intershard transactions were being included twice in blocks, due to some nodes erroneously keeping them inside the mempool.
 - Add a guard to ensure nodes cannot send a LaunchShard message to themselves. This was creating duplicate nodes (with the same private key) in the same shard during tests, which should not be possible and was causing issues. A check for this was added both in tests and in the real p2p coordinator.
 - Re-insert local-only (i.e. intershard) transactions into the mempool after a proposer builds a proposal, to allow the proposer to then vote on their own block without running into the issue of missing said transactions (due to the current architecture where proposers double-execute blocks, once to build the proposal for broadcasting, once to vote on it after they've broadcasted it). (closes #755)


zilliqa/src/pool.rs
```diff
@@ -151,15 +151,14 @@ impl TransactionPool {
...
-                    self.hash_to_index.remove(&existing_txn.hash);
                 }
+                // Remove the existing transaction from `hash_to_index` if we're about to replace it.
+                self.hash_to_index.remove(&existing_txn.hash);
             }
         } else {
             // now we've confirmed it's nonceless, so ensure its index isn't duplicate
             if let Some(existing_nonceless_txn) = self.transactions.get(&txn.mempool_index()) {
-                warn!(tx = ?existing_nonceless_txn, "Duplicate-indexed nonceless transactions encountered, this shouldn't happen. Ignoring the new one.");
+                info!(tx = ?existing_nonceless_txn, "Duplicate-indexed nonceless transactions encountered, ignoring the new one.");
```
theo-zil, line 161: This turns out to actually happen normally when `consensus.new_transaction()` is called for every transaction inside `consensus.execute_block()`. Thus downgraded to an `info()` diagnostic. The updated `mark_executed()` ensures these transactions still don't get double-executed, so it's not a problem.

zilliqa/src/state.rs
```diff
@@ -158,6 +156,17 @@ impl State {
         })
     }
 
+    pub fn mutate_account<F: FnOnce(&mut Account)>(
```
JamesHinshelwood, line 159: This is a great idea. How easily can we make `save_account` non-`pub` and use this everywhere?

JamesHinshelwood (approved):

zilliqa/src/state.rs
theo-zil: Only place where it wasn't obvious was in exec.rs `apply_delta()`. It can of course be done but it didn't seem obviously simpler to me to refactor that way so I left it for now, to avoid even more changes in this PR - could open a new small PR if you think the refactor's worth it

JamesHinshelwood (changes requested):
As agreed on Slack, lets fix the tests first.

zilliqa/src/pool.rs
JamesHinshelwood: Can we simplify lines 143-147 in this function by using `.mempool_index()`.

i.e.
```rust
if let Some(existing_txn) = self.transactions.get(txn.mempool_index()) {
  ...
}
```
should cover both cases?

theo-zil: Good call. This has to be a no-op for the nonceless case, and refactoring it that way _should_ keep that property - I've added a comment explaining why, and the tests are passing, so should be good and does make it more concise and readable.

JamesHinshelwood (approved):

PR #764, opened 2024-03-06
Gas tracker 

bzawisto:
This PR adds basic Gas Tracker that is currently unused but will be employed when we add support for revm traps and evm/scilla interop.


JamesHinshelwood (commented):
I think this should be handled by `reth` now we've switched to it?

JamesHinshelwood:
Thanks - In that case, I'll close this PR but we can leave the branch around for reference.

PR #785, opened 2024-03-19
Make empty block creation time configurable

bzawisto:
This introduces configurable empty block timeouts - PR is not that long but I extracted some parts of `propose` into `propose_block` so that it can be easily reused in other place. I also borrowed `NetworkMessage` type from other PR - this will have to be possible resolved later when other PR goes to main too.

### DIFF
--- a/zilliqa/tests/it/consensus.rs
+++ b/zilliqa/tests/it/consensus.rs
@@ -69,7 +69,7 @@ async fn block_production_even_when_lossy_network(mut network: Network) {
                 let index = n.random_index();
                 n.get_node(index).get_finalized_height() >= start_block
             },
-            80,
+            100,
         )
         .await
         .unwrap();


zilliqa/src/consensus.rs
```diff
@@ -142,6 +144,8 @@ pub struct Consensus {
     transaction_pool: TransactionPool,
     // PRNG - non-cryptographically secure, but we don't need that here
     rng: SmallRng,
+    /// Flag indicating that block creation should be postponed due to empty mempool
+    create_next_block_on_timeout: bool,
```
JamesHinshelwood, line 148: I'm probably missing it, but I can't find the logic that will send the block if we receive a `NewTransaction` while waiting?

zilliqa/src/consensus.rs
```diff
@@ -419,11 +426,37 @@ impl Consensus {
             return Ok(None);
         }
 
+        let head_block = self.head_block();
+        let head_block_view = head_block.view();
+
+        if head_block_view + 1 == self.view.get_view() && self.create_next_block_on_timeout {
```
JamesHinshelwood, line 432: I wonder if we can make the logic slightly more stateless here instead? I am worried keeping track of it in a separate flag might lead to weird bugs, especially when taking into accounts forks, new views, etc.

In what situations would we call `timeout()` and `create_next_block_on_timeout` be `false`?

JamesHinshelwood (commented):
Sorry, just had a few comments to help me understand the logic in `consensus.rs` - I'm probably being dim.

zilliqa/src/consensus.rs
DrZoltanFazekas: `empty_block_timeout` does not only have to be smaller than `consensus_timeout` , we also need to carefully determine the latest point when we would reach supermajority with an empty mempool and decide to wait for `empty_block_timeout` instead of proposing a block. This is what I tried to convey in my second comment below.

zilliqa/src/consensus.rs
```diff
@@ -778,73 +815,125 @@ impl Consensus {
...
-                        self.state.set_to_root(H256(parent.state_root_hash().0));
-                    }
+                if (self.view.get_view() == 1)
+                    || (block_view + 1 == self.view.get_view() && transactions_count > 0)
+                {
+                    return self.propose_new_block();
+                } else {
+                    self.create_next_block_on_timeout = true;
+                    self.reset_timeout
+                        .send(self.config.consensus.empty_block_timeout)?;
```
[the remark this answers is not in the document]
DrZoltanFazekas: If `transaction_count == 0` when we reach the supermajority, the above code will execute the `else` branch starting at line 834, waiting for the `empty_block_timeout`. However, if this happened shortly before the `consensus_timeout` expires, we will miss the chance to propose a block. If it happened earlier, we will propose a block but some honest nodes might run into the `consensus_timeout` before receiving it.

Let me give an example with concrete numbers. Let's say `consensus_timeout = 5000ms` and `empty_block_timeout = 1000ms`. If we reach supermajority at `4500ms` and the mempool is empty, we must not go and wait for another `1000ms`. If we reach supermajority at `3500ms`, it's still too risky to wait for another `1000ms` before we propose an empty block as some nodes will not receive it on time.

DrZoltanFazekas (commented):
@bzawisto I elaborated on my two comments, but we can also jump on a quick call to discuss what I mean.

zilliqa/src/consensus.rs
```diff
@@ -482,7 +513,24 @@ impl Consensus {
...
+    fn get_consensus_timeout_params(&self) -> (u64, u64) {
+        let head_block = self.head_block();
+        let head_block_view = head_block.view();
+        let consensus_timeout_ms = self.config.consensus.consensus_timeout.as_millis() as u64;
+        let time_since_last_view_change = SystemTime::now()
+            .duration_since(self.view.last_timeout())
+            .expect("last timeout seems to be in the future...")
+            .as_millis() as u64;
+        let view_difference = self.view.get_view().saturating_sub(head_block_view);
+        let exponential_backoff_timeout = consensus_timeout_ms * 2u64.pow(view_difference as u32);
```
DrZoltanFazekas, line 531: We must calculate the view difference from the high_qc, not the head block: `let view_difference = self.view.get_view().saturating_sub(self.high_qc.view);`

zilliqa/src/consensus.rs
```diff
@@ -832,83 +881,158 @@ impl Consensus {
...
-                        self.state.set_to_root(H256(parent.state_root_hash().0));
+                    if (self.view.get_view() == 1)
+                        || (block_view + 1 == self.view.get_view() && transactions_count > 0)
+                    {
+                        return self.propose_new_block();
+                    } else {
+                        // Check if there's enough time to wait on a timeout and then propagate an empty block in the network before other participants trigger NewView
+                        let (time_since_last_view_change, exponential_backoff_timeout) =
+                            self.get_consensus_timeout_params();
+                        let minimum_time_left_for_empty_block =
```
DrZoltanFazekas, line 909: Wouldn't it make sense if `minimum_time_left_for_empty_block()` also returned the `minimum_time_left_for_empty_block`?

zilliqa/src/consensus.rs
```diff
@@ -431,19 +437,39 @@ impl Consensus {
...
+                .duration_since(self.view.last_timeout())
+                .expect("last timeout seems to be in the future...")
+                .as_millis() as u64;
+
+            let empty_block_timeout_ms =
+                self.config.consensus.empty_block_timeout.as_millis() as u64;
+
+            let transactions_count = self.transaction_pool.size();
+            if time_since_last_block > empty_block_timeout_ms || transactions_count > 0 {
+                if let Ok(Some((block, transactions))) = self.propose_new_block() {
```
DrZoltanFazekas, line 454: We have to extend the if-condition by
`|| time_since_last_view_change + minimum_time_left_for_empty_block >= exponential_backoff_timeout`
as you did in the `vote()` function

DrZoltanFazekas (changes requested):
Added a few more comments

zilliqa/src/consensus.rs
```diff
@@ -457,6 +494,11 @@ impl Consensus {
 
         trace!("Considering view change: view: {} time since: {} timeout: {} last known view: {} last hash: {}", self.view.get_view(), time_since_last_view_change, exponential_backoff_timeout, head_block_view, head_block.hash());
 
+        let view_difference = self.view.get_view().saturating_sub(head_block_view);
+        let consensus_timeout_ms = self.config.consensus.consensus_timeout.as_millis() as u64;
+        let next_exponential_backoff_timeout =
+            consensus_timeout_ms * 2u64.pow((view_difference + 1) as u32);
```
DrZoltanFazekas, line 500: This needs to be fixed the same way as you fixed `get_consensus_timeout_params()`:
` let view_difference = self.view.get_view().saturating_sub(self.high_qc.view);`

Use `high_qc.view` instead of `head_block_view`

DrZoltanFazekas (changes requested):
Sorry @bzawisto, I found we're computing the exponential backoff at two places, the other one needs a fix too

DrZoltanFazekas (dismissed):

DrZoltanFazekas (dismissed):

DrZoltanFazekas (approved):

PR #787, opened 2024-03-21
Terraform code refactoring - phase 1

mauromedda:
The Terraform code supports:
* the usage of existing VPC and subnets
* the zq2 node configuration includes the genesis_deposits
* the genesis, bootstrap and nodes' private keys are not provided are now inputs to the terraform iac


infra/tf/.terraform-docs.yaml
JamesHinshelwood, line 1: What's this for? Do we need it? (IME the auto-generated TF docs are not useful - I'd rather just read the code)

infra/tf/variables.tf
```diff
@@ -0,0 +1,106 @@
...
+variable "subdomain" {
+  type     = string
+  nullable = false
+}
+
+variable "network_name" {
+  description = "(Optional) ZQ2 network name"
+  type        = string
+  nullable    = false
+  default     = "zq2-devnet"
```
JamesHinshelwood, line 40: IMO we shouldn't default this. The value of `zq2-devnet` only makes sense for a single network (the ZQ2 devnet), so 99% of the time we want to specify it anyway.

infra/tf/variables.tf
```diff
@@ -0,0 +1,106 @@
...
+  type     = string
+  default  = "europe-west2-a"
+  nullable = false
+}
+
+variable "eth_chain_id" {
+  description = "ZQ2 EVM Chain ID"
+  type        = number
+  nullable    = false
+  default     = 33469
```
JamesHinshelwood, line 28: Likewise for this - This value only makes sense for the devnet, so I don't think it should be the default.

infra/tf/variables.tf
```diff
@@ -0,0 +1,106 @@
...
+}
+
+variable "labels" {
+  type        = map(string)
+  description = "A single-level map/object with key value pairs of metadata labels to apply to the GCP resources. All keys should use underscores and values should use hyphens. All values must be wrapped in quotes."
+  nullable    = true
+  default     = {}
+}
+
+variable "zq2_type" {
```
JamesHinshelwood, line 50: Maybe a style choice, but IMO this is a bit of an anti-pattern. I don't think we should specify a 'type' which controls how different resources are configured. I think we should expose those resource configurations more directly.

infra/tf/locals.tf
```diff
@@ -0,0 +1,16 @@
+################################################################################
+# ZQ2 GCP Terraform locals variables
+################################################################################
+
+locals {
+  binary_location      = "${path.module}/../../target/x86_64-unknown-linux-gnu/release/zilliqa"
+  labels               = merge(var.labels, { "zq2-network" = var.network_name })
+  devnet_resources     = var.zq2_type == "devnet" ? 1 : 0
```
JamesHinshelwood, line 8: What is this variable doing? It seems to conditionalise a whole bunch of resources, but I don't really understand why we want those resources conditionalised and how it relates to the devnet?

infra/tf/variables.tf
```diff
@@ -0,0 +1,106 @@
...
+}
+
+variable "node_count" {
+  description = "(Optional) ZQ2 Node count"
+  type        = number
+  nullable    = false
+  default     = 3
+}
+
+variable "secret_keys" {
```
JamesHinshelwood, line 101: Do we need this any more if keys are coming from 1password? (are they?)

infra/tf/locals.tf
mauromedda: I kept it for backward compatibility on how the existing devnet works. It provision VPCs, subnet and also gives people the possibility to generate random secrets without dependencies from 1password, so you can use this code to deploy dummy network avoiding the "path-to-prod" dependencies we have specifically on our environment and you still have the option to pass the options as variable inputs if needed (like secret keys or subnets). We can just remove this and add a skip_<resource_name> var instead or just have mandatory inputs for almost all the other stuff and simplify the code removing vpc/sn and so on. Happy both ways, no special need for me except keep the process which was in place working

infra/tf/variables.tf
mauromedda: same as below - backward compatibility with the in use provisioning. Can be removed.

infra/tf/.terraform-docs.yaml
mauromedda: It can be used to run terraform-docs and have the input/outputs and provider required in the readme of the iac folder a quite standard approach if people feel it's useful to have them on a mk doc. I don't use it either, so I can zap it, if the team feel like they are not going to use it

infra/tf/variables.tf
mauromedda: As mention above, if we assume that all this inputs will come from as input and not generated as tf random string, just need the secrets as inputs and not generated in the code.

JamesHinshelwood (approved):
LGTM, thanks!

PR #790, opened 2024-03-21
Fix response format of `GetVersion`

JamesHinshelwood:
It now returns an object containing `Commit` and `Version` fields, which is what ZQ1 does.

`Commit` is populated by the full SHA of the git commit at which the node was built.

`Version` is populated by a git tag, if one exists or an abbreviated commit SHA if not. If the working directory is dirty, `-dirty` is appended to the version. This is close to the output of `git describe --always --dirty --tags`.

Example output when `HEAD` is tagged:
```
{
    "Commit": "c2ecdc27e588a0ababcbe9a00567351d739664b4",
    "Version": "foobar"
}
```

Example output when `HEAD` is not tagged:
```
{
    "Commit": "c2ecdc27e588a0ababcbe9a00567351d739664b4",
    "Version": "c2ecdc2"
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2219,6 +2219,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4271d37baee1b8c7e4b708028c57d816cf9d2434acb33a549475f78c181f6253"
 
 [[package]]
+name = "git2"
+version = "0.18.2"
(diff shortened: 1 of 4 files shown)


rrw-zilliqa (approved):
Looks good to me - thank you!

PR #802, opened 2024-03-26
Use publically accessible faucet image

JamesHinshelwood:


docker-compose.yaml
```diff
@@ -84,7 +84,7 @@ services:
   faucet:
     depends_on:
       - node0
-    image: asia-docker.pkg.dev/prj-p-devops-services-tvwmrf63/zilliqa/eth-spout:89eb40d3
+    image: asia-docker.pkg.dev/prj-p-devops-services-tvwmrf63/zilliqa-public/eth-spout:89eb40d3
```
mauromedda, line 87: let's use the virtual registry `/zilliqa`, as the permissions should now be fixed

PR #819, opened 2024-04-04
Implement `eth_subscribe` for `newHeads`

JamesHinshelwood:
This adds the `eth_subscribe` and `eth_unsubscribe` APIs for the Ethereum pub-sub protocol.

Currently, only `newHeads` are supported (`logs` and `newPendingTransactions` are also specified). The API only works over a websocket, because they need a duplex connection.

The implementation works by adding a `broadcast::Sender` to `Consensus` which publishes `BlockHeader`s whenvever they are produced. When a client makes a new subscription, it subscribes to this sender. The sender keeps a buffer of the 4 latest blocks in memory. If any reciever falls behind these 4 blocks (because sending the blocks to the client is too slow), it will be disconnected.

### DIFF
--- a/zilliqa/tests/it/eth.rs
+++ b/zilliqa/tests/it/eth.rs
@@ -14,7 +14,7 @@ use ethers::{
     },
     utils::keccak256,
 };
-use futures::future::join_all;
+use futures::{future::join_all, StreamExt};
 use primitive_types::{H160, H256};
 use serde::Serialize;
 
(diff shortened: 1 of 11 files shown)


zilliqa/tests/it/eth.rs
```diff
@@ -1031,3 +1031,29 @@ async fn get_transaction_by_index(mut network: Network) {
         .unwrap();
     assert_eq!(txn.hash, h2);
 }
+
+#[zilliqa_macros::test]
+async fn block_subscription(mut network: Network) {
+    let wallet = network.genesis_wallet().await;
+    let provider = wallet.provider();
+
+    let mut block_stream = provider.subscribe_blocks().await.unwrap();
```
theo-zil, line 1040: I think the `wallet` already exposes the method directly - there's no need to access the inner `provider`:
```suggestion
    let mut block_stream = wallet.subscribe_blocks().await.unwrap();
```

theo-zil (approved):
Everything looks reasonable to me!

zilliqa/tests/it/eth.rs
JamesHinshelwood: Thanks!

theo-zil (approved):

JamesHinshelwood:
Fixed conflicts, sorry @theo-zil  - Needs one more :heavy_check_mark:

theo-zil (approved):

Issue #803, opened 2024-03-26
Allow stakers to not get rewarded


rrw-zilliqa:
There are situations in which stakers do not want to be rewarded; we should allow the rewardAddress for a staker to be 0 (meaning "don't pay rewards to me").
This means that in the code when peerId is retrieved from contract - it should no longer be validated against rewardAddress.

PR #826, opened 2024-04-09
Use committee from contract instead of a one attached to block

bzawisto:
Committee is now taken from staking contract instead being attached directly to block. Since Peer identities are not kept in staking contract - there's a list of Peers which are used for routing messages. Some of the tests now need more time for execution.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -158,7 +158,7 @@ checksum = "1a047897373be4bbb0224c1afdabca92648dc57a9c9ef6e7b0be3aff7a859c83"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -242,9 +242,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.82"
+version = "1.0.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f538837af36e6f6a9be0faa67f9a314f8119e4e4b5867c6ab40ed60360142519"
+checksum = "25bdb32cbbdce2b519a9cd7df3a678443100e265d5e25ca763b7572a5104f5f3"
 dependencies = [
  "backtrace",
 ]
@@ -482,7 +482,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -493,7 +493,7 @@ checksum = "c6fa2087f2753a7da8cc1c0dbfcf89579dd57458e36769de5ac750b4671737ca"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -549,7 +549,7 @@ checksum = "3c87f3f15e7794432337fc718554eaa4dc8f04c9677a950ffe366f20a162ae42"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -1027,7 +1027,7 @@ dependencies = [
  "heck 0.5.0",
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -1352,7 +1352,7 @@ checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -1511,7 +1511,7 @@ checksum = "487585f4d0c6655fe74905e2504d8ad6908e4db67f744eb140876906c2f3175d"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -1650,7 +1650,7 @@ dependencies = [
  "heck 0.4.1",
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -1661,7 +1661,7 @@ checksum = "6fd000fd6988e73bbe993ea3db9b1aa64906ab88766d654973924340c8cddb42"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -1831,7 +1831,7 @@ dependencies = [
  "reqwest 0.11.27",
  "serde",
  "serde_json",
- "syn 2.0.60",
+ "syn 2.0.63",
  "toml",
  "walkdir",
 ]
@@ -1849,7 +1849,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "serde_json",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -1875,7 +1875,7 @@ dependencies = [
  "serde",
  "serde_json",
  "strum",
- "syn 2.0.60",
+ "syn 2.0.63",
  "tempfile",
  "thiserror",
  "tiny-keccak",
@@ -2248,7 +2248,7 @@ checksum = "87750cf4b7a4c0625b1529e4c543c2182106e4dedc60a2a6455e00d212c489ac"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -3252,7 +3252,7 @@ dependencies = [
  "proc-macro-crate 3.1.0",
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -3781,7 +3781,7 @@ dependencies = [
  "heck 0.4.1",
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -4303,7 +4303,7 @@ dependencies = [
  "proc-macro-crate 3.1.0",
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -4405,7 +4405,7 @@ checksum = "a948666b637a0f465e8564c73e89d4dde00d72d4d473cc972f390fc3dcee7d9c"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.60",
+ "syn 2.0.63",
 ]
 
 [[package]]
@@ -4627,9 +4627,9 @@ dependencies = [
 
 [[package]]
 name = "paste"
-version = "1.0.14"
+version = "1.0.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "de3145af08024dea9fa9914f381a17b8fc6034dfb00f3a84013f7ff43f29ed4c"
+checksum = "57c0d7b74b563b49d38dae00a0c37d4d6de9b432382b2892f0574ddcae73fd0a"
 
 [[package]]
 name = "path-slash"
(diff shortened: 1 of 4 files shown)


zilliqa/src/consensus.rs
```diff
@@ -731,21 +769,29 @@ impl Consensus {
             return Ok(None);
         }
 
+        let Ok(committee) = self.state.get_stakers_at_block(&block) else {
+            info!("Skipping vote outside of committee");
+            return Ok(None);
+        };
+
+        //info!("VOTE PUB KEY: {}", hex::encode(vote.public_key.as_bytes()));
```
theo-zil, line 777: Dev comment

zilliqa/src/consensus.rs
```diff
@@ -731,21 +769,29 @@ impl Consensus {
             return Ok(None);
         }
 
+        let Ok(committee) = self.state.get_stakers_at_block(&block) else {
+            info!("Skipping vote outside of committee");
+            return Ok(None);
```
theo-zil, line 774: Similarly here, if we aren't able to get the committee at a block should we perhaps return an `Err()`?

zilliqa/src/consensus.rs
```diff
@@ -731,21 +769,29 @@ impl Consensus {
...
-            .committee
+        let Some((index, _)) = committee
             .iter()
             .enumerate()
-            .find(|(_, v)| v.public_key == vote.public_key)
-            .unwrap();
+            .find(|(_, &v)| v == vote.public_key)
+        else {
+            warn!("Skipping vote outside of committee");
+            return Ok(None);
```
theo-zil, line 785: Nice catch, that `unwrap()` was definitely wrong

zilliqa/src/consensus.rs
```diff
@@ -819,9 +865,9 @@ impl Consensus {
                         qc,
                         parent_hash,
                         self.state.root_hash()?,
-                        applied_transaction_hashes,
+                        applied_transaction_hashes.clone(),
```
theo-zil, line 868: clone left over from dev?

zilliqa/src/consensus.rs
```diff
@@ -852,64 +898,55 @@ impl Consensus {
...
-        let parent = match parent {
-            Ok(Some(parent)) => parent,
-            _ => {
-                warn!("parent not found during committee_for_hash");
-                return Err(anyhow!("parent not found during committee_for_hash"));
-            }
-        };
+        let committee = self.state.get_stakers()?;
+
+        self.state.set_to_root(H256(curr_root_hash.0));
```
theo-zil, line 947: Requiring `&mut self` can be avoided by using `state.at_root()` rather than `state.set_to_root()`. Keeping it const should allow eliminating `mut` in other places up the chain and I think is definitely worthwhile.

zilliqa/src/consensus.rs
```diff
@@ -1335,7 +1372,17 @@ impl Consensus {
 
     /// Check the validity of a block
     fn check_block(&mut self, block: &Block) -> Result<()> {
-        block.verify_hash()?;
+        let Some(parent) = self.get_block(&block.parent_hash())? else {
```
theo-zil, line 1375: What was the reason for moving the parent check above the `block.view == 0` check?
If you think it's useful, then I think a comment should be added to the function that it cannot be used for genesis, and the `if block.view() == 0` block below can be deleted outright.

But I think having that short-circuit in case it's ever called for genesis is a nice thing to have. If were to want to prevent `check_block` from ever being applied to genesis, then an explicit check returning an dedicated error (rather than `MissingBlockError`) would be better.

zilliqa/src/consensus.rs
```diff
@@ -1694,22 +1733,43 @@ impl Consensus {
...
     }
 
-    pub fn leader(&self, committee: &Committee, view: u64) -> Validator {
+    pub fn leader_at_block(&mut self, block: &Block, view: u64) -> Validator {
+        let curr_root_hash = self.state.root_hash().unwrap();
+        let block_root_hash = block.state_root_hash();
+
+        self.state.set_to_root(H256(block_root_hash.0));
+        let leader = self.leader(view);
+        self.state.set_to_root(H256(curr_root_hash.0));
```
theo-zil, line 1742: Similarly, here also using `state.at_root()` would allow eliminating the `mut`, which would avoid `mut`s up the chain - including changes to the RPCs to require `mut Node` that this PR makes due to this function taking `mut self`

zilliqa/src/consensus.rs
```diff
@@ -1694,22 +1733,43 @@ impl Consensus {
...
+        leader
+    }
+
+    pub fn leader(&mut self, view: u64) -> Validator {
```
theo-zil, line 1746: This `mut` looks unnecessary too - I'm surprised `clippy` doesn't seem to highlight it

zilliqa/src/exec.rs
```diff
@@ -316,6 +316,18 @@ impl State {
...
     }
 
+    pub fn get_stakers_at_block(&mut self, block: &Block) -> Result<Vec<NodePublicKey>> {
+        let curr_root_hash = self.root_hash()?;
+        let block_root_hash = block.state_root_hash();
+        self.set_to_root(H256(block_root_hash.0));
+
+        let committee = self.get_stakers();
+
+        self.set_to_root(H256(curr_root_hash.0));
```
theo-zil, line 326: As above, `self.at_root()` to eliminate `mut`

theo-zil (changes requested):
Very nice change, a few comments about style and a couple of cleanups but otherwise looking forward to getting this in

theo-zil (dismissed):

zilliqa/src/consensus.rs
```diff
@@ -345,39 +361,47 @@ impl Consensus {
...
+        Ok(Some((
+            Some(peer_id),
+            ExternalMessage::CommitteeJoined(self.public_key()),
+        )))
+    }
+
+    // This is a callback method that is called by a 'joining' node
+    // It happens in reply to JoinCommittee message (triggered via CommitteeJoined)
+    // It works as follows: Peer wants to join, it broadcasts JoinCommittee. Each committee member
+    // that receives this message replies with its own peer_id and pub_key by sending CommiteeJoined
```
JamesHinshelwood, line 388: Maybe I'm misunderstanding some of the solution, but couldn't we store all of this data in the contract and remove the need for the two P2P messages?

zilliqa/src/consensus.rs
```diff
@@ -572,16 +596,26 @@ impl Consensus {
                 );
             }
 
-            if !block.committee.iter().any(|v| v.peer_id == self.peer_id()) {
-                trace!(
+            // Get possibly updated list of stakers
+            let Ok(stakers) = self.state.get_stakers_at_block(&block) else {
+                return Ok(None);
```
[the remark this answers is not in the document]
JamesHinshelwood: Agreed that we shouldn't swallow this error.

JamesHinshelwood (commented):
Sorry, a quick clarification comment about the general approach. Also I agree with Theo's unsolved comments.

zilliqa/src/consensus.rs
theo-zil: I can agree with this approach - the peer_id is a networking address and storing it in the contract feels wrong.

On the other hand, on further thought, the peer_id is derived from the node's private key anyway, just as the pubkey is. So a node with a given pubkey is always going to have the same peer_id; and the only reason we need to store both is because there's no way to derive one from the other without knowing the privkey. (There isn't, right? If there was one it'd be very nice.) So maybe storing it in the contract is fine? It'd reduce the complexity of the node's state machine, too. @JamesHinshelwood thoughts?

JamesHinshelwood: The peer ID is a node's identity on the network. A single peer ID can have multiple IP addresses which change over the course of time - `libp2p` handles this mapping (at least that's my understanding anyway).

And yeah - The peer ID is a deterministic function of the secret key, just like the public key.

IMO we should store it in the contract state. We will have the two identities problem wherever we store them, because there are two ways of identifying a node, depending on if you are at the network level (peer ID) or at the consensus level (BLS public key). If we still keep some of the data _outside_ of the contract (like in the current state of this PR), then we still need network-level messages to maintain that data. If we move everything inside the contract, that complexity disappears and nodes don't even need to be able to 'join' themselves. A node can simply keep synced with the chain, notice when its own identity turns up in the consensus committee (because the user externally submitted a transaction to add it with some stake) and then start participating.

zilliqa/src/exec.rs
JamesHinshelwood: FWIW I think we should only have `get_stakers_at_block` and we should remove `get_stakers` (at the implicit block that someone else set before you called this and may as well be a random value).

JamesHinshelwood: I would also prefer if the state root hash manipulation happened one level up in the call stack - i.e. we manipulate the `State` in `Consensus` to set it to the right height, then call `.get_stakers()`.

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
JamesHinshelwood: From reading `get_stakers_at_block` I think the only time it would return an `Err` is if something was fatally wrong (e.g. we weren't able to read data from our disk because someone drove their forklift through it). In these cases we should definitely propagate the error upwards. I think the same applies for the other cases Theo pointed out too.

zilliqa/src/consensus.rs
theo-zil: My only thought is that conceptually, network and blockchain identities might be different. Sure, right now they're derived from the same private key, but what if someone modified their node software such that the peer ID uses a different private key - and then ran multiple nodes that use the same on-chain priv/pub key identity (and can therefore do things like authorize each other's transactions, participate in consensus together, etc.), but have different peer IDs?

But, yeah, on further thought I think this is bikeshedding and also needless complexity at this point. Other things will probably break already if you do this, and we can probably just say that we don't support it instead, and link both identities together at the protocol level and store them in the contract, eliminating the network complexity.

theo-zil (approved):

zilliqa/tests/it/consensus.rs
```diff
@@ -24,7 +24,7 @@ async fn network_can_die_restart(mut network: Network) {
                 let index = n.random_index();
                 n.get_node(index).get_finalized_height() >= start_block
             },
-            50,
+            100,
```
JamesHinshelwood, line 27: Are all the changes to the tests still necessary?
If they make things more reliable then :+1:

zilliqa/src/contracts/deposit.sol
```diff
@@ -68,4 +72,12 @@ contract Deposit {
     function getStakers() public view returns (bytes[] memory) {
         return _stakerKeys;
     }
+
+    function getPeerId(bytes calldata blsPubKey) public view returns (bytes memory) {
```
JamesHinshelwood, line 76: Maybe an optimisation for the future, but potentially we should have a single method that returns `(stake, rewardAddress, peerId)` to limit the number of EVM calls we need to make?

zilliqa/src/contracts/deposit.sol
```diff
@@ -68,4 +72,12 @@ contract Deposit {
...
+        require(blsPubKey.length == 48);
+        if (_stakersMap[blsPubKey].rewardAddress == address(0)) {
```
JamesHinshelwood, line 78: Small nit, but I'd prefer if we checked `_stakersMap[blsPubKey].peerId` here for zero-ness.
We plan on allowing zero addresses later, which will involve removing the check in `getRewardAddress`, but that would cause a problem here (see #803)

zilliqa/src/contracts/mod.rs
```diff
@@ -111,7 +113,7 @@ mod tests {
     use ethers_solc::CompilerOutput;
 
     #[test]
-    #[cfg_attr(not(feature = "test_contract_bytecode"), ignore)]
+    //#[cfg_attr(not(feature = "test_contract_bytecode"), ignore)]
```
JamesHinshelwood, line 116: Did you mean to comment this out?

zilliqa/src/consensus.rs
```diff
@@ -1887,27 +1840,47 @@ impl Consensus {
...
+                .get_stake(*pub_key)
                 .unwrap()
                 .context("Committee member has no stake")
                 .unwrap();
             stake.get()
         }))
         .unwrap();
         let index = dist.sample(&mut rng);
-        committee.iter().nth(index).unwrap()
+        let public_key = *committee.get(index).unwrap();
```
JamesHinshelwood, line 1864: Maybe another performance point for the future - We could avoid a 2nd EVM call here if `getStakers` returned the `(pubKey, rewardAddress, peerId, stake, etc)` instead of just the pub key?

JamesHinshelwood (approved):

PR #828, opened 2024-04-10
CI: run rust fmt before kicking off the test suite, for fail-early behaviour

theo-zil:
Avoids the risk of having CI run for 10-15 minutes only to then fail on a rustfmt mismatch.

Assuming there isn't any deeper reason for why it was put after `cargo test` in the first place?


JamesHinshelwood:
#502 :stuck_out_tongue:

PR #829, opened 2024-04-12
Make perf tool work on zilliqa2 code base

chetan-zilliqa:
Added files `InitSigners.ts` and `SignersHelpers.ts` from the zq1 code base to make perf work and the rest are small tweaks.


rrw-zilliqa (approved):
Looks good to me :-)

PR #836, opened 2024-04-16
Implement `ots_searchTransactions{Before,After}`

JamesHinshelwood:
This is quite similar to the old implementation that I removed. When voting on a block proposal, we run transactions with an `Inspector` which records the 'touched' addresses and stores them in a separate index. This occurs for both EVM and Scilla transactions.

### DIFF
--- a/zilliqa/src/api/ots.rs
+++ b/zilliqa/src/api/ots.rs
@@ -8,7 +8,7 @@ use super::{
     eth::{get_transaction_inner, get_transaction_receipt_inner},
     types::ots,
 };
-use crate::{crypto::Hash, message::BlockNumber, node::Node, state::Contract};
+use crate::{crypto::Hash, message::BlockNumber, node::Node, state::Contract, time::SystemTime};
 
 pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
     super::declare_module!(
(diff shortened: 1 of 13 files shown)


zilliqa/src/api/ots.rs
```diff
@@ -124,3 +126,122 @@ fn has_code(params: Params, node: &Arc<Mutex<Node>>) -> Result<bool> {
...
+        last_page,
+    })
+}
+
+fn search_transactions_after(params: Params, node: &Arc<Mutex<Node>>) -> Result<ots::Transactions> {
+    let mut params = params.sequence();
+    let address: H160 = params.next()?;
+    let block_number: u64 = params.next()?;
+    let page_size: usize = params.next()?;
+
```
rrw-zilliqa, line 228: Does block_number==0 => most recent block? 
(yes, it's useless, but is it what's expected?)

rrw-zilliqa (approved):
Looks good to me :-)

zilliqa/src/api/ots.rs
JamesHinshelwood: It depends on if you're searching forwards or backwards. The docs say:

> A value of 0 means you want to search from the most recent block (ots_searchTransactionsBefore) or from the genesis (ots_searchTransactionsAfter).

The actual list of transactions in the UI displays in reverse order by default, so the client makes an initial request to `ots_searchTransactionsBefore` with `block_number = 0` to get the latest transactions.

I'm not sure this is useless, so I'll leave this unresolved in case I've misunderstood you?

PR #837, opened 2024-04-16
Implement `ots_getContractCreator`

JamesHinshelwood:
This API returns the 'creator' of a given contract. For Scilla contracts and most EVM contracts this is trivial - It is just the `from_addr` of the transaction which created the contract.

However, some EVM contracts are created by other contracts.

Therefore, we implement this by looping through the transactions which 'touched' the contract address and replaying each transaction. When replaying, we use a new `Inspector` to record the creator of a contract if one is created. When we find the transaction which created this contract, we can terminate.

I haven't added a test case for creating a Scilla contract yet, because I don't want to slow the test suite down more. :(


JamesHinshelwood:
Review after #836 is merged.

rrw-zilliqa (approved):
Looks good to me :-)

PR #838, opened 2024-04-16
Implement `ots_traceTransaction`

JamesHinshelwood:
This replays the specified transaction with a custom `Inspector` that produces a trace in the desired format.

I haven't added a test case for creating a Scilla contract yet, because I don't want to slow the test suite down more. :(


JamesHinshelwood:
Review after #837 is merged.

rrw-zilliqa (approved):
Looks good to me :-)

PR #841, opened 2024-04-17
Add `accepted`, `errors` and `exceptions` to Scilla receipts

JamesHinshelwood:
```rust
/// If the transaction was a call to a Scilla contract, whether the called contract accepted the ZIL sent to it.
pub accepted: Option<bool>,
/// Errors from calls to Scilla contracts. Indexed by the call depth of erroring contract.
pub errors: BTreeMap<u64, Vec<ScillaError>>,
/// Exceptions from calls to Scilla contracts.
pub exceptions: Vec<ScillaException>,
```


rrw-zilliqa (approved):
Looks good to me :-)

PR #845, opened 2024-04-17
Implement `GetBlockchainInfo`

JamesHinshelwood:
I left a bunch of the fields at `0` because calculating them is hard and requires us to store extra data. #860 tracks fixing them.


rrw-zilliqa (approved):
Looks good to me - at least it's better than the previous implementation.

PR #848, opened 2024-04-18
Fix precision of account balances provided to Scilla

JamesHinshelwood:
Previously, when the Scilla interpreter queried an account's balance, we would return the result in units of (10^-18) ZILs. Scilla expects the result in units of (10^-12) ZILs.

To mitigate similar bugs in the future, I have introduced a newtype for quantities of ZIL expressed in units of (10^-12) ZILs: `ZilAmount`. The intention is that we continue to use `u128` in most places (e.g. for internal storage of amounts and for EVM transactions), where we continue to expressed amounts as (10^-18) ZILs. But for places we know we want amounts expressed as (10^-12) ZILs (such as the Zilliqa API requests and responses), we use `ZilAmount` to make that clear. A `ZilAmount` can't be converted to a `u128` without multiplying it by `(10^6)` and vice-versa.

### DIFF
--- a/zilliqa/src/api/types/zil.rs
+++ b/zilliqa/src/api/types/zil.rs
@@ -10,7 +10,9 @@ use crate::{
     schnorr,
     serde_util::num_as_str,
     time::SystemTime,
-    transaction::{ScillaLog, SignedTransaction, TransactionReceipt, VerifiedTransaction},
+    transaction::{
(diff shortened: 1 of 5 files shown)


rrw-zilliqa (dismissed):
Looks good to me :-)

I would've slightly preferred ZilAmount and EthAmount, but probably not practical given the number of external libraries that assume u128 === wei

rrw-zilliqa (approved):
Looks good to me :-)

PR #851, opened 2024-04-19
Use stable rust compiler in CI for everything except formatting

JamesHinshelwood:
We were using nightly everywhere, because `cargo fmt` requires it with our `rustfmt.toml` configuration. However, this causes other unintended problems, such as:
* Code failing on new nightly clippy lints which aren't yet stable
* Libraries using latest nightly features when they detect a nightly compiler.

Instead, we run compilation and tests with the normal stable compiler that is specified in `rust-toolchain.toml`.


rrw-zilliqa (approved):
Looks good to me :-)

PR #853, opened 2024-04-22
Implement `ots_getTransactionError`

JamesHinshelwood:


rrw-zilliqa (approved):
Looks good to me :-)

Though one day we will need to cope with txns which have both exceptions and reverts, probably .. (eg. for cross-domain calls)

PR #858, opened 2024-04-23
Charge gas for Scilla transactions

JamesHinshelwood:
I added newtypes for both `EvmGas` and `ScillaGas` to prevent confusion between them and ensure we convert at the correct exchange rate (420:1).


rrw-zilliqa (approved):
Looks good to me :-)

PR #861, opened 2024-04-23
Make `eth_getTransactionCount` respect the `pending` block number

JamesHinshelwood:
When the block number is `pending`, we also count the transactions in the mempool.

### DIFF
--- a/zilliqa/src/node.rs
+++ b/zilliqa/src/node.rs
@@ -463,6 +463,25 @@ impl Node {
             .get_account(address)
     }
 
+    /// Return the `nonce` of an account. If the `block_number` is [BlockNumber::Pending], any pending transactions in
+    /// our pool are also added to the nonce.
+    pub fn get_transaction_count(
+        &self,
+        address: Address,
+        block_number: BlockNumber,
+    ) -> Result<u64> {
+        let nonce = self.get_account(address, block_number)?.nonce;
+
+        if matches!(block_number, BlockNumber::Pending) {
+            Ok(self
+                .consensus
+                .transaction_pool
+                .pending_nonce(address, nonce))
+        } else {
+            Ok(nonce)
+        }
+    }
+
     pub fn get_account_storage(
         &self,
         address: Address,
--- a/zilliqa/src/pool.rs
+++ b/zilliqa/src/pool.rs
@@ -192,6 +192,21 @@ impl TransactionPool {
         std::mem::take(&mut self.transactions).into_values()
     }
 
+    /// Calculate the effective nonce of the sender, if all the currently pending transactions in the pool were mined
+    /// in order. Effectively, this calculates the current nonce plus the number of transactions with sequential nonces
+    /// in the pool.
+    pub fn pending_nonce(&self, sender: Address, mut current_nonce: u64) -> u64 {
+        loop {
+            let index = TxIndex::Nonced(sender, current_nonce);
+            if self.transactions.get(&index).is_none() {
+                break;
+            }
+            current_nonce += 1;
+        }
+
+        current_nonce
+    }
+
     pub fn size(&self) -> usize {
         self.transactions.len()
     }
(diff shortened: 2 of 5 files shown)


zilliqa/src/pool.rs
```diff
@@ -192,6 +192,21 @@ impl TransactionPool {
         std::mem::take(&mut self.transactions).into_values()
     }
 
+    /// Calculate the effective nonce of the sender, if all the currently pending transactions in the pool were mined
+    /// in order. Effectively, this calculates the current nonce plus the number of transactions with sequential nonces
+    /// in the pool.
+    pub fn pending_nonce(&self, sender: Address, mut current_nonce: u64) -> u64 {
+        loop {
+            let index = TxIndex::Nonced(sender, current_nonce);
```
JamesHinshelwood, line 200: @theo-zil I could do with a double check of this logic from you. We basically ignore any nonce-less transactions in the mempool. Does that seem correct?

(by the way, the use case for this API is that Eth clients call it between each transaction to set the nonce of the next one)

zilliqa/src/node.rs
```diff
@@ -450,6 +450,25 @@ impl Node {
...
 
+    /// Return the `nonce` of an account. If the `block_number` is [BlockNumber::Pending], any pending transactions in
+    /// our pool are also added to the nonce.
+    pub fn get_transaction_count(
+        &self,
+        address: Address,
+        block_number: BlockNumber,
+    ) -> Result<u64> {
+        let nonce = self.get_account(address, block_number)?.nonce;
+
```
rrw-zilliqa, line 461: What happens when you've put the transactions into a block, but that block is not yet confirmed? Will get_account(address, block_number) respect the pending block? If not, you'll get the wrong nonce here and thus think there are no consecutive nonces in the mempool, whereas actually (once the pending block is confirmed) there are ..

rrw-zilliqa (approved):
Approving, because it's better than the previous implementation, but ..

zilliqa/src/node.rs
JamesHinshelwood: Oof good point, thank you for spotting. `get_account` won't respect the pending block so I believe your bug exists. Just to be clear - When we build a block proposal, we remove transactions from our `TransactionPool`.

Just thinking out loud - What if we keep track of any pending block `Proposal`s in `Consensus` which we've sent out, but haven't been voted on yet. Then any `block_number = Pending` APIs can query from there instead of the tip of the chain?

Tagging @theo-zil and @DrZoltanFazekas for thoughts as well.

JamesHinshelwood: Maybe an alternative would be not removing transactions from the mempool until they are mined in a block? We could remove them and re-add them when we're building a `Proposal`?

theo-zil: Hmm - I already do this only for intershard transactions (see the bottom of the `vote()` definition), so this honestly sounds like a solution. Just remove the redundant re-insertion of `opaque_transactions` inside `vote()`.

zilliqa/src/pool.rs
theo-zil: Yup, looks good to me.
I mean, the name of the RPC is wrong for Zilliqa, since what it really returns is the pending nonce and not the transaction count; on eth they're the same but we break that assumption. But since eth software mostly cares about the nonce, we have to follow that usecase even though it doesn't match the name.

theo-zil (dismissed):
Approved, except pending the fix for pending blocks as discussed

zilliqa/src/node.rs
rrw-zilliqa: I think it'll work - certainly the right approach - but myself, I'd probably mark transactions explicitly as having gone into a block rather than relying on the nonce comparison, partly for visibility (which might be a bad idea, since denormalising), and partly in case someone other than this code spotted that they were invalid and ejected them from the mempool in the meantime, which would result in this bug reoccurring as a heisenbug... but that's just me.

JamesHinshelwood: I'm not sure what you mean by "relying on the nonce comparison"? I don't think it will be simpler to mark transactions in the mempool as having gone into a proposal, since then we need to deal with the case where they go into the proposal, but don't get mined and that flag needs to be reset.

I've implemented the solution discussed above where we retain transactions in the mempool until they are mined.

rrw-zilliqa (approved):
Looks good to me :-)

PR #864, opened 2024-04-24
Implement `GetSmartContracts`

JamesHinshelwood:
This API returns the list of smart contracts deployed by a given account.

I have chosen to replicate the ZQ1 implementation and exclude EVM contracts, but maybe we should include them? In this case, should we just include direct creations or also creations via other contracts?

### DIFF
--- a/zilliqa/src/api/zil.rs
+++ b/zilliqa/src/api/zil.rs
@@ -14,11 +14,12 @@ use serde_json::{json, Value};
 
 use super::{
     to_hex::ToHex,
-    types::zil::{self, BlockchainInfo, ShardingStructure},
+    types::zil::{self, BlockchainInfo, ShardingStructure, SmartContract},
 };
 use crate::{
     api::types::zil::{CreateTransactionResponse, GetTxResponse},
     crypto::Hash,
+    exec::zil_contract_address,
     message::BlockNumber,
     node::Node,
     schnorr,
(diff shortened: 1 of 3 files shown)


zilliqa/src/api/zil.rs
```diff
@@ -378,3 +380,34 @@ fn get_tx_block(
...
+        .lock()
+        .unwrap()
+        .get_account(address, BlockNumber::Latest)?
+        .nonce;
+
+    let mut contracts = vec![];
+
+    for i in 0..nonce {
+        let contract_address = zil_contract_address(address, i);
+        // TODO: Should we expose created EVM contracts in this API too?
```
JamesHinshelwood, line 397: Resolve this.

rrw-zilliqa: (will leave you to close the conversation if you feel happy)

JamesHinshelwood: Will remove the `TODO` and resolve, thanks!

zilliqa/src/api/zil.rs
```diff
@@ -378,3 +380,34 @@ fn get_tx_block(
...
+
+    let nonce = node
+        .lock()
+        .unwrap()
+        .get_account(address, BlockNumber::Latest)?
+        .nonce;
+
+    let mut contracts = vec![];
+
+    for i in 0..nonce {
```
theo-zil, line 395: This feels like the kind of thing a hypothetical indexer would cache in its database rather than looping through every integer up to the nonce to check the address, on every RPC call.

The fact that it's a required RPC for zilliqa nodes (rather than optional functionality that could be enabled only for voluntary indexerers) makes me hesistant to say that we should therefore force every node to process and cache this in the database, though. I suppose it'll probably be fine (especially if we have a way to charge for API access anyway).

JamesHinshelwood: Yeah fair point. I copied the implementation from ZQ1. I am quite sure we will add configuration to enable/disable specific APIs soon and this might be one that is only necessary for nodes serving devex. We could also reasonably limit the number of returned transactions to bound the work here.

Issue #882, opened 2024-05-02
Otterscan's DS block view doesn't work against zq2


rrw-zilliqa:
.. partly because of the wrong field names in BlockChainInfo, partly because GetDsBlock is not yet implemented.

DrZoltanFazekas:
we'll remove the DS block list and only show TX blocks for now

yaron-zilliqa:
Duplicate of https://github.com/Zilliqa/otterscan/issues/39

PR #883, opened 2024-05-02
Fix more field names for GetBlockchainInfo

rrw-zilliqa:
relates to #882


JamesHinshelwood (approved):
FWIW you only need to rename the `DS` fields. The rest are covered by `#[serde(rename_all = "PascalCase")]`

rrw-zilliqa:
> FWIW you only need to rename the `DS` fields. The rest are covered by `#[serde(rename_all = "PascalCase")]`

True - I did think about that, but I thought that nailing them down was a bit more explicit than leaving the default to get them right - we don't really mean "make this PascalCase" - we mean "I want this exact name" . Not a big thing, but I was on a fence and had to get off on one side or the other!

PR #884, opened 2024-05-02
Exclude some `null` fields from `eth_getTransactionByHash`

JamesHinshelwood:
These fields are documentation as optional rather than nullable and this behaviour matches the implementation in ZQ1.


rrw-zilliqa (approved):
Looks good to me :-)

PR #885, opened 2024-05-02
Fix formatting in `GetTransaction` response

JamesHinshelwood:
* `version` should be a string.
* `senderPubKey` should be a compressed SEC1 encoed point.
* `signature` should be prefixed with `0x`.


rrw-zilliqa (approved):
Looks good to me :-)

PR #886, opened 2024-05-02
Support batch trie insertions into sled

JamesHinshelwood:
This makes large state manipulations faster to flush to disk. This is particularly useful for the ZQ1->ZQ2 state converter.


rrw-zilliqa (approved):
Looks good to me :-)

PR #887, opened 2024-05-02
Fix JS test suite

JamesHinshelwood:
The Scilla-image swizzling was pointless and caused the runner to run out of disk space. We will run Scilla properly once we start making these tests pass.


rrw-zilliqa (approved):
Looks good to me :-)

PR #898, opened 2024-05-07
Replace `primitive_types` with `alloy-primitives`

JamesHinshelwood:
Sorry for the large commit. 99% of it is just find-and-replace.

`alloy` is used internally by `revm`, so converting to it in our own code removes a load of conversion code between the two types.

Also, `alloy` has a bunch of useful types and utilities we can use to simplify our code. This commit replaces the
`Tx{Legacy,Eip{2930,1559}}` types we had with alloy's version. In future, we can also utilise the API types defined by alloy.

(specifically, I wanted to use the alloy-provided types for `trace_replayTransaction` which prompted this PR)


JamesHinshelwood:
Blocked by #899

rrw-zilliqa (approved):
Looks good to me :-)

PR #913, opened 2024-05-13
Implement `trace_replayTransaction`

JamesHinshelwood:
This mostly relies on the supporting code in alloy and `revm-inspectors`. I did have to do some work on our side to split the existing `TransactionApplyResult` into an `Evm` and `Scilla` variant, so we can get at the necessary data to build a replay for EVM transactions.

Support for Scilla transactions will come next.


rrw-zilliqa (approved):
Looks good to me :-)

PR #916, opened 2024-05-14
Refactor Scilla transaction execution

JamesHinshelwood:
Previously when we applied a Scilla transaction to state (via `State::apply_transaction_scilla`), the state would be mutated directly by execution engine while the transaction was run. This differed from the behaviour of EVM transactions, where the execution engine only read from the state and kept changes to state in memory (in a `HashMap<Address, Account>`). This PR updates Scilla execution to work in the same way.

The advantage of this approach is that we get a state diff when executing a transaction, which makes implementing APIs similar to `trace_replayTransaction` possible. It also avoids more roundtrips from/to the database, since each account is loaded at most once and then saved once at the end of execution.

### DIFF
--- a/zilliqa/src/exec.rs
+++ b/zilliqa/src/exec.rs
@@ -1,11 +1,12 @@
 //! Manages execution of transactions on state.
 
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{hash_map::Entry, BTreeMap, HashMap},
     error::Error,
     fmt::{self, Display, Formatter},
+    mem,
     num::NonZeroU128,
-    sync::Arc,
+    sync::{Arc, MutexGuard},
 };
 
 use alloy_primitives::{Address, U256};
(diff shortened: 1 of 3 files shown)


zilliqa/src/exec.rs
```diff
@@ -1047,3 +748,386 @@ pub fn zil_contract_address(sender: Address, nonce: u64) -> Address {
...
+    if output.accepted {
+        if let Some(result) = state.deduct_from_account(from_addr, txn.amount)? {
+            return Ok((result, state));
+        }
+
+        let to = state.load_account(txn.to_addr)?;
+        to.balance += txn.amount.get();
+    }
+
+    // TODO: Handle `output.messages` for multi-contract calls.
```
rrw-zilliqa, line 1099: Issue or it didn't happen ..

rrw-zilliqa (approved):
Looks good to me :-)

zilliqa/src/exec.rs
JamesHinshelwood: :saluting_face: #920

Issue #917, opened 2024-05-14
Implement dynamic documentation generation


rrw-zilliqa:
.. so that we can create zq2 documentation with the docs in the zq2 repository rather than separately maintained.

rrw-zilliqa:
I think we can now close this :-)

PR #931, opened 2024-05-16
#917 - API docs, try 3

rrw-zilliqa:
.. as advertised in slack.


z2/src/docgen.rs
```diff
@@ -144,6 +163,56 @@ fn insert_key(val: &mut serde_yaml::Value, components: &Vec<String>, idx: usize,
...
+        (
+            address!("7E5F4552091A69125d5DfCb7b8C2659029395Bdf"),
+            "5000000000000000000000".to_string(),
+        ),
+        // privkey db11cfa086b92497c8ed5a4cc6edb3a5bfe3a640c43ffb9fc6aa0873c56f2ee3
+        (
+            address!("cb57ec3f064a16cadb36c7c712f4c9fa62b77415"),
+            "5000000000000000000000".to_string(),
+        ),
+    ];
```
JamesHinshelwood, line 186: FWIW I'd prefer
```suggestion
    let genesis_accounts: Vec<(Address, String)> = vec![];
```
and removing https://github.com/Zilliqa/zq2/blob/96cd025e0f1efb5fe11e9aa9292183e101d17ab4/zilliqa/src/state.rs#L87-L89

There's no reason networks without any genesis accounts shouldn't be valid (except that on they aren't very useful).

z2/src/docgen.rs
```diff
@@ -144,6 +163,56 @@ fn insert_key(val: &mut serde_yaml::Value, components: &Vec<String>, idx: usize,
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Hash, Eq)]
+pub enum ApiMethod {
+    JsonRpc { name: String },
+    Rest { uri: String },
```
JamesHinshelwood, line 169: Do we need this? I'd avoid the enum if we don't have any rest endpoints. Easy to add in the future, but annoying to keep it around if we never use it.

z2/src/docgen.rs
```diff
@@ -202,7 +273,9 @@ impl Docs {
         println!("After remove {0}", serde_yaml::to_string(&contents_map)?);
 
         while let Some(ref path_entry) = stack.pop() {
```
JamesHinshelwood, line 275: Existing code but consider using https://docs.rs/walkdir/latest/walkdir/ to make your life easier.

JamesHinshelwood (approved):

z2/src/docgen.rs
rrw-zilliqa: We have the health check endpoint, and anything to do with websockets (which I will admit is going to be hard to include in this anyway)

z2/src/docgen.rs
rrw-zilliqa: Yeah - I did consider it, but felt that for the amount of effort, I should avoid adding trivial dependencies (every dependency being a supply chain risk)

z2/src/docgen.rs
rrw-zilliqa: Cool! That's been annoying me for a while, so will address in a future PR - https://github.com/Zilliqa/zq2/issues/932

PR #935, opened 2024-05-17
Estimate gas bsearch

bzawisto:
Do the gas estimation using binary search.

### DIFF
--- a/zilliqa/src/exec.rs
+++ b/zilliqa/src/exec.rs
@@ -678,11 +678,6 @@ impl State {
         let gas_price = gas_price.unwrap_or(GAS_PRICE);
 
         let mut max = gas.unwrap_or(MAX_EVM_GAS_LIMIT).0;
-
-        if let Some(gas) = gas {
-            max = gas.0;
-        }
-
         let upper_bound = max;
 
         // Check if estimation succeeds with the highest possible gas
@@ -698,11 +693,15 @@ impl State {
             value,
         )?;
 
-        // Execute the while loop iff min/max < MINIMUM_PERCENT_RATIO_FOR_MAX_AND_MIN [%]
-        const MINIMUM_PERCENT_RATIO_FOR_MAX_AND_MIN: u64 = 15;
+        // Execute the while loop iff (max - min)/max < MINIMUM_PERCENT_RATIO [%]
+        const MINIMUM_PERCENT_RATIO: u64 = 3;
 
         // result should be somewhere in (min, max]
-        while min < ((max * MINIMUM_PERCENT_RATIO_FOR_MAX_AND_MIN) / 100) {
+        while min < max {
+            let break_cond = (max - min) <= (max * MINIMUM_PERCENT_RATIO) / 100;
+            if break_cond {
+                break;
+            }
             let mid = (min + max) / 2;
 
             let ResultAndState { result, .. } = self.apply_transaction_evm(
@@ -726,21 +725,6 @@ impl State {
                     _ => return Err(anyhow!("halted due to: {reason:?}")),
                 },
             }
-        }
-
-        if max == upper_bound {
-            self.estimate_gas_inner(
-                from_addr,
-                to_addr,
-                data.clone(),
-                chain_id,
-                current_block,
-                EvmGas(max),
-                gas_price,
-                value,
-            )?;
-
-            return Ok(max);
         }
         debug!("Estimated gas: {}", max);
         Ok(max)


zilliqa/src/exec.rs
```diff
@@ -673,23 +676,101 @@ impl State {
         value: u128,
     ) -> Result<u64> {
         let gas_price = gas_price.unwrap_or(GAS_PRICE);
-        let gas = gas.unwrap_or(BLOCK_GAS_LIMIT);
 
+        let mut right = gas.unwrap_or(MAX_EVM_GAS_LIMIT).0;
+        let mut left = lower_bound_gas_estimate(to_addr, &data);
```
JamesHinshelwood, line 681: What if we instead generating the lower bound by first running the transaction with `gas_limit = BLOCK_GAS_LIMIT` and took the `gas_used` from the result?

JamesHinshelwood, line 681: nit: I'd find it clearer if these were called `min` and `max`.

zilliqa/src/exec.rs
```diff
@@ -673,23 +676,101 @@ impl State {
...
+            data.clone(),
+            chain_id,
+            current_block,
+            EvmGas(upper_bound),
+            gas_price,
+            value,
+        )?;
+
+        // We don't do deeper search if the range is smaller than the following param
+        const MINIMUM_ABSOLUTE_RANGE_DIFF: u64 = 1000;
```
JamesHinshelwood, line 702: Did we consider using a %-based diff instead of absolute? That was what I saw in other implementations. Happy to go with this if you've thought about it and think its better?

zilliqa/src/exec.rs
```diff
@@ -673,23 +676,101 @@ impl State {
         value: u128,
     ) -> Result<u64> {
         let gas_price = gas_price.unwrap_or(GAS_PRICE);
-        let gas = gas.unwrap_or(BLOCK_GAS_LIMIT);
 
+        let mut right = gas.unwrap_or(MAX_EVM_GAS_LIMIT).0;
```
JamesHinshelwood, line 680: Maybe a nitpick, but can we save some type-safety by keeping this as an `EvmGas`? You may need to implement some extra traits to do the arithmetic, but I think that's reasonable.

zilliqa/src/exec.rs
```diff
@@ -673,23 +676,101 @@ impl State {
...
+        let mut left = lower_bound_gas_estimate(to_addr, &data);
+
+        if gas.is_some() {
+            right = gas.unwrap().0;
+        }
```
JamesHinshelwood, line 685:
```suggestion
        if let Some(gas) = gas {
            right = gas.0;
        }
```

zilliqa/src/exec.rs
```diff
@@ -673,23 +676,101 @@ impl State {
...
+            to_addr,
+            data.clone(),
+            chain_id,
+            current_block,
+            EvmGas(upper_bound),
+            gas_price,
+            value,
+        )?;
+
+        // We don't do deeper search if the range is smaller than the following param
```
JamesHinshelwood, line 701: wording nitpick, sorry
```suggestion
        // We don't do deeper search if the difference between `min` and `max` is smaller than the following param
```

zilliqa/src/exec.rs
JamesHinshelwood: In fact, I think you can just set `left`/`min` to `gas_used` from your "Check if estimation succeeds with the highest possible gas" call below?

JamesHinshelwood: Can we also add in a comment if these are inclusive or exclusive? (I think `left` is inclusive and `right` is exclusive but I'm not sure?)

zilliqa/src/exec.rs
```diff
@@ -673,23 +676,101 @@ impl State {
...
+                data.clone(),
+                None,
+                chain_id,
+                current_block,
+                inspector::noop(),
+            )?;
+
+            match result {
+                ExecutionResult::Success { .. } => right = mid,
+                ExecutionResult::Revert { .. } => left = mid + 1,
```
JamesHinshelwood, line 722: Is a revert possible if the transaction runs out of gas? I though revert would only be for explicit reverts in the contract?

zilliqa/src/exec.rs
```diff
@@ -673,23 +676,101 @@ impl State {
...
+                data.clone(),
+                chain_id,
+                current_block,
+                EvmGas(right),
+                gas_price,
+                value,
+            )?;
+
+            return Ok(right);
+        }
```
JamesHinshelwood, line 743: What's the reason for this extra block?

evm_scilla_js_tests/contracts/scilla/GasCreateTest.sol
JamesHinshelwood, line 1: We aren't actually running this test suite so tests here aren't as valuable as they should be :stuck_out_tongue: 

Can we instead write this test in Rust in `zilliqa/tests/it/`? There should be examples in there of deploying and calling a contract - Let me know if not.

config-single.toml
```diff
@@ -0,0 +1,19 @@
+p2p_port = 0
+
+[[nodes]]
+json_rpc_port = 4201
+eth_chain_id = 0x8001
+consensus.consensus_timeout = { secs = 5, nanos = 0 }
+consensus.genesis_committee = [ ["b27aebb3b54effd7af87c4a064a711554ee0f3f5abf56ca910b46422f2b21603bc383d42eb3b927c4c3b0b8381ca30a3", "12D3KooWESMZ2ttSxDwjfnNe23sHCqsJf6sNEKwgHkdgtCHDsbWU"] ]
```
JamesHinshelwood, line 7: Side question - What is `genesis_committee` used for now that we've moved the committee into `deposit.sol`? I would assume that `genesis_deposits` makes it redundant?

JamesHinshelwood (approved):

zilliqa/src/exec.rs
JamesHinshelwood: Yeah fair enough, I guess its completely local to this function.

zilliqa/src/exec.rs
JamesHinshelwood: FWIW, `reth` uses 1.5%

JamesHinshelwood (dismissed):

config-single-node.toml
```diff
@@ -9,7 +9,7 @@ consensus.genesis_deposits = [
 ]
 consensus.genesis_accounts = [
     # Accounts with private key 0x1, 0x2, 0x3, 0x4.
-    ["7E5F4552091A69125d5DfCb7b8C2659029395Bdf", "5000000000000000000000"],
+    # Don't use account with priv key = 0x1 since it's also used for voting rewards
```
JamesHinshelwood, line 12: Why is it a problem to have it in both?

JamesHinshelwood (approved):

config-single-node.toml
JamesHinshelwood: Because the test assertions don't work, thanks :)

Issue #937, opened 2024-05-17
Create the docker image for zq2 node


mauromedda:
This ticket is to track the activity related to create the container images for the ZQ2 validator node.

The Docker image will contain zilliqa2 and scilla binaries.
They will be pushed to our GCP artifactory the following registry based on the below strategy:
(private) registry asia-docker.pkg.dev/prj-p-devops-services-tvwmrf63/zilliqa-private/zq2 
(public)  registry asia-docker.pkg.dev/prj-p-devops-services-tvwmrf63/zilliqa-public/zq2 
(public)  https://hub.docker.com/r/zilliqa/zq2.

Here the image versioning strategy:
merge to main -> will build and push it to our private repo an image tagged with the first 8 character of the commit-id (we will keep only 10 versions of these images)
tag to release/v* -> will build and push an image tagged with the given version tag to the public repo.

the release/v* tag will be protected and only people with given repo admin role will be able to create public releases.

rrw-zilliqa:
I'm usually against providing binaries publically - by all means build them, but allowing folks to pick up binaries results in dependencies on weird commits (so you can never remove the binaries) and on dependencies on non-recreatable objects. Specifically , in ZQ1 it's resulted in folks trying to run containers on incompatible kernels and then calling for support when it didn't work. Perhaps we should either trim the public repos quite hard (which will make anyone who depends on our imags break), or only push things we intend to keep forever...

mauromedda:
Thanks for the feedback @rrw-zilliqa :) Let's push the merge to main to internal / private repo so we can test eventually (and clean any 10 images) and push on public repo only the intend to release and kept forver aka release/v*

rrw-zilliqa:
That's what I'd do :-). One thiing it might be worth thinking about is building the image and then using that to run the CI/CD tests (not sure if gha is good at that). Just an idea..

PR #938, opened 2024-05-17
(feat): Create the docker image for zq2 node

mauromedda:
Rudimentary process to build and push the zq2 docker images as per issue #937 defines.

Will open follow-up issues for:
1. add an image test step
2. export the version as env variable so the software can use it as input for specifying the version
3. use the image to run the tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM rust:1.78.0-slim-bullseye as builder
 
-ARG is_ci=false
+ARG is_release=false
 RUN apt update -y && \
     apt upgrade -y && \
     apt install -y protobuf-compiler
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM rust:1.78.0-slim-bullseye as builder
 
+ARG is_ci=false
```
JamesHinshelwood, line 3: Maybe rename this flag to `is_release` or `release`? I think its clearer if the `Dockerfile` exposes the configuration as-is and the CI say "I will build with release"?

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #939, opened 2024-05-17
Implement `eth_subscribe` for `newPendingTransactions`

JamesHinshelwood:


zilliqa/src/api/types/eth.rs
```diff
@@ -158,6 +159,67 @@ pub struct Transaction {
...
+            v,
+            r,
+            s,
+            chain_id: transaction.chain_id(),
+            access_list: transaction.access_list().map(|a| a.to_vec()),
+            transaction_type: match transaction {
+                transaction::Transaction::Legacy(_) => 0,
+                transaction::Transaction::Eip2930(_) => 1,
+                transaction::Transaction::Eip1559(_) => 2,
+                // Set Zilliqa transaction types to a unique number. This is "ZIL" encoded in ASCII.
```
rrw-zilliqa, line 214: Is there a registry for these? If so, worth referencing. If not, (shall I?) file an issue to reference them in the docs when that PR gets merged?

zilliqa/src/api/eth.rs
```diff
@@ -862,7 +808,32 @@ async fn subscribe(
...
+            let full = match params {
+                pubsub::Params::None => false,
+                pubsub::Params::Bool(b) => b,
+                pubsub::Params::Logs(_) => {
+                    return Err("invalid params for newPendingTransactions".into());
+                }
+            };
+
+            if full {
+                let mut txns = node.lock().unwrap().subscribe_to_new_transactions();
```
rrw-zilliqa, line 821: I'd slightly campaign for this to be called subscribe_to_new_pending_transactions() , since otherwise it's ambiguous whether these are final or pending txns (or if you need both, use an enum in the args). Don't feel strongly, but the more ambiguity we allow to creep in, the higher the cognitive load of reading the code will get...

zilliqa/src/api/types/eth.rs
```diff
@@ -158,6 +159,67 @@ pub struct Transaction {
...
+                (transaction.max_fee_per_gas(), None, None)
+            }
+            transaction::Transaction::Eip1559(TxEip1559 {
+                max_fee_per_gas,
+                max_priority_fee_per_gas,
+                ..
+            }) => (
+                // The `gasPrice` for EIP-1559 transactions should be set to the effective gas price of this transaction,
+                // which depends on the block's base fee. We don't yet have a base fee so we just set it to the max fee
+                // per gas.
```
rrw-zilliqa, line 184: Worth an issue if we're going to have a base fee in the future, or this will turn up as a bug when we do ..

rrw-zilliqa (approved):
Looks good to me :-)

zilliqa/src/api/types/eth.rs
JamesHinshelwood: #961

zilliqa/src/api/eth.rs
JamesHinshelwood: Yeah fair point, will update in a follow-up PR.

zilliqa/src/api/types/eth.rs
JamesHinshelwood: #962

PR #957, opened 2024-05-21
Removing genesis committee option

bzawisto:
As spotted in other review - we no longer need to provide genesis committee in config file since all of the validators are kept in deposit contract.


rrw-zilliqa (approved):
Looks good to me :-)

PR #960, opened 2024-05-21
(feat): run zq2 on containerized infrastructure

mauromedda:
the ZQ2 networks deployed on GCP will run on containers and it will enable also the support for scilla.
The IaC will now let us the possibility to deploy a set of core components (bootstrap+validators+apps) on a single region and optionally a set of decentralized validators in a multi-region fashion.
It also allows us to run multi-zonal deployment of ZQ2 nodes.
Removed the k8s dependencies. Now all the components are deployed on GCE VM by a provisioning script that selects which components install and run based on a given role value: validator or apps.


infra/tf/modules/node/main.tf
```diff
@@ -168,6 +175,21 @@ sudo systemctl restart google-cloud-ops-agent
 apt update -y
 apt install -y pigz
 
+# Add Docker's official GPG key:
+apt install -y ca-certificates curl
+install -m 0755 -d /etc/apt/keyrings
+curl -fsSL https://download.docker.com/linux/debian/gpg -o /etc/apt/keyrings/docker.asc
```
rrw-zilliqa, line 181: You really want to put this in the repo, or anyone who compromises download.docker.com can compromise your image, which rather obviates the point of checking signatures.

infra/tf/main.tf
```diff
@@ -229,6 +203,10 @@ data "google_compute_global_address" "api" {
   name = "api-${replace(var.subdomain, ".", "-")}"
 }
 
+data "google_compute_global_address" "bootstrap" {
+  name = "boostrap-${replace(var.subdomain, ".", "-")}"
```
rrw-zilliqa, line 207: Did you want to call this bootstrap?

rrw-zilliqa (approved):
Looks good to me (modulo comments)

infra/tf/modules/node/main.tf
mauromedda: Hey Richard, not really. I just did a proof of concept to see if the containerized version works. This has to be refactored properly Monday when I'm back. I just make this as draft pr to record what I'm doing

infra/tf/main.tf
mauromedda: As per above, I kept the current naming to do a test on the containered version. Bootstrap is just how has been named this node currently

JamesHinshelwood: @rrw-zilliqa What made you think that bootstrap was the incorrect name? I'm not against renaming, but in my mind bootstrap was an accurate name. So I'm worried that we have a different understanding of what this node does.

I thought bootstrap was a good name to describe the node which new nodes contact in order to learn about the existence of the rest of the network.

rrw-zilliqa: bootstrap is fine. "boostrap" seems odd :-p

pavlops (approved):

pavlops (approved):

JamesHinshelwood (approved):

PR #970, opened 2024-05-24
Implement `iter_by_prefix` for `EthTrie`

JamesHinshelwood:
This is a pre-requisite for flattening Scilla storage and I figured it would make things smoother to submit it as a separate commit first.

I've tried to include explanatory comments in the code itself - Do complain if the implementation isn't clear as it is easy to get wrong and thus very possible that I have messed something up.


rrw-zilliqa (approved):
Looks good to me :-)

Issue #981, opened 2024-05-27
Refactor message hashing/signing


JamesHinshelwood:
For all messages that include signatures:

* Move the code around to ensure that we're only calculating those signatures in a single place. See https://github.com/Zilliqa/zq2/issues/981#issuecomment-2134506270 for an example of where we're doing it in multiple places right now.
* Ensure all fields in the message are signed over for the signature.

DrZoltanFazekas:
`BlockHeader` hash doesn't include the `timestamp` either, but it should:
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/message.rs#L586

DrZoltanFazekas:
Btw, computing the block hash is duplicated at least 4 times - shouldn't we add a function which implements it at one place?
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/message.rs#L586
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/message.rs#L618
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/message.rs#L654
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/message.rs#L690

DrZoltanFazekas:
Imo, `AggregateQC` hash should include the view number too:
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/message.rs#L359

DrZoltanFazekas:
There are a lot more things missing from the block signature. Imo the signature should attest to all fields of the block header and e.g. Polygon's validators do sign everything afaic: 
https://github.com/maticnetwork/bor/blob/master/consensus/bor/bor.go#L1034
https://github.com/maticnetwork/bor/blob/master/core/types/block.go#L74

JamesHinshelwood:
Agreed. We should centralise the computation of message hashes and ensure all fields are included. I'll edit this issue's title and description to state that.

PR #982, opened 2024-05-27
Add `gasUsed` to block headers

### LINKED ISSUE #978: Return `gasUsed` in `eth_getBlockByNumber` and `eth_getBlockByHash`
Currently `gasUsed` is set to 0

JamesHinshelwood:
I chose to add the value to block headers, rather than calculate it on the fly. Feel free to discuss if you think the denormalisation is a bad idea.

I also removed a (IMO) misguided `Display` impl for blocks.

Resolves #978.

### DIFF
--- a/z2/src/converter.rs
+++ b/z2/src/converter.rs
@@ -219,6 +219,7 @@ pub async fn convert_persistence(
                 state.root_hash()?,
                 txn_hashes.iter().map(|h| Hash(h.0)).collect(),
                 SystemTime::UNIX_EPOCH + Duration::from_micros(block.timestamp),
+                ScillaGas(block.gas_used).into(),
             );
 
             let mut block_receipts = Vec::new();
--- a/zilliqa/src/api/types/eth.rs
+++ b/zilliqa/src/api/types/eth.rs
@@ -105,7 +105,7 @@ impl Header {
             total_difficulty: 0,
             extra_data: vec![],
             gas_limit: BLOCK_GAS_LIMIT,
-            gas_used: EvmGas(0),
+            gas_used: header.gas_used,
             timestamp: header
                 .timestamp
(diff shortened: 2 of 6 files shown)


DrZoltanFazekas:
> I chose to add the value to block headers, rather than calculate it on the fly. Feel free to discuss if you think the denormalisation is a bad idea.

The `gasUsed` belongs into the block header according to the Ethereum execution layer spec: https://github.com/ethereum/execution-specs/blob/master/src/ethereum/shanghai/blocks.py#L58

What do you mean by denormalisation?

JamesHinshelwood:
> What do you mean by denormalisation?

`gasUsed` can be deterministically derived from other other fields in the block (i.e. the parent's state root and the transactions in the block). I meant it in the sense of https://en.wikipedia.org/wiki/Denormalization.

DrZoltanFazekas:
I don't have an issue with the denormalisation as long as the validators check the correctness of the `gasUsed` stored in the block headers they receive from the block proposer.

DrZoltanFazekas:
What feels unnatural though is that we're breaking what I believed was an intentional separation between the consensus-layer's `BlockHeader` and the execution-layer's `Header` struct, but maybe they were never meant to be clearly separated:
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/message.rs#L388
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/api/types/eth.rs#L55

DrZoltanFazekas:
But as you spotted in https://github.com/Zilliqa/zq2/issues/981, the timestamp and more is missing from the block signatures, so we might need to add further fields to the `Header` anyway

rrw-zilliqa (approved):
Looks good to me :-)

PR #984, opened 2024-05-28
Move some of the consts into config file

bzawisto:


zilliqa/src/api/types/eth.rs
```diff
@@ -88,7 +87,7 @@ pub struct Header {
 }
 
 impl Header {
-    pub fn from_header(header: message::BlockHeader, miner: Address) -> Self {
+    pub fn from_header(header: message::BlockHeader, miner: Address, block_gas_limit: u64) -> Self {
```
JamesHinshelwood, line 90: I do think its important to keep this as an `EvmGas` for as far as we can. It would be easy to confuse a ScillaGas parameter here.

zilliqa/src/cfg.rs
```diff
@@ -99,7 +102,7 @@ impl Default for ConsensusConfig {
             rewards_per_hour: 32_000_000_000_000u128,
             blocks_per_hour: 50_000,
             minimum_stake: 32_000_000_000_000_000_000u128,
-            block_gas_limit: 84_000_000,
+            eth_block_gas_limit: EvmGas(84_000_000),
             gas_price: 4_761_904_800_000,
```
JamesHinshelwood, line 106: Small nit - I think we should remove these defaults or remove the values from `infra/config.toml` - I don't like having the 'sensible' values in two places where they can diverge. My light preference would be that we remove these defaults and force people to have to specify some values in `config.toml`?

JamesHinshelwood (approved):

zilliqa/src/cfg.rs
```diff
@@ -88,6 +99,11 @@ impl Default for ConsensusConfig {
             minimum_time_left_for_empty_block: Duration::from_millis(3000),
             scilla_address: "http://localhost:3000".to_owned(),
             local_address: "localhost".to_owned(),
+            rewards_per_hour: 32_000_000_000_000u128,
+            blocks_per_hour: 50_000,
+            minimum_stake: 32_000_000_000_000_000_000u128,
```
DrZoltanFazekas, line 104: Is there a point in not using the same values as defined in `config.toml`? With an `empty_block_timeout` of 1000ms we would never end up with 50,000 `blocks_per_hour` anyway :-)

DrZoltanFazekas (changes requested):

zilliqa/src/cfg.rs
JamesHinshelwood: Maybe it makes more sense to remove the `Default` impl for `Config` and instead just default the fields that do have sensible values with `#[serde(default = "...")]`? I agree with you that having `Option`s here and unwrapping everywhere is quite horrible.

JamesHinshelwood: I think the bigger problem with making them optional and unwrapping everywhere is that the error messages to node operators who forget to fill these fields will be much worse. Instead of "missing field X" at startup, their node will start but then later they will get a random panic and have to search the backtrace to work out why.

How painful would removing the `Default` impl be? If it is painful, then I guess we can merge this for now and change it later? Up to you.

JamesHinshelwood (approved):
LGTM except the one default that I think might be wrong.

zilliqa/src/cfg.rs
JamesHinshelwood: Thanks :)

JamesHinshelwood (approved):

DrZoltanFazekas (approved):

theo-zil (commented):
Only comment is that consensus-modifying parameters should also be configurable in the `src/contracts/shard.sol` Shard contract, so that shards can dynamically modify them. The config file is effectively a way to bootstrap a node from the chicken-and-egg problem (of needing to read the blockchain to get the params, and needing a working node to connect to the blockchain) which may/will be alleviated once light client protocols are working.

Happy for you to dismiss this comment (or ping me to change it to an approval) and update the shard contract in a separate PR if you prefer.

@rrw-zilliqa this should answer your concern as well by the way - we already have the infra for that and it's the intended way to set configs in the future.

Issue #979, opened 2024-05-27
Unify the representation of signers in `QuorumCertificate` and `AggregateQC`


DrZoltanFazekas:
Replace the vector of signer indices in `AggregateQC`:
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/message.rs#L353
with a bit vector like in `QuorumCertificate`:
https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/message.rs#L277
and transform the bit vector to signer indices on the fly when needed

PR #985, opened 2024-05-28
Unified representation of signers in QCs and AggQcs.

bzawisto:
Fixes: https://github.com/Zilliqa/zq2/issues/979

### DIFF
--- a/zilliqa/src/message.rs
+++ b/zilliqa/src/message.rs
@@ -357,15 +357,6 @@ pub struct AggregateQc {
 
 impl AggregateQc {
     pub fn compute_hash(&self) -> Hash {
-        let mut signers = Vec::new();
-        for (index, bit) in self.cosigned.iter().enumerate() {
-            if *bit {
-                let index: u16 = index.try_into().expect("Unable to convert cosigned index!");
-                let bytes = index.to_be_bytes().to_vec();
-                signers.extend_from_slice(&bytes);
-            }
-        }
-
         let hashes: Vec<_> = self.qcs.iter().map(|qc| qc.compute_hash()).collect();
         let signers = self.cosigned.as_raw_slice();
         Hash::compute([


zilliqa/src/consensus.rs
```diff
@@ -37,9 +37,8 @@ use crate::{
 
 #[derive(Debug)]
 struct NewViewVote {
-    signatures: Vec<NodeSignature>,
-    signers: Vec<u16>,
-    cosigned: BitVec,
+    // Keeps signer index in deposit contract and sent signature
+    signatures: Vec<(u16, NodeSignature)>,
```
rrw-zilliqa, line 41: I would put this in a structure for readability. Otherwise, particularly with type inference, everyone is going to lose track of which order the tuple is in fairly quickly and then we'll get confused, and ..

Also, why store this as a vector? Given the ops you're doing on it later on, would a HashMap<> or a BTreeMap<> be faster?

zilliqa/src/consensus.rs
```diff
@@ -1088,26 +1087,22 @@ impl Consensus {
             debug!("ignoring new view from unknown node (buffer?) - committee size is : {:?} hash is: {:?} high hash is: {:?}", committee.len(), new_view.qc.block_hash, self.high_qc.block_hash);
             return Ok(None);
         };
+
+        let index = index as u16;
```
rrw-zilliqa, line 1091: this will wrap if (and I admit this is unlikely!) someone ever has an index > 65535. Suggest try_into() ..

zilliqa/src/consensus.rs
```diff
@@ -1116,10 +1111,15 @@ impl Consensus {
...
-            signers.push(index as u16);
-            cosigned.set(index, true);
+        let has_voted = signatures.iter().any(|(idx, _)| *idx == index);
+
+        if !has_voted {
+            // Keeps (signer, signature) sorted by signer index in deposit contract
+            let pos = signatures
+                .binary_search_by(|(elem, _)| elem.cmp(&index))
+                .unwrap();
+            signatures.insert(pos, (index, new_view.signature));
```
rrw-zilliqa, line 1121: Why keep this sorted if you're going to iterate through it all with any()? (also, I'd be wary of maintaining order like this - it's far too easy for an off-by-one error - introduced by you or later by someone else - to cause the list to be "almost" sorted, and subsequent subtle "the signature was there, but then we thought it wasn't" errors).

zilliqa/src/consensus.rs
```diff
@@ -1308,14 +1305,18 @@ impl Consensus {
...
-        signatures: &[NodeSignature],
-        signers: Vec<u16>,
+        signatures: &[(u16, NodeSignature)],
     ) -> Result<AggregateQc> {
         assert_eq!(qcs.len(), signatures.len());
-        assert_eq!(signatures.len(), signers.len());
+
+        let mut cosigned: BitVec = bitvec![u8, bitvec::order::Msb0; 0; 0];
+        for (idx, _) in signatures.iter() {
+            cosigned.set(*idx as usize, true);
```
rrw-zilliqa, line 1314: this one is actually safe, but into() not as ..

zilliqa/src/consensus.rs
```diff
@@ -1525,7 +1526,14 @@ impl Consensus {
         self.verify_qc_signature(&block.qc, committee.clone())?;
         if let Some(agg) = &block.agg {
             // Check if the signers of the block's aggregate QC represent the supermajority
-            self.check_quorum_in_indices(&agg.signers, &committee)?;
+
+            let mut signers = Vec::new();
+            for (index, bit) in agg.signers.iter().enumerate() {
+                if *bit {
+                    signers.push(index as u16);
```
rrw-zilliqa, line 1533: You're taking one field called signers (different from the Vote field called signers) and turning it into  a variable called signers which contains different values.

Comments, please ! (and preferably let's name things differently - perhaps signer_index_vec<> and signer_bitvec<> ?)

zilliqa/src/message.rs
```diff
@@ -350,18 +350,24 @@ impl Display for QuorumCertificate {
...
-    pub signers: Vec<u16>,
+    pub signers: BitVec,
     pub view: u64,
     pub qcs: Vec<QuorumCertificate>,
 }
 
 impl AggregateQc {
     pub fn compute_hash(&self) -> Hash {
+        let mut signers = Vec::new();
+        for (index, bit) in self.signers.iter().enumerate() {
```
rrw-zilliqa, line 361: Second time we've done this op - perhaps worth a method?

rrw-zilliqa (approved):
I'll say looks good to me, because I think it works, but I'd like to see the "as" es gone at the very least..

zilliqa/src/message.rs
JamesHinshelwood, line 361: Why not just calculate the signature over the bit vector? Why do we need to convert it to an index list?

zilliqa/src/consensus.rs
JamesHinshelwood, line 41: Given we need to convert this to a bitvec later when we construct the aggregate QC and given it will generally be quite a dense list, why not just use a bitvec here too? Especially since we're sorting the index list later.

JamesHinshelwood (changes requested):
I think we should use a bitvec in `NewViewVote` too. But feel free to convince me otherwise if I'm being dumb :)

zilliqa/src/consensus.rs
```diff
@@ -1088,38 +1093,33 @@ impl Consensus {
...
-                cosigned: bitvec![u8, bitvec::order::Msb0; 0; committee_size],
+                signatures: BTreeMap::new(),
                 cosigned_weight: 0,
                 qcs: Vec::new(),
             });
 
         let mut supermajority = false;
 
+        let index: u16 = index.try_into()?;
         // the index is not checked here...
```
theo-zil, line 1118: With `BTreeMap::entry()`, the index _is_ checked now - comment is outdated

theo-zil (commented):
Minor nit but otherwise agree with James's point w.r.t. storing signatures as a bitvec - since we keep converting it to and from a bitvec anyway.
If it turns out to not be sparse enough for a bitvec to be efficient, then we should probably use bitvecs nowhere - and if it is, then why not use them everywhere?

zilliqa/src/consensus.rs
```diff
@@ -37,9 +43,8 @@ use crate::{
 
 #[derive(Debug)]
 struct NewViewVote {
-    signatures: Vec<NodeSignature>,
-    signers: Vec<u16>,
-    cosigned: BitVec,
+    // Represents signer index as kept in deposit contract and received signature
+    signatures: BTreeMap<u16, NodeSignature>,
```
JamesHinshelwood, line 47: I still don't really understand why this is a map and not just a `BitVec` of signers, given that we need to convert it to a bitvec later anyway and that the bitvec is probably more efficient because we expect a high proportion of the committee to be signers.

zilliqa/src/consensus.rs
```diff
@@ -1525,7 +1518,14 @@ impl Consensus {
...
             // Check if the signers of the block's aggregate QC represent the supermajority
-            self.check_quorum_in_indices(&agg.signers, &committee)?;
+
+            let mut signers = Vec::new();
+            for (index, bit) in agg.cosigned.iter().enumerate() {
+                if *bit {
+                    signers.push(index as u16);
+                }
+            }
+            self.check_quorum_in_indices(&signers, &committee)?;
```
JamesHinshelwood, line 1528: Can this method instead take the bitvec? From a glance we should be able to implement it just as efficiently by iterating over the bitvec directly rather than a list of indices? Then we can remove the conversion preceding this call?

PR #988, opened 2024-05-28
Enforce block gas limit and keep track of cumulative gas used

### LINKED ISSUE #967: Enforce gas block limit
We enforce it for individual transactions, but not for all the transactions in the block :man_facepalming:

JamesHinshelwood:
When filling a block with transactions from the mempool, we stop taking transactions once the block gas limit is reached.

Resolves #967

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -736,6 +736,7 @@ impl Consensus {
     }
 
     pub fn get_txns_to_execute(&mut self) -> Vec<VerifiedTransaction> {
+        let mut gas_left = self.config.consensus.eth_block_gas_limit;
         std::iter::from_fn(|| self.transaction_pool.best_transaction())
             .filter(|txn| {
                 let account_nonce = self.state.must_get_account(txn.signer).nonce;
@@ -746,6 +747,14 @@ impl Consensus {
                     .nonce()
                     .map(|tx_nonce| tx_nonce >= account_nonce)
                     .unwrap_or(true)
+            })
+            .take_while(|txn| {
+                if let Some(g) = gas_left.checked_sub(txn.tx.gas_limit()) {
+                    gas_left = g;
+                    true
+                } else {
+                    false
+                }
             })
             .collect()
     }
@@ -971,16 +980,18 @@ impl Consensus {
 
         let transactions = self.get_txns_to_execute();
 
-        let mut gas_used = EvmGas(0);
+        let mut gas_left = self.config.consensus.eth_block_gas_limit;
         let applied_transactions: Vec<_> = transactions
             .into_iter()
             .filter_map(|tx| {
                 self.apply_transaction(tx.clone(), parent_header, inspector::noop())
                     .transpose()
                     .map(|r| {
-                        r.map(|result| {
-                            gas_used += result.gas_used();
-                            tx
+                        r.and_then(|result| {
+                            gas_left = gas_left
+                                .checked_sub(result.gas_used())
+                                .ok_or_else(|| anyhow!("too much gas used"))?;
+                            Ok(tx)
                         })
                     })
             })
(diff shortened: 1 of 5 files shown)


zilliqa/src/consensus.rs
```diff
@@ -750,6 +751,14 @@ impl Consensus {
...
                     .unwrap_or(true)
             })
+            .take_while(|txn| {
+                if let Some(g) = gas_left.checked_sub(txn.tx.gas_limit()) {
+                    gas_left = g;
+                    true
+                } else {
+                    false
+                }
+            })
```
rrw-zilliqa, line 761: I'm assuming (haha) that there is an additional check which stops txns which have tx.gas_limit() > BLOCK_GAS_LIMIT  entering the pool (because they will now never leave it), and that best_transaction() orders by gas descending (otherwise txns can starve - and worse, I can starve them with enough gas - though perhaps you're OK with that?)

zilliqa/src/consensus.rs
```diff
@@ -974,16 +983,18 @@ impl Consensus {
...
                 self.apply_transaction(tx.clone(), parent_header, inspector::noop())
                     .transpose()
                     .map(|r| {
-                        r.map(|result| {
-                            gas_used += result.gas_used();
-                            tx
+                        r.and_then(|result| {
+                            gas_left = gas_left
+                                .checked_sub(result.gas_used())
+                                .ok_or_else(|| anyhow!("too much gas used"))?;
```
rrw-zilliqa, line 996: If this anyhow() fires, does anyone then propose a new block? (in what way does the next block to be proposed differ st the next potential leader won't fire this anyhow! again?)

zilliqa/src/transaction.rs
```diff
@@ -557,6 +567,20 @@ impl FromStr for ScillaGas {
...
+    pub fn checked_sub(self, rhs: EvmGas) -> Option<EvmGas> {
+        Some(EvmGas(self.0.checked_sub(rhs.0)?))
+    }
+}
+
+impl Sub for EvmGas {
+    type Output = EvmGas;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        self.checked_sub(rhs).expect("evm gas underflow")
```
rrw-zilliqa, line 580: What happens if I can contrive for this to panic?

rrw-zilliqa (dismissed):
Not actively horrid, but worth thinking about the comments?

zilliqa/src/consensus.rs
JamesHinshelwood: There isn't, but there is an issue to add it :) #1003  However just to clarify, the current behaviour (in this PR) is that transactions with `gas_limit > BLOCK_GAS_LIMIT` will immediately be flushed from the pool because `checked_sub` will always return false.


Yes, `best_transaction()` always orders by gas descending.

zilliqa/src/consensus.rs
JamesHinshelwood: The intention is that this `anyhow!` never fires, thanks to the checks in `get_txns_to_execute`. If it does fire (because I am mistaken and there is a bug), this node won't propose a block and the other nodes will vote for a new view (after a timeout). Whether the same bug occurs again depends on the bug i guess. Though in theory once this node is the proposer again, the transactions which caused problems will no longer be in its mempool.

zilliqa/src/transaction.rs
JamesHinshelwood: Bad things. We catch panics in the context of API calls, but not when processing messages. I've raised #1004 to fix this.

Once merged, it will mean this node fails to propose a block and the network new views. Hopefully the same doesn't occur again, but again.. It depends on what the bug is :)

rrw-zilliqa (approved):
Looks good to me :-)

PR #990, opened 2024-05-29
Remove test timeout

JamesHinshelwood:
The tests only fail for taking 'too long' once they've actually completed anyway. This means you can have tests that pass, but exit with a failure.

We can enforce a test timeout in CI if we wish, but I don't think doing it directly in the test suite is sensible.


theo-zil (approved):
I'm in agreement

Issue #1008, opened 2024-06-04
Make z2 perf work against eth funding sources.


rrw-zilliqa:
Currently it doesn't, which means that to make z2 perf (eg. `scripts/z2 perf perf/lots.yaml`) work you need a Zilliqa-derived genesis account, which means you need to remember which genesis accounts were ZIL and which were ETH, which is in turn quite confusing.

PR #1009, opened 2024-06-04
Start of code to make perf work on eth addresses

rrw-zilliqa:
Addresses (but does not close) #1008 ; this PR itself doesn't work because it tries to poll eth transactions with GetTransaction , which will claim they don't exist.


JamesHinshelwood:
#1010 is the problem

its-saeed:
Will be merged in #1051

PR #1012, opened 2024-06-04
Implement `GetTransaction` for non-Zilliqa transactions

### LINKED ISSUE #1010: `GetTransaction` doesn't return non-Zilliqa transactions

JamesHinshelwood:
Fixes #1010

### DIFF
--- a/docs/api/zilliqa/gettransaction.doc.md
+++ b/docs/api/zilliqa/gettransaction.doc.md
@@ -23,9 +23,9 @@ Querying for non-existent transactions or transactions that have not yet been mi
 | `amount` | Number as string | The value of the transaction. This amount is always returned in units of Qa (10^-12 ZILs). For Zilliqa transactions, this means we return the exact amount that was passed into `CreateTransaction`. For Ethereum transactions, the true amount is truncated from 18 digits to 12.
 | `signature` | Hex string with `0x` | The transaction signature. Zilliqa signatures are 64 bytes, consisting of `r` followed by `s`. Ethereum signatures are 65 bytes, consisting of `r`, followed by `s`, followed by the `v` value in 'Electrum' notation. Intershard transactions have no signature, so contain an empty string here.
 | `receipt.accepted` | Optional boolean | If the transaction was a Zilliqa transaction and was a call to a Scilla contract, whether the called contract accepted the ZIL sent to it. |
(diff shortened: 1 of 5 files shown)


rrw-zilliqa (approved):
Whilst this LGTM, I fear it may not be the whole answer..  I'll comment in #1010 and we may need to raise another issue (because github)

docs/api/zilliqa/gettransaction.doc.md
```diff
@@ -9,6 +9,40 @@ transaction,get,hash
...
+| `signature` | Hex string with `0x` | The transaction signature. Zilliqa signatures are 64 bytes, consisting of `r` followed by `s`. Ethereum signatures are 65 bytes, consisting of `r`, followed by `s`, followed by the `v` value in 'Electrum' notation. Intershard transactions have no signature, so contain an empty string here.
+| `receipt.accepted` | Optional boolean | If the transaction was a Zilliqa transaction and was a call to a Scilla contract, whether the called contract accepted the ZIL sent to it. |
+| `receipt.cumulativeGas` | Number as string | The gas used by this transactions and all transactions in the same block that preceded it. This amount is always returned in units of Scilla gas. Internally, gas is tracked in units of EVM gas. When the true amount is not an exact multiple of the EVM to Scilla gas exchange rate, this value will be rounded. |
+| `receipt.epochNum` | Number as string | The number of the block in which this transaction was mined. |
+| `receipt.eventLogs` | Optional array | If the transaction was a Zilliqa transaction, the logs from any Scilla contracts that were executed. EVM logs are not included. |
```
rrw-zilliqa, line 28: For eth txns I get event_logs, not eventLogs from ZQ1 - there are probably other differences, though, and I haven't checked what I get back from ZQ1 for Scilla txns :-(

JamesHinshelwood: Ooh quite right. All the fields in the `receipt` object are `snake_case`. Will fix, thanks for spotting.

rrw-zilliqa (approved):
Looks good to me :-) - though note that I've not actually checked that the fields are identical with what happens in ZQ1 - I'm afraid I'm just trusting you so far. One day hopefully I/we will have a test suite which will test this, but probably not worth duplicating this manually just now?

PR #1022, opened 2024-06-05
Txpool content API impl

### LINKED ISSUE #965: Implement `txpool_content`
Blockscout said they would like it.

bzawisto:
This implements txpool_content.

Resolves #965

### DIFF
--- a/docs/api/others/txpool_content.doc.md
+++ b/docs/api/others/txpool_content.doc.md
@@ -1,0 +1,83 @@
+# Title
+
+txpool_content
+
+# Keywords
+
+pool,transaction
+
+# Description
+
+Returns the exact details of all the transactions currently pending for inclusion in the next block(s), as well as the ones that are being scheduled for future execution only.
+
+# Curl
+
+```shell
+    curl -d '{
+        "id": "1",
+        "jsonrpc": "2.0",
+        "method": "txpool_content",
+        "params": [ 1000 ]
+    }' -H "Content-Type: application/json" -X POST "{{ _api_url }}"
+```
+
+# Response
+
+```json
+{
+  pending: {
+    0x0216d5032f356960cd3749c31ab34eeff21b3395: {
+      806: {
+        blockHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
+        blockNumber: null,
+        from: "0x0216d5032f356960cd3749c31ab34eeff21b3395",
+        gas: "0x5208",
+        gasPrice: "0xba43b7400",
+        hash: "0xaf953a2d01f55cfe080c0c94150a60105e8ac3d51153058a1f03dd239dd08586",
+        input: "0x",
+        nonce: "0x326",
+        to: "0x7f69a91a3cf4be60020fb58b893b7cbb65376db8",
+        transactionIndex: null,
+        value: "0x19a99f0cf456000"
+      }
+    },
+    0x24d407e5a0b506e1cb2fae163100b5de01f5193c: {
+      34: {
+        blockHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
+        blockNumber: null,
+        from: "0x24d407e5a0b506e1cb2fae163100b5de01f5193c",
+        gas: "0x44c72",
+        gasPrice: "0x4a817c800",
+        hash: "0xb5b8b853af32226755a65ba0602f7ed0e8be2211516153b75e9ed640a7d359fe",
+        input: "0xb61d27f600000000000000000000000024d407e5a0b506e1cb2fae163100b5de01f5193c00000000000000000000000000000000000000000000000053444835ec580000000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+        nonce: "0x22",
+        to: "0x7320785200f74861b69c49e4ab32399a71b34f1a",
+        transactionIndex: null,
+        value: "0x0"
+      }
+    }
+  },
+  queued: {
+    0x976a3fc5d6f7d259ebfb4cc2ae75115475e9867c: {
+      3: {
+        blockHash: "0x0000000000000000000000000000000000000000000000000000000000000000",
+        blockNumber: null,
+        from: "0x976a3fc5d6f7d259ebfb4cc2ae75115475e9867c",
+        gas: "0x15f90",
+        gasPrice: "0x4a817c800",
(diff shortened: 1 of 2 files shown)


zilliqa/src/pool.rs
```diff
@@ -127,6 +133,60 @@ impl TransactionPool {
...
+            pending.push(txn.clone());
+            pending_set.insert(txn.hash);
+
+            let Some(next) = tx_index.next() else {
+                continue;
+            };
+
+            if let Some(next_txn) = self.transactions.get(&next) {
+                ready.push(next_txn.into());
+            }
```
JamesHinshelwood, line 171: Is it worth de-duplicating this logic with the same logic in `fn best_transaction`? Or is it just superficially similar and actually different? I ask because it took me a second to verify this was maintaining the heap in the same way.

zilliqa/src/pool.rs
```diff
@@ -127,6 +133,60 @@ impl TransactionPool {
...
+        }
+
+        // Find remaining transactions that are scheduled for execution in the future
+        let mut queued: Vec<VerifiedTransaction> = Vec::new();
+
+        for (index, txn) in self.transactions.iter() {
+            if let TxIndex::Intershard(_, _) = index {
```
JamesHinshelwood, line 178: See above - I think we should include these.

zilliqa/src/node.rs
```diff
@@ -589,6 +590,10 @@ impl Node {
         self.consensus.get_transaction_by_hash(hash)
     }
 
+    pub fn txpool_content(&self) -> TxPoolContent {
+        self.consensus.txpool_content()
```
JamesHinshelwood, line 594: FWIW I'd consider just making `transaction_pool` in `Consensus` `pub` instead and accessing it directly here. We already do similar for other fields and it saves an extra level of functions.

JamesHinshelwood: Or maybe not given you have to filter out transactions with low nonces anyway.

JamesHinshelwood (approved):
LGTM - A few comments you might want to think about.

zilliqa/src/pool.rs
```diff
@@ -127,6 +133,60 @@ impl TransactionPool {
         }
     }
 
+    pub fn preview_content(&self) -> TxPoolContent {
+        // First make a copy of 'ready' transactions
+        let mut ready: BinaryHeap<ReadyItem> = BinaryHeap::new();
+        for item in self.ready.iter() {
+            ready.push(*item);
+        }
```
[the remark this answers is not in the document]
JamesHinshelwood: I think it does, as long as the `T` implements `Clone`:
https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#impl-Clone-for-BinaryHeap%3CT,+A%3E

So maybe it didn't before you added the `Clone` impl for `ReadyItem` but now it does?

zilliqa/src/pool.rs
```diff
@@ -127,6 +133,60 @@ impl TransactionPool {
...
+
+        let mut pending = Vec::new();
+        let mut pending_set = HashSet::new();
+
+        // Find all transactions that are pending for inclusion in the next block
+        while let Some(ReadyItem { tx_index, .. }) = ready.pop() {
+            // Should we care about nonceless txns?
```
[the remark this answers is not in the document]
JamesHinshelwood: Oh yeah, that's annoying. Fair enough. I guess we should exclude them in that case.

Then can we at least update the two comments to explain that yes we are deliberately excluding them because they don't have a nonce?

zilliqa/src/api/erigon.rs
```diff
@@ -1,13 +1,22 @@
...
-use crate::{message::BlockNumber, node::Node};
+use crate::{api::types::eth::Transaction, message::BlockNumber, node::Node};
 
 pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
-    super::declare_module!(node, [("erigon_getHeaderByNumber", get_header_by_number)])
+    super::declare_module!(
+        node,
+        [
+            ("erigon_getHeaderByNumber", get_header_by_number),
+            ("txpool_content", txpool_content)
```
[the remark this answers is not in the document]
JamesHinshelwood: Yeah fair point - Maybe just `other.rs` in `api/` with a short comment at the top explaining that this collects together the implementations of a few namespaces that only have 1/a few methods?

zilliqa/src/pool.rs
```diff
@@ -127,6 +133,60 @@ impl TransactionPool {
...
+                continue;
+            }
+
+            // Expired txn?
+            let Some(txn) = self.transactions.get(&tx_index) else {
+                continue;
+            };
+
+            if pending_set.contains(&txn.hash) {
+                continue;
```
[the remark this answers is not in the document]
JamesHinshelwood: I'm still slightly confused, sorry. Do you mean if there was a stale transaction in the heap?

JamesHinshelwood (dismissed):
LGTM, but I'm still confused about one comment. Feel free to merge if I'm being dumb :)

zilliqa/src/api/others.rs
```diff
@@ -0,0 +1,36 @@
...
+use jsonrpsee::{types::Params, RpcModule};
+
+use super::types::eth;
+use crate::{api::types::eth::Transaction, node::Node};
+
+pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
+    super::declare_module!(node, [("txpool_content", txpool_content)])
+}
+
+fn txpool_content(_params: Params, node: &Arc<Mutex<Node>>) -> Result<Option<eth::TxPoolContent>> {
```
rrw-zilliqa, line 17: Might now be a good time to document this API call? 'cos otherwise someone (read: me) is going to have to reverse-engineer what you intended it to do, figure out whether it does it, and then document it ..

JamesHinshelwood: Very good point, thanks for the reminder.

rrw-zilliqa: thanks! Yep - that'll do .. there are a few formatting things I'll fix up when I get there, that you couldn't know about until the PR I'm about to submit! Thank you!

zilliqa/src/pool.rs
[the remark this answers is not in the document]
JamesHinshelwood: Thanks!

JamesHinshelwood (approved):

PR #1023, opened 2024-06-05
Merge configuration files if more than one is passed

JamesHinshelwood:
After some discussion with Mauro, we realised that node configuration can be split broadly into two categories:

* Stuff that never changes and should have the same values across all nodes in a network - e.g. configuration for the genesis block.
* Stuff that is specific to each node - e.g. port configuration, enabled APIs.

We decided that a simple way to handle this split was to allow multiple configuration files which get merged by the application.

We deliberately ban configuration keys from occurring in more than one file. Therefore, no inheritance or overriding occurs when merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,13 @@ Run:
 docker-compose up
 ```
 
+## Node configuration
+
+Nodes are configured by TOML files.
+Pass the path to configuration files on the command line with `-c` or `--config-file`.
+If multiple configuration files are provided, they will be merged together.
+If a configuration key occurs in more than one configuration file, the process will exit with an error.
+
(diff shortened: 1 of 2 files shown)


rrw-zilliqa (approved):
Looks good to me :-)

If you explicitly want to support the model of a network and a node configuration file, there is no shame in having a network.toml and a node.toml - and this prevents node ops and others from getting the location of a key in the wrong file. No strong feelings, but it's a way...

JamesHinshelwood:
Thanks, lets try this for now since it solves @mauromedda 's immediate problem and adapt later if we think this way is bad.

PR #1026, opened 2024-06-06
Flatten Scilla contract storage on disk

### LINKED ISSUE #919: Split Scilla contract storage into separate keyspace
Currently we store a Scilla contract's storage as follows:
```rust
storage: BTreeMap<String, (ScillaValue, String)>, // <var name> -> (<value>, <type>)

enum ScillaValue {
    Bytes(Vec<u8>),
    Map(BTreeMap<String, ScillaValue>),
}
```

This means whenever we load any part of a Scilla contract, we must load its full state. Some contracts have very large state, so this is a bad idea.

Also our Scilla state diffs (#916) will include _all_ state of a changed contract, rather than just the changed state.

Instead, we should do something similar to EVM contract where the account trie stores `(nonce, balance, code, storage_root)` and perform further lookups for storage slots using the storage root. We'll probably have to flatten Scilla map keys to fit them into a trie, but that shouldn't be difficult. We should probably also move the `Scilla` and `Evm` variants down to just the `code` field too.

JamesHinshelwood:
Previously, we would serialise the full `BTreeMap` representing a Scilla contract's storage to disk and deserialise the same thing to memory whenever we needed it. Now, we flatten the map (concatenating the map indices) and store the values in a contract-specific trie. This is closer to how EVM contract storage works and means we can partially load/write a Scilla contract's storage. Also, it means the state deltas generated by our execution engine only need to store what was actually changed.

`Account` changes to always have a `storage_root` (since both EVM and Scilla contracts now store their state in a trie). `code` is now split into an enum which differentiates between EVM and Scilla contracts. Note that Scilla contracts also store their `init_data` and type information here.

I'm almost certainly missing some explanatory comments in places, so please complain if you find things hard to follow. I've been working on this for a while so I have some complexity blindness.

I made one other drive-by change to support arbitrary values in Scilla log parameters that are returned by the interpreter. Some of the example contracts on `ide.zilliqa.com` use these, so it was useful for testing.

Resolves #919

### DIFF
--- a/zilliqa/src/exec.rs
+++ b/zilliqa/src/exec.rs
@@ -11,7 +11,8 @@ use std::{
 
 use alloy_primitives::{Address, U256};
 use anyhow::{anyhow, Result};
-use eth_trie::Trie;
+use bytes::Bytes;
+use eth_trie::{EthTrie, Trie};
 use ethabi::Token;
 use libp2p::PeerId;
 use revm::{
@@ -30,12 +31,13 @@ use tracing::{debug, info, trace, warn};
 use crate::{
     contracts,
     crypto::{Hash, NodePublicKey},
+    db::TrieStorage,
     eth_helpers::extract_revert_msg,
     inspector::{self, ScillaInspector},
     message::{Block, BlockHeader},
     precompiles::get_custom_precompiles,
-    scilla::{self, Scilla},
-    state::{contract_addr, Account, Contract, ScillaValue, State},
+    scilla::{self, split_storage_key, storage_key, Scilla},
+    state::{contract_addr, Account, Code, State},
     time::SystemTime,
     transaction::{
         total_scilla_gas_price, EvmGas, EvmLog, Log, ScillaGas, ScillaLog, ScillaParam,
@@ -43,7 +45,7 @@ use crate::{
     },
 };
 
-type ScillaResultAndState = (ScillaResult, HashMap<Address, Account>);
+type ScillaResultAndState = (ScillaResult, HashMap<Address, PendingAccount>);
 
 /// Data returned after applying a [Transaction] to [State].
 pub enum TransactionApplyResult {
@@ -242,7 +244,7 @@ impl DatabaseRef for &State {
             nonce: account.nonce,
             code_hash: KECCAK_EMPTY,
             code: Some(Bytecode::new_raw(
-                account.contract.evm_code().unwrap_or_default().into(),
+                account.code.evm_code().unwrap_or_default().into(),
             )),
         };
 
@@ -408,11 +410,11 @@ impl State {
     ) -> Result<ScillaResultAndState> {
         let mut state = PendingState::new(self.try_clone()?);
 
-        let code = self
-            .get_account(txn.to_addr)?
-            .contract
-            .scilla_code()
-            .unwrap_or_default();
+        let has_code = !txn.to_addr.is_zero()
+            && matches!(
+                state.load_account(txn.to_addr)?.account.code,
+                Code::Scilla { .. }
+            );
 
         let deposit = total_scilla_gas_price(txn.gas_limit, txn.gas_price);
         if let Some(result) = state.deduct_from_account(from_addr, deposit)? {
@@ -431,7 +433,7 @@ impl State {
                 current_block,
                 inspector,
             )
-        } else if code.is_empty() {
+        } else if !has_code {
             scilla_transfer_to_eoa(state, from_addr, txn, inspector)
         } else {
             scilla_call(state, self.scilla(), from_addr, txn, inspector)
@@ -440,8 +442,8 @@ impl State {
         let from = state.load_account(from_addr)?;
         let refund =
             total_scilla_gas_price(gas_limit - ScillaGas::from(result.gas_used), gas_price);
-        from.balance += refund.get();
-        from.nonce += 1;
+        from.account.balance += refund.get();
+        from.account.nonce += 1;
 
         Ok((result, state.finalize()))
     }
@@ -490,9 +492,57 @@ impl State {
     }
 
     /// Applies a state delta from a Scilla execution to the state.
-    fn apply_delta_scilla(&mut self, state: &HashMap<Address, Account>) -> Result<()> {
+    fn apply_delta_scilla(&mut self, state: &HashMap<Address, PendingAccount>) -> Result<()> {
         for (&address, account) in state {
-            self.save_account(address, account.clone())?;
+            let mut storage = self.get_account_trie(address)?;
+
+            /// Recursively called internal function which assigns `value` at the correct key to `storage`.
+            fn handle(
+                storage: &mut EthTrie<TrieStorage>,
+                var: &str,
+                value: &StorageValue,
+                indices: &mut Vec<Vec<u8>>,
+            ) -> Result<()> {
+                match value {
+                    StorageValue::Map { map, complete } => {
+                        // If this is a complete view of the map, delete any existing values first.
+                        if *complete {
+                            storage.remove_by_prefix(&storage_key(var, indices))?;
+                        }
+
+                        // Add the next level of index before making the recursive call.
+                        indices.push(vec![]);
+                        for (k, v) in map {
+                            indices.last_mut().unwrap().clone_from(k);
+                            handle(storage, var, v, indices)?;
+                        }
+                    }
+                    StorageValue::Value(Some(value)) => {
+                        let key = storage_key(var, indices);
+                        storage.insert(&key, value)?;
+                    }
+                    StorageValue::Value(None) => {
+                        let key = storage_key(var, indices);
+                        // A deletion may occur at any depth of the map. Therefore we remove all keys with the relevent
+                        // prefix.
+                        storage.remove_by_prefix(&key)?;
+                    }
+                }
+                Ok(())
+            }
+
+            for (var, value) in &account.storage {
+                handle(&mut storage, var, value, &mut vec![])?;
+            }
+
+            let account = Account {
+                nonce: account.account.nonce,
+                balance: account.account.balance,
+                code: account.account.code.clone(),
+                storage_root: storage.root_hash()?,
+            };
+
+            self.save_account(address, account)?;
         }
 
         Ok(())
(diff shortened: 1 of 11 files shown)


zilliqa/src/exec.rs
```diff
@@ -826,14 +890,183 @@ impl PendingState {
...
+            .entry(var_name.to_owned())
+            .or_insert_with(StorageValue::incomplete);
+        for key in indices {
+            let current_map = match current {
+                StorageValue::Map { map, .. } => map,
+                StorageValue::Value(Some(_)) => {
+                    return Err(anyhow!("expected a map"));
+                }
+                StorageValue::Value(None) => {
+                    unimplemented!("deletes of whole maps are unsupported")
```
rrw-zilliqa, line 930: It's hard to tell  but I think ZQ1 supports this case -  needs an issue or a fix?
At this point you will also need a StorageValue::Deleted ...

zilliqa/src/exec.rs
```diff
@@ -826,14 +890,183 @@ impl PendingState {
...
+                }
+            };
+            let child = current_map
+                .entry(key.clone())
+                .or_insert_with(StorageValue::incomplete);
```
rrw-zilliqa, line 935: Perhaps call this StorageValue::incomplete_map() ?

zilliqa/src/exec.rs
```diff
@@ -826,14 +890,183 @@ impl PendingState {
...
+                StorageValue::incomplete()
+            });
+            for key in indices {
+                let current_map = match current {
+                    StorageValue::Map { map, .. } => map,
+                    StorageValue::Value(Some(_)) => {
+                        return Err(anyhow!("expected a map"));
+                    }
+                    StorageValue::Value(None) => {
+                        unimplemented!("deletes of whole maps are unsupported")
```
rrw-zilliqa, line 969: Whilst they may be, that's not what this unimplemented! should be saying :-p

zilliqa/src/exec.rs
```diff
@@ -826,14 +890,183 @@ impl PendingState {
...
+                }
+
+                let map = match current_value {
+                    StorageValue::Map { map, .. } => map,
+                    StorageValue::Value(_) => {
+                        return Err(anyhow!("expected map"));
+                    }
+                };
+                // Note that we insert 'complete' maps here, because we know we are going to add *ALL* values from
+                // within this map.
```
rrw-zilliqa, line 1051: map -> prefix ?

zilliqa/src/exec.rs
```diff
@@ -826,14 +890,183 @@ impl PendingState {
...
+                current_value = map.entry(index).or_insert_with(StorageValue::complete);
             }
+
+            *current_value = if let Some(cached) = current_cached {
+                cached.clone()
+            } else {
+                StorageValue::Value(Some(v.into()))
+            };
         }
+
```
rrw-zilliqa, line 1061: If I:

 - Create foo.bar..baz = {"wombats": 2 }
 - Set foo.bar = "fish" (deleting the map)
 - Ask for foo.bar.baz

I think what'll happen here is that we'll walk the tree, note the value at foo.bar, ignore it and return { "wombats": 2}?

rrw-zilliqa (dismissed):
Looks good to me :-) - but check comments in case you think otherwise..

zilliqa/src/exec.rs
```diff
@@ -408,11 +410,11 @@ impl State {
     ) -> Result<ScillaResultAndState> {
         let mut state = PendingState::new(self.try_clone()?);
 
-        let code = self
-            .get_account(txn.to_addr)?
-            .contract
-            .scilla_code()
-            .unwrap_or_default();
+        let has_code = !txn.to_addr.is_zero()
```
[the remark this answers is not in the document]
JamesHinshelwood: Good question - The default account (which would be returned by `state.load_account(Address::ZERO)`) has `code = Code::Evm(vec![])` which makes this check unnecessary. I must have left it in from a previous version of my code. Thanks for spotting.

zilliqa/src/exec.rs
```diff
@@ -490,9 +492,57 @@ impl State {
...
+                        // If this is a complete view of the map, delete any existing values first.
+                        if *complete {
+                            storage.remove_by_prefix(&storage_key(var, indices))?;
+                        }
+
+                        // Add the next level of index before making the recursive call.
+                        indices.push(vec![]);
+                        for (k, v) in map {
+                            indices.last_mut().unwrap().clone_from(k);
+                            handle(storage, var, v, indices)?;
```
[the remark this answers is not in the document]
JamesHinshelwood: That is basically what I'm doing - I'll explain in a comment :)

zilliqa/src/exec.rs
```diff
@@ -490,9 +492,57 @@ impl State {
...
+                    }
+                    StorageValue::Value(Some(value)) => {
+                        let key = storage_key(var, indices);
+                        storage.insert(&key, value)?;
+                    }
+                    StorageValue::Value(None) => {
+                        let key = storage_key(var, indices);
+                        // A deletion may occur at any depth of the map. Therefore we remove all keys with the relevent
+                        // prefix.
+                        storage.remove_by_prefix(&key)?;
```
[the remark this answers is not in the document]
JamesHinshelwood: The Scilla interpreter has no way of setting `foo[bar] = None`, other than `delete foo[bar]`. Just to be clear `StorageValue::Value(None)` always means a deletion - The values are not actually optional. Of course, you can use `Option`s at the Scilla language level, but both `Some` and `None` on Scilla's side will turn up as `Some(value)` here.

The only reason I have `StorageValue::Value(Option<Bytes>)` rather than `StorageValue::Value(Bytes)` and `StorageValue::Deleted` is so that `load_storage` can return `&mut Option<Bytes>` and the caller can set it to `None` to delete a value.

I will add a comment in `StorageValue` to clarify the usage. I hope my explanation made sense?

zilliqa/src/exec.rs
JamesHinshelwood: Just for context - Both the `unimplemented!`s are related to the assertion here (which you also commented on):
https://github.com/Zilliqa/zq2/blob/e660f240c5becf704a4af548589649878c3e7253/zilliqa/src/scilla.rs#L738

In the sense that, if that assertion holds, these branches are unreachable. I realise now that I was meant to use `unreachable!` instead.

In terms of whether we do actually need to support map-level deletions, I did some digging and couldn't find Scilla syntax for doing so, nor could I find any contracts which fail my assertion. I figured if we did find such a contract later (e.g. when replaying ZQ1 mainnet), the assertion would explode violently and we could do the necessary work to fix it?

zilliqa/src/exec.rs
JamesHinshelwood: See above.

zilliqa/src/scilla.rs
```diff
@@ -689,67 +726,48 @@ impl ActiveCall {
...
-            }
+        if indices.len() > depth {
+            return Err(anyhow!("too many indices"));
+        }
 
-            let Some(mut map) = storage_slot.as_map_mut() else {
-                return Err(anyhow!("not a map"));
+        if ignore_value {
+            // We only supporting deleting a single value of a map.
+            assert_eq!(indices.len(), depth);
```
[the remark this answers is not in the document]
JamesHinshelwood: See above.

zilliqa/src/exec.rs
JamesHinshelwood: Thanks, that does make more sense.

zilliqa/src/exec.rs
JamesHinshelwood: I'm 99% sure your second bullet point is impossible. Scilla won't let you assign a value above the bottom level of a map. In other words, the type of `foo` is `Map String (Map String Int)`. When you do `foo[bar] = "fish"`, the Scilla type checker will say "expected a `(Map String Int)` but got the string `"fish"`"?

JamesHinshelwood:
> Looks good to me :-) - but check comments in case you think otherwise..

Thanks :) I've resolved the things I've agreed with and updated. I left replies to your other comments where I think additional discussion is needed. Let me know if I haven't made sense!

zilliqa/src/exec.rs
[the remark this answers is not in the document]
rrw-zilliqa: Yep. Absolutely - and agree that a comment would have been useful to me. Thanks!

zilliqa/src/exec.rs
```diff
@@ -517,19 +567,16 @@ impl State {
...
                         .as_ref()
-                        .map(|c| c.original_bytes().to_vec())
-                        .unwrap_or_default(),
-                    storage_root: if storage.iter().count() != 0 {
-                        Some(storage.root_hash()?)
-                    } else {
-                        None
-                    },
-                },
+                        .expect("code_by_hash is not used")
```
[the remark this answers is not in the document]
rrw-zilliqa: Absolutely, but "code_by_hash is not used" is a bit of a cryptic way of saying it. Might "code expected non-empty, but found to be None" be better?

zilliqa/src/exec.rs
rrw-zilliqa: Agree.  Perhaps worth a:

```
// StorageValue::Value(None) means delete, and since we don't support deleting whole elements here .. 
```
 ?

zilliqa/src/exec.rs
rrw-zilliqa: Scilla may not, but your interface allows it and one day some bright spark (possibly me :-))  may decide to  let the Scilla interpreter allow it somehow, on which day you probably want a better way of reacting than generating an obscure bug that occasionally returns previously deleted maps - or you could prevent the situation from arising by rejecting updates to a non-terminal key at your level as well..

rrw-zilliqa (approved):
re-approving for the sake of keeping things mergeable should you want to do so.

zilliqa/src/exec.rs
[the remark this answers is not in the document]
JamesHinshelwood: Good point, thank you.

zilliqa/src/exec.rs
JamesHinshelwood: Will add an assertion in a follow-up :+1:

PR #1041, opened 2024-06-13
Debug trace block by number

bzawisto:
Adding new tracing call (geth specific). I had to extend the variant returned from EVM because the `evm environment` is needed for constructing traces. I thought it would be the least intrusive way - otherwise the env would have to be passed from caller which would become rather painful (every time you want to run EVM - pass extra env configuration).

### DIFF
--- a/docs/api/trace/debug_traceblockbynumber.doc.md
+++ b/docs/api/trace/debug_traceblockbynumber.doc.md
@@ -51,3 +51,4 @@ Accepts a block number and tracer type and will replay the block that is already
 | `params`  | array  | Required | `[block_number, tracerConfig ]` |
 
 For more details on tracerConfig see https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#traceconfig
+Currently supported tracers are: opcode logger (if no tracer is give), calltracer, 4bytetracer, muxtracer, prestatetracer and jstracer.


docs/api/trace/debug_traceblockbynumber.doc.md
```diff
@@ -0,0 +1,53 @@
...
+# Arguments
+
+| Parameter | Type   | Required | Description                     |
+|-----------|--------|----------|---------------------------------|
+| `id`      | string | Required | `"1"`                           |
+| `jsonrpc` | string | Required | `"2.0"`                         |
+| `method`  | string | Required | `"debug_traceBlockByNumber"`    |
+| `params`  | array  | Required | `[block_number, tracerConfig ]` |
+
+For more details on tracerConfig see https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#traceconfig
```
rrw-zilliqa, line 53: Unless we guarantee to support everything geth has ever supported (ie. any retrieval of this URL will yield a spec we must comply with), which I suggest would be rash, we should indicate what we actually do comply with - I've been putting this in the description.

zilliqa/src/api/trace.rs
```diff
@@ -4,14 +4,23 @@ use std::{
...
-use crate::{crypto::Hash, node::Node};
+use crate::{crypto::Hash, message::BlockNumber, node::Node};
 
 pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
-    super::declare_module!(node, [("trace_replayTransaction", replay_transaction)])
+    super::declare_module!(
+        node,
+        [
+            ("trace_replayTransaction", replay_transaction),
+            ("debug_traceBlockByNumber", debug_trace_block_by_number),
```
rrw-zilliqa, line 21: It'd be neater to have two modules - one for trace_ and one for debug_, but probably only if they are both going to acquire lots of API calls. If there are only a few, I guess this will have to do (the problem is that it means I'll have to introduce extra complexity in the doc generator to reformat API calls by prefix with a bunch of heuristics - because ZIL API calls have no prefix .. )

rrw-zilliqa (approved):
Looks good to me, modulo a couple of comments which if you want to merge this, I can fix externally ..

rrw-zilliqa (approved):
Looks good to me :-) (though when I get there I might expand the docs a bit if no-one minds)

PR #1044, opened 2024-06-14
Improve block download speed when syncing with a network

### LINKED ISSUE #968: Speed up syncing
Fiddle with batch sizes to see if syncing from genesis can be made faster.

JamesHinshelwood:
There are quite a few changes here but the motivation for all of them is to allow nodes to request multiple blocks in parallel.

The previous behaviour was that our `Consensus` was (seemingly) randomly scattered with calls to `download_blocks_up_to_head`. This would send out a `BlockRequest` for the current height plus one. Nodes would respond to these requests with the requested block, plus up to 100 additional blocks in the same batch.

All of this code has been removed and the logic is mostly handled in `BlockStore`. We now inform `BlockStore` when we see blocks that don't form a chain with our existing blocks. It buffers these proposals in case they can form a chain later
and keeps track of the `highest_known_view` it has seen. When it is poked to do so (after a new block proposal is received), it sends out requests for as many blocks as possible, respecting its configuration for the batch size and maximum
number of in-flight blocks. Importantly, these requests are now sent and handled in parallel.

Block requests and responses have been simplified to a single type for each. The non-batch versions were unused anyway.

Closes #968


zilliqa/src/cfg.rs
```diff
@@ -124,8 +142,6 @@ pub struct ConsensusConfig {
     /// (public key, peerId, amount, reward address) tuples.
     #[serde(default)]
     pub genesis_deposits: Vec<(NodePublicKey, PeerId, Amount, Address)>,
-    #[serde(default)]
-    pub genesis_hash: Option<Hash>,
```
JamesHinshelwood, line 128: cc @theo-zil for opinion

z2/src/setup.rs
```diff
@@ -218,6 +219,9 @@ impl Setup {
                     blocks_per_hour: 3600,
                     rewards_per_hour: 51_000_000_000_000_000_000_000u128.into(),
                 },
+                block_request_limit: block_request_limit_default(),
+                max_blocks_in_flight: max_blocks_in_flight_default(),
+                block_request_batch_size: block_request_batch_size_default(),
```
rrw-zilliqa, line 224: Could we have a brief docs/ file on the algorithm you intended to use here, please? I can kind of work it out by looking at the patch, but the patch is in several files - a quick one-liner that just points to the right comments might work out well?

zilliqa/src/block_store.rs
```diff
@@ -18,18 +19,109 @@ use crate::{
 pub struct BlockStore {
     db: Arc<Db>,
     block_cache: RefCell<LruCache<Hash, Block>>,
+    /// The maximum view of blocks we've seen.
+    highest_known_view: u64,
```
rrw-zilliqa, line 23: define "seen" ? Received? Know to exist some other way?

zilliqa/src/block_store.rs
```diff
@@ -18,18 +19,109 @@ use crate::{
...
+    /// The maximum view of blocks we've sent a request for.
+    requested_view: u64,
+    /// The maximum number of blocks to send requests for at a time.
+    max_blocks_in_flight: u64,
+    /// The maximum number of blocks to request at a time.
+    batch_size: u64,
+    /// Buffered block proposals, indexed by their parent hash.
+    //buffered: HashMap<Hash, Proposal>,
```
rrw-zilliqa, line 31: Maybe delete the commented line?

zilliqa/src/block_store.rs
```diff
@@ -18,18 +19,109 @@ use crate::{
...
+    buffered: LruCache<Hash, Proposal>,
```
rrw-zilliqa, line 32: Why do we buffer block proposals?

zilliqa/src/block_store.rs
```diff
@@ -18,18 +19,109 @@ use crate::{
...
+        proposal: Proposal,
+        random_peer: Option<PeerId>,
+    ) -> Result<()> {
+        let view = proposal.view();
+
+        // We want to be careful about buffering too many proposals. There is no guarantee or proof that any of them
+        // will eventually form our canonical chain. Therefore we limit the number of buffered proposals to
+        // `max_blocks_in_flight + 100`. This number is chosen because it guarantees we will always be able to handle
+        // pending requests made by ourselves that arrive out-of-order, while also giving some extra space for newly
+        // created blocks that arrive while we are syncing.
```
rrw-zilliqa, line 64: Given that we can have an arbitrary number of proposals from other nodes, why does '+100' guarantee that we will always be able to handle pending requests made by us?

zilliqa/src/block_store.rs
```diff
@@ -18,18 +19,109 @@ use crate::{
...
+        let current_view = self
+            .get_block_by_number(
+                self.db
+                    .get_highest_block_number()?
+                    .ok_or_else(|| anyhow!("no highest block"))?,
+            )?
+            .ok_or_else(|| anyhow!("missing highest block"))?
+            .view();
+        // We can consider all blocks we currently have committed to already be requested.
+        if self.requested_view < current_view {
```
rrw-zilliqa, line 89: Really? Why? Presumably because we've already got them, so perhaps clearer to say
```
// We've already got any block we've previously committed, so don't request it again
```
or have I misunderstood?

zilliqa/src/block_store.rs
```diff
@@ -18,18 +19,109 @@ use crate::{
...
+            batch_size: config.block_request_batch_size,
+            buffered: LruCache::new(
+                NonZeroUsize::new(config.max_blocks_in_flight as usize + 100).unwrap(),
+            ),
             message_sender,
         })
     }
 
+    /// Buffer a block proposal whose parent we don't yet know about.
+    pub fn buffer_proposal(
```
rrw-zilliqa, line 53: Ah! _that_'s why we buffer proposals!

zilliqa/src/consensus.rs
```diff
@@ -515,18 +500,26 @@ impl Consensus {
...
+                            block,
+                            transactions
+                                .into_iter()
+                                .map(|tx| {
+                                    let hash = tx.calculate_hash();
+                                    (tx, hash)
+                                })
+                                .collect(),
+                        ),
+                        random_peer,
```
rrw-zilliqa, line 518: Note that if the proposal buffer contains more than its max configured size, we will in fact not buffer the proposal - is that OK (or at least, given that the buffer is LRU, we will throw a proposal away, which may have been the result of a previous call from here). Happy with that?

zilliqa/src/consensus.rs
```diff
@@ -2146,6 +2126,10 @@ impl Consensus {
             ));
         }
 
+        // Tell the block store to request more blocks if it can.
+        let random_peer = self.get_random_other_peer();
+        self.block_store.request_missing_blocks(random_peer)?;
+
```
rrw-zilliqa, line 2132: Might be worth flagging this in an issue/in the docs - if you end up choosing your peer badly (and some folks run peers that refuse to respond to the request_missing_blocks() request), the network could presumably slow down quite significantly?

Also, reading this, it looks as though we will request_missing_blocks() on every block execution, which means that we might not slow the network down, but if anyone is slow about sending us blocks, we will issue multiple overlapping requests for a large number of blocks, which will presumably become a bandwidth issue - particularly if lots of nodes join at the same time?

zilliqa/src/message.rs
```diff
@@ -231,6 +229,35 @@ impl ExternalMessage {
...
+                    (Some(first), None) => write!(f, "BlockResponse([{first}])"),
+                    (Some(first), Some(last)) => {
+                        write!(f, "BlockResponse([{first}, ..., {last}])")
+                    }
+                    (None, Some(_)) => unreachable!(),
+                }
+            }
+            ExternalMessage::NewTransaction(_) => write!(f, "NewTransaction"),
+            ExternalMessage::RequestResponse => write!(f, "RequestResponse"),
+        }
```
rrw-zilliqa, line 257: yay! :-)

rrw-zilliqa (approved):
Looks good to me - sorry there are so many comments!

z2/src/setup.rs
JamesHinshelwood: Will add a doc-comment to `BlockStore` if you don't disagree - I think all of the relevant code is here and that way we keep the algorithm description close to the implementation?

zilliqa/src/block_store.rs
JamesHinshelwood: It doesn't and "guarantee" is a silly word, thanks for spotting :)

zilliqa/src/block_store.rs
JamesHinshelwood: That's much better, thank you :)

zilliqa/src/consensus.rs
JamesHinshelwood: I will be happy with that once we have retries...
https://github.com/Zilliqa/zq2/issues/1096

I've added a comment to point to this.

zilliqa/src/consensus.rs
JamesHinshelwood: Good questions.

The random peer thing seems problematic to me for a few reasons:
* We don't retry requests for blocks, so if enough nodes doesn't respond to us we will get stuck - #1096 
* We only consider peers who are in the consensus committee which seems overly restrictive - #1097

And I agree that bandwidth could be an issue if many nodes are syncing at once, but I'm not sure how to solve it... My best suggestions would just be making sure the code path for "send blocks in response to a block request" is fast and doesn't contend with other operations, plus rate limiting/scoring for peers to lessen the impact of malicious requests?

JamesHinshelwood:
@rrw-zilliqa Thanks for the detailed review - I hope I've addressed most/all of your concerns with extra comments :) Let me know if not - I've left the conversations unresolved so you can decide.

z2/src/setup.rs
rrw-zilliqa: Sounds good to me - thanks!

zilliqa/src/block_store.rs
rrw-zilliqa: :-)

zilliqa/src/consensus.rs
rrw-zilliqa: :-) - sounds good to me; not hitting the resolve button because I can't see the comment (yet), but feel free to click the resolve button if you are happy.

zilliqa/src/consensus.rs
rrw-zilliqa: It is a hard thing to do - I'd:

 * Don't request blocks directly. Ask a strategy function what to do and have it request blocks on your behalf.
 * Write the strategy function so that it requests block ranges from "the best" current peer.
 * The strategy function maintains a set of ranges which we don't have, but want. If you can, separate these into "blocking - need these now" and "just want these sometime".
 *  Strategy function maintains a window and a latency figure for each peer
 * When a block comes in, update the bandwidth and latency for the peer - maybe start with AIMD + fast start.
 * On each tick (fsvo "tick"), fill the windows for random peers .
 * If you _really_ feel bad about it, introduce a NAK. When you have too many requests pending, send some NAKs to close your window to those who are being greedy.
 * That's it.

This is basically TCP, but for blocks. It has the tidy advantage that you incidentally get a map you can display of with whom you are communicating and how well you think they are doing. It has the disadvantage that I just thought of it and it might be utter nonsense :-p

rrw-zilliqa (approved):
Looks good to me :-)

zilliqa/src/consensus.rs
JamesHinshelwood: Will add this to #1097 and continue discussing there, I agree broadly :) Thanks!

PR #1047, opened 2024-06-17
Attach consensus information in Block responses

bzawisto:
(Aggregate)QuorumCertificate are now part of block response.

### DIFF
--- a/zilliqa/src/api/types/eth.rs
+++ b/zilliqa/src/api/types/eth.rs
@@ -10,8 +10,10 @@ use sha3::{Digest, Keccak256};
 
 use super::{bool_as_int, hex, option_hex, vec_hex};
 use crate::{
+    api::types::ser_display,
     crypto::Hash,
     message,
+    message::BitVec,
     time::SystemTime,
     transaction::{self, EvmGas, EvmLog},
 };
@@ -28,7 +30,8 @@ pub enum HashOrTransaction {
 pub struct QuorumCertificate {
     #[serde(serialize_with = "hex")]
     pub signature: Vec<u8>,
-    pub cosigned: String,
+    #[serde(serialize_with = "ser_display")]
+    pub cosigned: BitVec,
     #[serde(serialize_with = "hex")]
     pub view: u64,
     #[serde(serialize_with = "hex")]
@@ -39,14 +42,7 @@ impl QuorumCertificate {
     pub fn from_qc(qc: &message::QuorumCertificate) -> Self {
         Self {
             signature: qc.signature.to_bytes(),
-            cosigned: qc
-                .cosigned
-                .clone()
-                .iter()
-                .fold(String::new(), |mut acc, bit| {
-                    acc.push(if *bit { '1' } else { '0' });
-                    acc
-                }),
+            cosigned: qc.cosigned.clone(),
             view: qc.view,
             block_hash: qc.block_hash.into(),
         }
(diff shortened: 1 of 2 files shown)


zilliqa/src/api/types/eth.rs
```diff
@@ -24,6 +24,63 @@ pub enum HashOrTransaction {
...
+    pub view: u64,
+    #[serde(serialize_with = "hex")]
+    pub block_hash: B256,
+}
+
+impl QuorumCertificate {
+    pub fn from_qc(qc: &message::QuorumCertificate) -> Self {
+        Self {
+            signature: qc.signature.to_bytes(),
+            cosigned: qc
```
rrw-zilliqa, line 42: This isn't all that relevant, but I'll just leave here that not listing the committee in this structure makes me slightly nervous. You can get it easily enough from `eth_call:getStakers()` with block = block_hash, but doing these side-calls when you're writing bridging code has been gnarly for me in the past, and therefore this data structure gives me an undefined sense of the heebie-jeebies.

Tagging @JamesHinshelwood in case he can spot anything more concrete.

zilliqa/src/api/types/eth.rs
```diff
@@ -24,6 +24,63 @@ pub enum HashOrTransaction {
...
+}
+
+#[derive(Clone, Serialize)]
+pub struct AggregateQc {
+    #[serde(serialize_with = "hex")]
+    pub signature: Vec<u8>,
+    pub cosigned: String,
+    #[serde(serialize_with = "hex")]
+    pub view: u64,
+    pub qcs: Vec<QuorumCertificate>,
```
rrw-zilliqa, line 63: Just checking here that the AggregateQc is based on the block hash in the Header it's included in? (otherwise you don't know at which block to `eth_call()` to get your committee) - mention it just because you included the block_hash in the Qc in a way that suggested that the Qc wasn't necessarily ..

zilliqa/src/api/types/eth.rs
```diff
@@ -34,6 +91,9 @@ pub struct Block {
     pub size: u64,
     pub transactions: Vec<HashOrTransaction>,
     pub uncles: Vec<B256>,
+    pub qc: QuorumCertificate,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub agg_qc: Option<AggregateQc>,
```
rrw-zilliqa, line 96: Placeholder for a slack debate on whether we should be adding extra fields to eth_ APIs ..

rrw-zilliqa (approved):
Looks good to me :-) - if my docs had made it in before this PR, I would've said document the extra fields, but since they didn't :-)

DrZoltanFazekas (approved):

zilliqa/src/api/types/eth.rs
```diff
@@ -24,6 +24,63 @@ pub enum HashOrTransaction {
...
+
+impl QuorumCertificate {
+    pub fn from_qc(qc: &message::QuorumCertificate) -> Self {
+        Self {
+            signature: qc.signature.to_bytes(),
+            cosigned: qc
+                .cosigned
+                .clone()
+                .iter()
+                .fold(String::new(), |mut acc, bit| {
```
JamesHinshelwood, line 46: I think this implementation basically [already exists](https://docs.rs/bitvec/latest/src/bitvec/vec/traits.rs.html#351-365) in the `Display` impl of `BitVec`. Can we instead do
```
#[serde(serialize_with = "ser_display")]
pub cosigned: BitVec,

...

fn ser_display<T, S>(value: &T, serializer: S) -> Result<S::Ok, S::Error>
where
    T: Display,
    S: Serializer
{
    serializer.collect_str(value)
}
```
for both fields?

zilliqa/src/api/types/eth.rs
```diff
@@ -34,6 +91,9 @@ pub struct Block {
     pub size: u64,
     pub transactions: Vec<HashOrTransaction>,
     pub uncles: Vec<B256>,
+    pub qc: QuorumCertificate,
```
JamesHinshelwood, line 94: Bit of a nit - Can we use `quorum_certificate` (and `aggregate_quorum_certificate`) instead of `qc` everywhere for the response fields? I don't think we should expect clients to know what `qc` means.

JamesHinshelwood (approved):

zilliqa/src/api/types/eth.rs
JamesHinshelwood: IMO representing an array of bits as a JSON array is more sensible than a string. :man_shrugging:

zilliqa/src/api/types/eth.rs
JamesHinshelwood: Nothing stops us from adding new fields later so lets resolve this for now, see how good/bad the client UX is and add more if needed.

PR #1062, opened 2024-06-19
Remove unused `genesis_hash` from configuration

JamesHinshelwood:
The intention was that a node could just configure a genesis hash, rather than the full genesis block. The node would download the genesis block from the network.

However, this isn't used anywhere and adds complexity, so we remove it. Specifying the genesis block isn't difficult.


rrw-zilliqa (approved):
Looks good to me :-)

PR #1065, opened 2024-06-19
Add benchmark for block processing

JamesHinshelwood:
Once #1044 we should be saturating our CPU with blocks when syncing, so in theory the bottleneck will be block processing.

This adds a benchmark to measure the speed of block processing.

We can think about running this benchmark and systematically recording the processing speed in CI later.

Relates to #968


rrw-zilliqa (approved):
Looks good to me :-)

Issue #989, opened 2024-05-28
Ensure we are validating all fields of a block


JamesHinshelwood:
Some validation is missing when checking blocks. Off the top of my head we should check:
* That the gas price in transactions is at least the minimum gas price.
* That the block gas limit is respected (maybe this is already validated when executing the transactions, we should check)
* That the `gasUsed` in the block is truthful

Review all fields of a block to work out what we need to validate :)

DrZoltanFazekas:
Here are a few more:

- the timestamp is plausible
- the transactions root reflects the transactions in the block
- the receipts root reflects the receipts generated after executing the transactions
- the state root reflects the state after executing the transactions
- the size of extra data is limited

DrZoltanFazekas:
I left out the `logsBloom` we will also need to check once we implement light clients, and the `prevRandao` field once we implement it.

Issue #1003, opened 2024-06-03
Validate transactions before entering them into the mempool


JamesHinshelwood:
Copy/use code from the validation checks in revm. Do something similar for Scilla transactions.

e.g. Is the gas limit less than the block gas limit and is the nonce greater than the account's nonce.

JamesHinshelwood:
Fixed by #1128

PR #1066, opened 2024-06-19
Hardening block verification

bzawisto:
This PR fixes: https://github.com/Zilliqa/zq2/issues/989

There are more checks during block validation: gas_limit, timestamp, etc - as described in the issue. Additionally I added tx/receipt root hashes to be part of the block header.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -1,7 +1,7 @@
 use std::{collections::BTreeMap, error::Error, fmt::Display, sync::Arc, time::Duration};
 
 use alloy_primitives::{Address, U256};
-use anyhow::{anyhow, bail, Context as _, Result};
+use anyhow::{anyhow, Context as _, Result};
 use bitvec::bitvec;
 use eth_trie::{MemoryDB, Trie};
 use libp2p::PeerId;
@@ -517,12 +517,13 @@ impl Consensus {
         if block.gas_limit() > self.config.consensus.eth_block_gas_limit
             || block.gas_used() > block.gas_limit()
         {
-            bail!(
+            warn!(
                 "Block gas used/limit check failed. Used: {}, Limit: {}, config limit: {}",
                 block.gas_used(),
                 block.gas_limit(),
                 self.config.consensus.eth_block_gas_limit
             );
+            return Ok(None);
         }
 
         match self.check_block(&block, during_sync) {
@@ -1259,8 +1260,6 @@ impl Consensus {
         if self.db.contains_transaction(&txn.hash)? {
             return Ok(false);
         }
-
-        txn.tx.validate_gas_limit()?;
 
         let account_nonce = self.state.get_account(txn.signer)?.nonce;
         let txn_hash = txn.hash;
@@ -2129,7 +2128,8 @@ impl Consensus {
             cumulative_gas_used += gas_used;
 
             if cumulative_gas_used > block.gas_limit() {
-                bail!("Cumulative gas used by executing transactions exceeded block limit!");
+                warn!("Cumulative gas used by executing transactions exceeded block limit!");
+                return Ok(());
             }
 
             let receipt = Self::create_txn_receipt(result, tx_hash, tx_index, cumulative_gas_used);
@@ -2148,25 +2148,26 @@ impl Consensus {
         }
 
         if cumulative_gas_used != block.gas_used() {
-            bail!("Cumulative gas used by executing all transactions: {cumulative_gas_used} is different that the one provided in the block: {}", block.gas_used());
+            warn!("Cumulative gas used by executing all transactions: {cumulative_gas_used} is different that the one provided in the block: {}", block.gas_used());
+            return Ok(());
         }
 
         let receipts_root_hash: Hash = receipts_trie.root_hash()?.into();
         if block.header.receipts_root_hash != receipts_root_hash {
-            bail!(
+            warn!(
                 "Receipt root mismatch. Specified in block: {} vs computed: {}",
-                block.header.receipts_root_hash,
-                receipts_root_hash
+                block.header.receipts_root_hash, receipts_root_hash
             );
+            return Ok(());
         }
 
         let transactions_root_hash: Hash = transactions_trie.root_hash()?.into();
         if block.header.transactions_root_hash != transactions_root_hash {
-            bail!(
+            warn!(
                 "Transactions root mismatch. Specified in block: {} vs computed: {}",
-                block.header.transactions_root_hash,
-                transactions_root_hash
+                block.header.transactions_root_hash, transactions_root_hash
             );
+            return Ok(());
         }
 
         for (receipt, tx_index) in &mut block_receipts {
--- a/zilliqa/src/transaction.rs
+++ b/zilliqa/src/transaction.rs
@@ -8,7 +8,7 @@ use std::{
 use alloy_consensus::{SignableTransaction, TxEip1559, TxEip2930, TxLegacy};
 use alloy_primitives::{keccak256, Address, Signature, B256, U256};
 use alloy_rlp::{Encodable, Header, EMPTY_STRING_CODE};
-use anyhow::{anyhow, bail, Result};
+use anyhow::{anyhow, Result};
 use bytes::{BufMut, BytesMut};
 use itertools::Itertools;
 use k256::elliptic_curve::sec1::ToEncodedPoint;
@@ -24,7 +24,6 @@ use sha3::{
 };
 
 use crate::{
-    constants::{MINIMUM_ETH_GAS, MINIMUM_ZIL_GAS},
     crypto,
     crypto::Hash,
     exec::{ScillaError, ScillaException, ScillaTransition},
@@ -319,34 +318,6 @@ impl SignedTransaction {
                 crypto::Hash(Keccak256::digest(buffer).into())
             }
         }
-    }
-
-    pub fn validate_gas_limit(&self) -> Result<()> {
-        // Verify minimum gas requirement
-        match self {
-            SignedTransaction::Legacy { .. }
-            | SignedTransaction::Eip2930 { .. }
-            | SignedTransaction::Eip1559 { .. } => {
-                if self.gas_limit() < MINIMUM_ETH_GAS {
-                    bail!(
-                        "Provided evm transaction gas limit is too low: {}",
-                        self.gas_limit()
-                    );
-                }
-            }
-            SignedTransaction::Zilliqa { ref tx, .. } => {
-                // Minimum gas needed for scilla run - contract invocation
-
-                if tx.gas_limit < MINIMUM_ZIL_GAS {
-                    bail!(
-                        "Provided scilla transaction gas limit is too low: {}",
-                        tx.gas_limit
-                    );
-                }
-            }
-            _ => {}
-        }
-        Ok(())
     }
 }
 
--- a/zilliqa/src/constants.rs
+++ b/zilliqa/src/constants.rs
@@ -1,5 +1,0 @@
-use crate::transaction::{EvmGas, ScillaGas};
-
-pub const MINIMUM_ETH_GAS: EvmGas = EvmGas(21_000);
-
-pub const MINIMUM_ZIL_GAS: ScillaGas = ScillaGas(10);
--- a/zilliqa/src/lib.rs
+++ b/zilliqa/src/lib.rs
@@ -3,7 +3,6 @@ mod block_store;
 mod blockhooks;
 pub mod cfg;
 pub mod consensus;
-mod constants;
 pub mod contracts;
 pub mod crypto;
 pub mod db;


z2/src/converter.rs
```diff
@@ -355,11 +344,45 @@ pub async fn convert_persistence(
                     }
                 };
 
+                transactions_trie
```
rrw-zilliqa, line 347: Maybe insert a comment describing the process here - you look to be building your hashes in order so that you break the circular hashing dependency by assuming all hashes are 0, then performing the circular hash ops in a defined order. Leave a comment saying that's what you're doing and which order you think you are doing it in (or I can't check if you are!)

zilliqa/src/cfg.rs
```diff
@@ -135,6 +135,9 @@ pub struct ConsensusConfig {
     /// Minimum remaining time allowing to wait for empty block proposal
     #[serde(default = "empty_block_timeout_default")]
     pub minimum_time_left_for_empty_block: Duration,
+    /// Maximum allowed difference between block creation time and receive time
+    #[serde(default = "maximum_delay_for_received_block_default")]
```
rrw-zilliqa, line 139: I'm not sure there is any way to set this parameter, and so I'm not sure it should exist .. could we decide on either one or the other? It's very much like the timeout parameters we have in ZQ1, which constantly cause spurious view changes ..

Also, I think we should document why we care about validating this and what "creation" and "receive" time mean (and how they are measured!) - eg. it is quite possible for validators to have clocks > 30s out of sync with each other, and if I am joining the network I will definitely receive blocks created more than 1.5s ago! (at least, I would hope the genesis block was created more than 1s ago) - presumably these are excluded from the limit, but how?

zilliqa/src/consensus.rs
```diff
@@ -526,6 +527,28 @@ impl Consensus {
...
 
+        let now = SystemTime::now();
+        if !during_sync
+            && block.timestamp() + self.config.consensus.maximum_delay_for_received_block < now
+        {
+            let diff = now.duration_since(SystemTime::UNIX_EPOCH)?
+                - (block.timestamp().duration_since(SystemTime::UNIX_EPOCH)?
+                    + self.config.consensus.maximum_delay_for_received_block);
+            warn!(
+                "Rejecting block - It's too old, diff is {}ms",
```
rrw-zilliqa, line 538: Why do we care about this? If old blocks trog in slowly over a period of hours, why do we care?

rrw-zilliqa: Also, why do we have two settings for this? check_block() checks `config.allowed_timestamp_skew` .

zilliqa/src/consensus.rs
```diff
@@ -526,6 +527,28 @@ impl Consensus {
...
+                diff.as_millis()
+            );
+            return Ok(None);
+        }
+
+        if block.gas_used() > self.config.consensus.eth_block_gas_limit {
```
rrw-zilliqa, line 544: I don't like this test at all, I'm afraid.
 - It means that bar some odd situations (syncing from a preprepared archive), you can never reduce the block gas limit, or you will choke syncing old blocks which had gas_used() higher than your current block_gas_limit.
 - We won't be processing this block anyway if it didn't get a 2/3 stake vote. What is the point of having a number which, if it is below the number everyone else has, simply causes you to halt your node? You might as well simply read eth_block_gas_limit from the blocks being voted on by the committee.
 - If you ever pick a number which is higher than other folks' eth_block_gas_limit, you will produce proposals that others will reject, causing potentially spurious view changes and network problems.

If we mean to set a block gas limit for the network, we should put it in a contract, and simply never vote for blocks that have a block gas limit higher than the one we set. The only case in which this test makes sense, AFAICT, is that we avoid checking the votes for blocks with gas_used() > self.config.consensus.eth_block_gas_limit, and that will only save us if there is an attacker who is dumb enough to lie about everything else, but not block.gas_used()...

zilliqa/src/transaction.rs
```diff
@@ -299,6 +301,59 @@ impl SignedTransaction {
...
         }
     }
+
+    pub fn validate_gas_limit(&self) -> Result<()> {
+        // Verify minimum gas requirement
+        match self {
+            SignedTransaction::Legacy { .. }
+            | SignedTransaction::Eip2930 { .. }
+            | SignedTransaction::Eip1559 { .. } => {
+                const MINIMUM_ETH_GAS: EvmGas = EvmGas(21_000);
```
rrw-zilliqa, line 311: these constants (in fact, I would say, all constants) could do with being moved to a constants.rs, against the day when we trip over them later..

zilliqa/src/transaction.rs
```diff
@@ -299,6 +301,59 @@ impl SignedTransaction {
...
+                }
+            }
+            _ => {}
+        }
+        Ok(())
+    }
+
+    pub fn validate_input_size(&self) -> Result<()> {
+        // https://eips.ethereum.org/EIPS/eip-170
+        const MAX_EVM_CONTRACT_SIZE_BYTES: usize = 24576;
```
rrw-zilliqa, line 336: And definitely this one!

zilliqa/src/transaction.rs
```diff
@@ -299,6 +301,59 @@ impl SignedTransaction {
...
+        let status = match self {
+            SignedTransaction::Legacy { tx, .. } => tx.input.len() <= MAX_EVM_CONTRACT_SIZE_BYTES,
+
+            SignedTransaction::Eip2930 { tx, .. } => tx.input.len() <= MAX_EVM_CONTRACT_SIZE_BYTES,
+
+            SignedTransaction::Eip1559 { tx, .. } => tx.input.len() <= MAX_EVM_CONTRACT_SIZE_BYTES,
+
+            SignedTransaction::Zilliqa { ref tx, .. } => {
+                // Seems there was no limit in ZQ1 (let's have one)
+                const MAX_ZILLIQA_DATA_SIZE: usize = 32 * 1024;
```
rrw-zilliqa, line 346: Let's be *VERY* certain that legacy txns aren't going to hit it first. And definitely migrate to constants.rs or to the config file .. @JamesHinshelwood  for a decision.

zilliqa/src/transaction.rs
```diff
@@ -708,6 +763,25 @@ impl Log {
             _ => None,
         }
     }
+
+    pub fn hash(&self) -> Hash {
+        match self {
+            Log::Scilla(log) => Hash::compute([
+                log.event_name.as_bytes(),
```
rrw-zilliqa, line 770: I don't think this hash is as much of a hash as you think - Hash::compute() does not (I think?) insert boundaries between its preimages, so a log "foo" with param "bar" hashes to the same as a log with "foo<hash_of_bar>"

zilliqa/src/transaction.rs
```diff
@@ -736,6 +820,40 @@ pub struct TransactionReceipt {
     pub exceptions: Vec<ScillaException>,
 }
 
+impl TransactionReceipt {
+    pub fn hash(&self) -> Hash {
+        let success = [u8::from(self.success); 1];
+        let accepted = [u8::from(self.accepted.unwrap_or_default()); 1];
+        Hash::compute([
```
rrw-zilliqa, line 827: What's the spec for this? Raided from ZQ1?

rrw-zilliqa (changes requested):
Sorry - there are a bunch of comments on this, plus if you close #989 with it, you will lose the data that we also need to check the size of extraData , logsBloom, randaoReveal and prevRandao - feel free as far as I'm concerned to open a separate issue for those, but we shouldn't forget the request ever happened.

I'd also like to see references to specs for the way we compute transactions and receipts (and scilla!) hashes - we might have to invent our own hashes , but we should at least tell folks what they are so that we can spot bugs in our implementation and so that people other than us can check them, given that these are going to end up (presumably?) exposed via the eth_ API .

zilliqa/src/consensus.rs
DrZoltanFazekas: The block header should include both `gas_used` and `gas_limit` (the latter is missing), and we should reject the proposal if `block.gas_used > block.gas_limit` and also if `block.gas_limit > self.config.consensus.eth_block_gas_limit`.

@rrw-zilliqa the `propose()` function is used by validators receiving new block proposals, it's not executed when nodes are syncing historical blocks, and validators shall check if the proposed block is valid according to the current network settings before they vote for it

zilliqa/src/consensus.rs
DrZoltanFazekas: I'd suggest for now we only check the timestamp is increasing as per the [yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf). It's an imperfect solutions, but we can improve it in a separate issue later on...

zilliqa/src/consensus.rs
```diff
@@ -1233,6 +1283,9 @@ impl Consensus {
             return Ok(false);
         }
 
+        txn.tx.validate_gas_limit()?;
+        txn.tx.validate_input_size()?;
```
JamesHinshelwood, line 1287: Can we add this in a different PR? #1003 already exists and I think txn validation can rely on existing code from REVM. Also as @rrw-zilliqa points out there are implications for changing what we accept for Scilla that we should dig into.

I think transaction validation is a separate work item to block verification?

zilliqa/src/transaction.rs
rrw-zilliqa: Calling @JamesHinshelwood  ? Do you think this is the right number?

zilliqa/src/transaction.rs
rrw-zilliqa: @JamesHinshelwood @DrZoltanFazekas  ? Want to make a decision and we can put it in the docs? If you give a transaction a hash, someday someone is going to want to verify it ..

z2/src/converter.rs
```diff
@@ -248,8 +237,8 @@ pub async fn convert_persistence(
                         });
 
                         let receipt = TransactionReceipt {
-                            block_hash: block.hash(),
                             tx_hash: Hash(txn_hash.0),
+                            block_hash: Hash::ZERO,
```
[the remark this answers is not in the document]
rrw-zilliqa: This is what I meant about circular hashing and the fact that we should have a comment to say in which order we are doing it at the start of my review comments, FWIW :-)

z2/src/converter.rs
rrw-zilliqa: How about I file a followup issue and we can merge this PR and move on with our lives and I'll come back later and write a comment?

zilliqa/src/transaction.rs
JamesHinshelwood: As mentioned in some other comment I made, I think we should remove this.

z2/src/converter.rs
rrw-zilliqa: Cool :-) - https://github.com/Zilliqa/zq2/issues/1088

zilliqa/src/transaction.rs
rrw-zilliqa: https://github.com/Zilliqa/zq2/issues/1089

rrw-zilliqa (commented):
Enough has now been fixed that once everyone else's issues have been addressed, looks good to me. I'm not sure if a `Comment` review will dismiss my `request changes`, so let's find out ..
(edit: Sadly, no)

z2/src/converter.rs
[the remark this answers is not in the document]
JamesHinshelwood: I think it will be cleaner in future to add a constructor to `TransactionReceipt` which takes the non-block hash fields, calculates the block hash then populates the struct. I think this kind of pre-initialisation is a bit of a smell. But I can accept if we do that later since this PR has been around for a while.

zilliqa/src/consensus.rs
```diff
@@ -513,6 +514,17 @@ impl Consensus {
...
 
+        if block.gas_limit() > self.config.consensus.eth_block_gas_limit
+            || block.gas_used() > block.gas_limit()
+        {
+            bail!(
+                "Block gas used/limit check failed. Used: {}, Limit: {}, config limit: {}",
+                block.gas_used(),
+                block.gas_limit(),
+                self.config.consensus.eth_block_gas_limit
+            );
```
JamesHinshelwood, line 525: I think you want to log and return `Ok(None)` here - We haven't malfunctioned - Just someone has sent us an invalid block.

JamesHinshelwood: And if you agree, the same point for the rest of these `bail!`s.

zilliqa/src/transaction.rs
```diff
@@ -317,6 +320,34 @@ impl SignedTransaction {
...
+                    bail!(
+                        "Provided evm transaction gas limit is too low: {}",
+                        self.gas_limit()
+                    );
+                }
+            }
+            SignedTransaction::Zilliqa { ref tx, .. } => {
+                // Minimum gas needed for scilla run - contract invocation
+
+                if tx.gas_limit < MINIMUM_ZIL_GAS {
```
JamesHinshelwood, line 340: Didn't we agree to move these changes to a separate PR (fixing #1003) so that we can double check this minimum is correct and doesn't break existing mainnet transactions?

JamesHinshelwood (dismissed):
Sorry for the slow review, will approve now in case you prefer to fix in follow up PR/reminder issues.

rrw-zilliqa (approved):
Approved per discussion in slack - good luck & sorry to have caused delay!

PR #1072, opened 2024-06-20
Add 'blessed' transactions which are exempt from gas price rules

### LINKED ISSUE #1059: Enable deployment of the Arachnid proxy
Make an exemption to the 4761 gwei min gas price in case of the Arachnid transaction hash

JamesHinshelwood:
Currently this just includes the deployment transaction for the deterministic deployment proxy.

This PR also enables the REVM balance check. Previously, this was disabled because Scilla transfers would use the EVM execution flow and they often contained overly high gas limits. This is no longer the case. More alarmingly, it turns out that when the check is disabled, REVM just gives the sender of a transaction enough money to cover that transaction, meaning funds are invented out of thin air. So now mean that the sender of a transaction needs to have enough money to cover the provided `gas_limit * gas_price + amount`.

This means we've had to change how our 'system' transactions are processed. Previously, they would set a very high gas_limit and gas_price and the signer was the zero address. This obviously doesn't work any more since the zero address doesn't have any money at genesis. Instead, we now set the gas_price to zero and disable the gas price checks.

Resolves #1059


zilliqa/src/exec.rs
```diff
@@ -391,11 +393,7 @@ impl State {
             .append_handler_register(inspector_handle_register)
             .modify_cfg_env(|c| {
                 c.chain_id = chain_id;
-                // We disable the balance check (which ensures the from account's balance is greater than the
-                // transaction's `gas_price * gas_limit`), because Scilla transactions are often submitted with
-                // overly high `gas_limit`s. This is probably because there is no gas estimation feature for Scilla
-                // transactions.
-                c.disable_balance_check = true;
+                c.disable_base_fee = skip_base_fee_check;
```
rrw-zilliqa, line 396: What does ZQ1 do? Will check this afternoon if you don't get there first .. I'm a bit nervous about discounting the check altogether, though I do agree with the logic.

zilliqa/src/exec.rs
```diff
@@ -1557,3 +1561,15 @@ fn scilla_call(
...
+
+/// Blessed transactions bypass minimum gas price rules. These transactions have value to the network even at a lower
+/// gas price, so we accept them anyway.
+const BLESSED_TRANSACTIONS: [Hash; 1] = [
+    // Hash of the deployment transaction for the deterministic deployment proxy from
+    // https://github.com/Arachnid/deterministic-deployment-proxy. It is valuable to accept this transaction despite
+    // the low gas price, because it means the contract is deployed at the same address as other EVM-compatible chains.
+    // This means that contracts deployed using this proxy will be deployed to the same address as on other chains.
+    Hash(hex!(
+        "eddf9e61fb9d8f5111840daef55e5fde0041f5702856532cdbb5a02998033d26"
```
rrw-zilliqa, line 1573: https://github.com/Zilliqa/dev-portal/issues/21 is the issue to document this (though probably I now need to start writing a page for all this not-API-but-properties-of-this-version-of-Zilliqa code. Meh.)

zilliqa/src/exec.rs
```diff
@@ -367,6 +368,7 @@ impl State {
         chain_id: u64,
         current_block: BlockHeader,
         inspector: I,
+        skip_base_fee_check: bool,
```
rrw-zilliqa, line 371: I'm not all that happy about this being a boolean, and a negative-sense boolean at that - suggest change to an enum to make it obvious what you're doing so that accidental `true`s here stand out and folks spot them in code review. A failure to do so could be bad .. (not least because it would make code to replay history once you'd fixed it a pretty awful thing to have to write)

zilliqa/src/exec.rs
rrw-zilliqa: Would also be good to know what drives putting this txn in the list? Agree that it's useful, but would want to see passive acceptance on slack (at least) before we start adding too much to this array ...

rrw-zilliqa (approved):
Looks good to me, modulo comments .

zilliqa/src/exec.rs
JamesHinshelwood: For `eth_call`s, ZQ1 sets the `gas_price` to the DS block gas limit. That achieves the same thing as skipping the fee check.

ZQ1 doesn't use REVM, so doesn't have the balance check. If we employed the same approach, all `eth_call`s would fail unless the sender had a balance of `gas_limit * block_gas_price`.

zilliqa/src/exec.rs
JamesHinshelwood: I tried to write the motivation in the comment above

> It is valuable to accept this transaction despite the low gas price, because it means the contract is deployed at the same address as other EVM-compatible chains. This means that contracts deployed using this proxy will be deployed to the same address as on other chains.

Does that make sense or need adding to?

zilliqa/src/exec.rs
JamesHinshelwood: Done :+1:

zilliqa/src/exec.rs
rrw-zilliqa: How about I file an issue, write something in docs/ and then post to slack for visibility? I neither want this to come as a surprise to future developers, nor do I want people to start adding to this list "because it's useful ..". But let's not hold this PR up for that?

rrw-zilliqa (approved):
Looks good to me :-)

zilliqa/src/exec.rs
JamesHinshelwood: Okay with me, thank you :)

rrw-zilliqa:
https://github.com/Zilliqa/zq2/issues/1081 .

PR #1077, opened 2024-06-20
Fix transaction deserialization

### LINKED ISSUE #1076: Attempting to send a transaction via foundry halts the network
To reproduce, start a network with z2, then fire the txns from zilliqa-experimental/richard/hello-forge at it. You'll get a serialisation error and the nodes will panic.

They do this because `p2p_node.rs:299`  fails:
```
 let message = serde_json::from_slice::<ExternalMessage>(&data).unwrap();
```

and this in turn fails because JSON-encoded transactions encode their tx data as an array of bytes (rather than rlp) and the visitor for rlp_encoded txns in transaction.rs:96 doesn't implement visit_seq<>, presumably because you can't easily transmit a byte sequence in JSON.

Two things might help here:

 1. Fix our serde setup so that we send RLP-encoded tx data as (eg.) a string in JSON and continue to use JSON
 2. Abolish JSON/utterly and use CBOR for gossip'd transaction messages.

JamesHinshelwood:
The changes in #1061 fixed (de)serialization for CBOR, but broke it for JSON. This is because `serde_json` serialises bytes as an array by default. At deserialisation, it does not know this array represents bytes and so `visit_seq` is called. Previously, our visitor returned an error here, because it was only expecting `visit_bytes` to be called. Now we accept both.

Fixes #1076

### DIFF
--- a/zilliqa/src/transaction.rs
+++ b/zilliqa/src/transaction.rs
@@ -100,6 +100,21 @@ mod ser_rlp {
             {
                 T::decode(&mut v).map_err(de::Error::custom)
             }
+
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: de::SeqAccess<'de>,
+            {
+                // Limit the length we preallocate.
+                let len = seq.size_hint().unwrap_or(0).min(4096);
+                let mut bytes = Vec::with_capacity(len);
+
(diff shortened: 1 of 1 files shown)


JamesHinshelwood:
Fixes #1076

rrw-zilliqa (approved):
Looks good to me :-)

PR #1079, opened 2024-06-20
Implement EIP-1898

JamesHinshelwood:
We now support a `BlockId` for the following APIs:
* `eth_call`
* `eth_getBalance`
* `eth_getCode`
* `eth_getStorageAt`
* `eth_getTransactionCount`

I couldn't resist making a few related changes at the same time:
* We've deleted our `BlockNumber` and now use alloy's `BlockNumberOrTag` instead.
* `Node` is now much pickier about how blocks and state are accessed. Previously, there were many overlapping ways of getting the same blocks or state. Now, everything goes through `get_block` or `get_state`. A load of unused methods have also been removed thanks to this change.

### DIFF
--- a/evm_scilla_js_tests/test/rpc/eth_getBlockByNumber.ts
+++ b/evm_scilla_js_tests/test/rpc/eth_getBlockByNumber.ts
@@ -129,32 +129,6 @@ describe("Calling " + METHOD, function () {
     });
   });
 
-  it("should get full transactions objects by its block number", async function () {
-    let blockNumber = 0x1;
-
-    await sendJsonRpcRequest(METHOD, 2, [blockNumber, true], (result, status) => {
-      logDebug(result);
-
-      assert.equal(status, 200, "has status code");
-      // validate all returned fields
-      TestResponse(result);
-      assert.equal(+result.result.number, blockNumber, `Block number is not ${blockNumber}`);
-    });
-  });
-
-  it("should get only the hashes of transactions objects by its block number", async function () {
-    let blockNumber = 0x1;
-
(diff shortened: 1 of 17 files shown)


z2/resources/api_macros.tera.md
```diff
@@ -58,6 +58,12 @@ The block number can be:
 
 If the block number is not provided, `latest` is assumed.
 
+# blockid
+
+### Block ID
+
+A block identifier, as specified by https://eips.ethereum.org/EIPS/eip-1898.
```
JamesHinshelwood, line 65: cc @rrw-zilliqa - Do you think its acceptable to just delegate to the EIP here?

rrw-zilliqa: I don't see why not, assuming you aim to support the whole EIP. Mildly harder to look up, but that's about the only factor I can even think of ..

rrw-zilliqa (approved):
Looks good to  me :-)

evm_scilla_js_tests/test/rpc/eth_getBlockByNumber.ts
```diff
@@ -129,32 +129,6 @@ describe("Calling " + METHOD, function () {
     });
   });
 
-  it("should get full transactions objects by its block number", async function () {
```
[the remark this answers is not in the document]
JamesHinshelwood: They tested sending a JSON integer as a block number, which wasn't supported by the Ethereum client I checked (reth), not supported by alloy's `BlockNumberOrTag` and not documented in EIP-1898.

Previously, we did support this but I figured it wasn't necessary or correct, so removed the support and the tests for it.

I double checked our documentation and we correctly state that only strings (containing hex numbers) are allowed.

PR #1092, opened 2024-06-24
Ignore advertised private addresses from other nodes

JamesHinshelwood:
We already ignore loopback addresses, but since we've started deploying in Docker, nodes also now listen on 172.17.0.1 which is in one of the private address ranges (172.16.0.0/12). The behaviour of the Identify protocol is that a node advertises all addresses it could possibly be reached on and leaves it to the client to select the sensible ones. Therefore, we now ignore private addresses too.


rrw-zilliqa (approved):
Looks good to me :-)

PR #1093, opened 2024-06-24
Log at debug level if we receive a proposal we cannot vote on

JamesHinshelwood:
A debug log for this case seems quite reasonable, given it will only output once per block and is quite useful for debugging.


rrw-zilliqa (approved):
Looks good to me :-)

Issue #863, opened 2024-04-24
`BlockNumber` parameters in Eth APIs do not respect two-chain finalisation


JamesHinshelwood:
Currently if you pass "finalized" or "safe" for the block number, we just take the block 2 steps away from the latest. This isn't correct.

https://github.com/Zilliqa/zq2/blob/3b3e3981f67bb4213ca8ac716704cf0bceffbb85/zilliqa/src/node.rs#L288

JamesHinshelwood:
Also consider how to support the `require_canonical` flag from EIP-1898

PR #1102, opened 2024-06-26
Respect two chain finalization

bzawisto:
Caller can now make a query for finalized blocks. This PR also addresses presence of `requires_canonical` flag introduced earlier by https://github.com/Zilliqa/zq2/pull/1079. 

Linked issue: https://github.com/Zilliqa/zq2/issues/863

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -475,7 +475,11 @@ impl Consensus {
             .expect("last timeout seems to be in the future...")
             .as_millis() as u64;
         let view_difference = self.view.get_view().saturating_sub(self.high_qc.view);
-        let exponential_backoff_timeout = consensus_timeout_ms * 2u64.pow(view_difference as u32);
+        // in view N our highQC is the one we obtained in view N-1 (or before) and its view is N-2 (or lower)
+        // in other words, the current view is always at least 2 views ahead of the highQC's view
+        // i.e. to get `consensus_timeout_ms * 2^0` we have to subtract 2 from `view_difference`
+        let exponential_backoff_timeout =
+            consensus_timeout_ms * 2u64.pow((view_difference as u32).saturating_sub(2));
 
         let minimum_time_left_for_empty_block = self
             .config


zilliqa/src/node.rs
```diff
@@ -263,26 +263,65 @@ impl Node {
...
-        match number {
-            BlockNumberOrTag::Latest => self.get_chain_tip(),
-            BlockNumberOrTag::Finalized => self.get_chain_tip().saturating_sub(2),
-            BlockNumberOrTag::Safe => self.get_chain_tip().saturating_sub(2),
+    pub fn resolve_block_number(&self, block_number: BlockNumberOrTag) -> u64 {
+        match block_number {
+            BlockNumberOrTag::Number(n) => n,
             BlockNumberOrTag::Earliest => 0,
+            BlockNumberOrTag::Latest => self.get_chain_tip(),
             BlockNumberOrTag::Pending => self.get_chain_tip(),
```
DrZoltanFazekas, line 271: `latest` and `pending` have different semantics:
- `latest` = the most recent block that has been proposed and received
- `pending` = the block that we would be propose as next, containing transactions from the mempool

zilliqa/src/node.rs
```diff
@@ -263,26 +263,65 @@ impl Node {
...
+                let Ok(Some(block)) = self.db.get_block_by_view(&view) else {
+                    return 0u64;
+                };
+                block.number()
+            }
+            // From whitepaper: If the proposed block’s view number
+            // is one larger than its QC’s view number and
+            // is larger or equal to the validator’s local
+            // view number, the validator regards the
+            // proposed block as safe (which, in the codebase, is referenced as head_block())
```
DrZoltanFazekas, line 285: I'm afraid the terminology used in the HotStuff papers I adhere to in the whitepaper caused a misunderstanding. The term `safe` in HotStuff has nothing to do with the `safe` block tag in the eth APIs.

On post-Merge Ethereum, the `safe` block corresponds to a justified checkpoint: https://github.com/ethereum/consensus-specs/blob/v1.3.0/fork_choice/safe-block.md

The term `justified` comes from Tendermint, and it means it has received 2/3 of the validator votes. Note that the term `checkpoint` here has nothing to do with the weak-subjectivity checkpoints Theo is working on. Ethereum FFG (which is based on the Tendermint algorithm) uses the term `checkpoint` to refer to the first slot of an epoch, which is a candidate for justification (1. voting round) and finalization (2. voting round).

Long story short, the `safe` block tag in our consensus refers to the block that the node's highQC points to (highQC means it's the latest = high, and it's a QC = 2/3 validators voted for it).

rrw-zilliqa (commented):
Note that `z2/resources/api_macros.tera.md` contains references to what zq2 returns for pending and latest.- need to remember to update these ...

zilliqa/src/node.rs
```diff
@@ -263,26 +263,66 @@ impl Node {
...
+    pub fn resolve_block_number(&self, block_number: BlockNumberOrTag) -> u64 {
+        match block_number {
+            BlockNumberOrTag::Number(n) => n,
             BlockNumberOrTag::Earliest => 0,
+            BlockNumberOrTag::Latest => self.get_chain_tip(),
             BlockNumberOrTag::Pending => self.get_chain_tip(),
-            BlockNumberOrTag::Number(n) => n,
+            BlockNumberOrTag::Finalized => {
+                let Ok(Some(view)) = self.db.get_latest_finalized_view() else {
+                    return 0u64;
```
rrw-zilliqa, line 274: Let's not use 0u64 as an error indicator? Sooner or later it will escape from this function and we will get code attempting to reference (or return) block 0x0, not 0x .

zilliqa/src/consensus.rs
```diff
@@ -1028,6 +1037,27 @@ impl Consensus {
         Ok(Some((proposal, broadcasted_transactions)))
     }
 
+    pub fn propose_pending_block(&mut self) -> Result<Block> {
+        let (proposal, applied_transactions) = self
```
rrw-zilliqa, line 1041: I can't see this function used anywhere? What does it do and why is it necessary?

zilliqa/src/consensus.rs
```diff
@@ -1028,6 +1037,27 @@ impl Consensus {
...
+
+        // Check if the votes entry is empty - if so - remove it
+        if let Some((signatures, ..)) = self.votes.get(&proposal.hash()) {
+            if signatures.is_empty() {
+                self.votes.remove(&proposal.hash());
+            }
+        }
+
+        // Re-insert all transactions that were executed during block proposal - they were removed from tx_pool and have to be re-added
+        for tx in applied_transactions {
```
rrw-zilliqa, line 1053: I think the way to do this is to have `propose_block_inner()` take a flag which makes it remember to not remove txns from the pool (or indeed do any other mutations), rather than have the code "just remember what propose_block_inner() does and undo it here" - that means there are two sources of truth - propose_block_inner() and this loop, and inevitably they will drift out of alignment with each other, leading to bugs like "if you call eth_getBalance() with block=pending, and you happen to hit the same node in which there are pending transactions, some of them will be dropped". I'm not sure I want to debug that!

rrw-zilliqa (changes requested):
Couple of comments that I think I'm worried enough about to request changes on, but yell if you feel that I ought to be less brutal.

zilliqa/src/node.rs
JamesHinshelwood: I think you can reasonably and correctly argue that the genesis block (with view 0) is always finalized. In fact I would probably make that argument by changing the signature of `get_latest_finalized_view` to `Result<u64>` and returning 0 if there is nothing in the DB, along with a comment explaining why.

JamesHinshelwood: Oh there are two levels of pattern matching here - I would argue handling mapping `None` to `0` is valid, but `Err` to `0` isn't. Make `resolve_block_number` return `Result<u64>` if its fallible.

zilliqa/src/node.rs
```diff
@@ -263,26 +263,66 @@ impl Node {
...
             BlockId::Hash(RpcBlockHash {
                 block_hash,
-                require_canonical: _,
+                require_canonical,
             }) => {
-                // TODO(#863): require_canonical
-                self.consensus.get_block(&block_hash.into())
+                // See https://eips.ethereum.org/EIPS/eip-1898
+                let Ok(Some(block)) = self.consensus.get_block(&block_hash.into()) else {
+                    const RESOURCE_NOT_FOUND: i32 = -32001;
```
JamesHinshelwood, line 302: I don't think these eth-API level concerns should be happening in `node.rs`. Why not return `None` in this case and map to the error code in the eth API impls?

zilliqa/src/consensus.rs
JamesHinshelwood: Does `preview_content` not let you view the mempool without `&mut self`?

zilliqa/src/node.rs
theo-zil: I may just be being overly pedanting, but I think that `None` makes sense if there are no blocks in the database. Genesis always being finalized sounds correct, but if you don't even know/have genesis, then your "latest finalized block" isn't genesis, it's none.

JamesHinshelwood: Maybe we should document this invariant more clearly, but I'm 99% sure its impossible for `Consensus` to exist with fewer than one block (i.e. genesis always exists), therefore there is always a finalised block.

theo-zil: That's a good point. Then we can remove the optionality from a few database methods. (Probably in another PR, no need to hold up this one)

theo-zil (approved):
Personally I'm not a big fan of making the pool and state into mutexes - the ideal solution would be to have a mechanism to operate on it non-mutably to select transactions into a block, and then "commit" the changes once we've settled on a block to build on. There's no real need for concurrency, so having a mutex just feels like the wrong solution compared to having a robust way to "preview" changes non-mutably (e.g. by making a temporary clone somewhere?).

However I also appreciate that this PR has been here for a while and I also didn't get around to voicing my concerns until much earlier, so I'd also be happy with just moving faster and getting this in and potentially refactoring it in a follow-up if a nicer solution can be found. Plus I'm sure that if there was an easy way to solve it (like "just clone it") you'd have implemented it already, so again I'm in favour of merging the feature now and then considering whether it's worth the time to find a "pure" but complex solution later.

zilliqa/src/cfg.rs
```diff
@@ -176,11 +176,11 @@ pub struct ConsensusConfig {
...
 
 pub fn consensus_timeout_default() -> Duration {
-    Duration::from_secs(5)
+    Duration::from_secs(1)
 }
 
 pub fn empty_block_timeout_default() -> Duration {
-    Duration::from_millis(1000)
+    Duration::from_millis(500)
 }
```
JamesHinshelwood, line 184: Why did we change these defaults?

zilliqa/src/pool.rs
```diff
@@ -159,6 +161,10 @@ impl TransactionPool {
                 continue;
             }
 
+            if block_gas_limit.checked_sub(txn.tx.gas_limit()).is_none() {
```
JamesHinshelwood, line 164:
```suggestion
            if block_gas_limit < txn.tx.gas_limit() {
```

zilliqa/src/state.rs
```diff
@@ -27,7 +27,7 @@ use crate::{
 /// all the keys are hashed and stored in the same sled tree
 pub struct State {
     db: Arc<TrieStorage>,
-    accounts: PatriciaTrie<TrieStorage>,
+    accounts: Arc<Mutex<PatriciaTrie<TrieStorage>>>,
```
JamesHinshelwood, line 30: This feels wrong, given `PatriciaTrie` is basically just `root_hash: B256` + `Arc<TrieStorage>` + some caches for mutations. It is already cheaply cloneable, so why do we need the extra layer?

Also, where is the `Mutex` used? When do we need multiple people to have mutable access to the same storage?

JamesHinshelwood (commented):
Only half a review, but I think I would benefit from some extra explanation about why we need to change some `&mut self` to `&self` everywhere (and thus I guess why we've wrapped stuff in `Mutex`es?)?

PR #1107, opened 2024-07-01
Two state finalization - limited version

bzawisto:
Caller can now make a query for finalized and safe blocks. This PR also addresses presence of requires_canonical flag introduced earlier by https://github.com/Zilliqa/zq2/pull/1079.

Note: This is limited version which doesn't support `Pending` tag.

Linked issue: https://github.com/Zilliqa/zq2/issues/863

### DIFF
--- a/zilliqa/src/api/eth.rs
+++ b/zilliqa/src/api/eth.rs
@@ -425,9 +425,11 @@ fn get_logs(params: Params, node: &Arc<Mutex<Node>>) -> Result<Vec<eth::Log>> {
         (None, from, to) => {
             let from = node
                 .resolve_block_number(from.unwrap_or(BlockNumberOrTag::Latest))?
+                .unwrap()
                 .number();
             let to = node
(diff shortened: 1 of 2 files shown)


zilliqa/src/api/eth.rs
```diff
@@ -93,6 +93,48 @@ pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
...
+    let BlockId::Hash(RpcBlockHash {
+        require_canonical, ..
+    }) = request
+    else {
+        bail!("Unable to find a block with given id!");
+    };
+
+    let require_canonical = require_canonical.unwrap_or_default();
+
+    match require_canonical {
```
JamesHinshelwood, line 116: Doesn't this return the wrong error if I pass `require_canonical: true` and the block isn't found?

JamesHinshelwood (approved):

Issue #1119, opened 2024-07-01
Support pruning


theo-zil:
Nodes can currently run with a partial history (by loading from a checkpoint), but it's not a fully fledged feature.
 * Add API to query what blocks a node has available
 * Make sure RPCs fail gracefully for unknown blocks (might already be the case, just need to check)
 * Add dynamic pruning feature - i.e. a config parameter to keep latest N blocks and delete anything older than N from storage

DrZoltanFazekas:
As an addition to the above, it should also be possible to set the config parameter to a block height we currently don't have (e.g. a block number before the checkpoint from which we synced), to fetch the missing blocks. If the config parameter is set to 0 (genesis), we can sync all blocks older than the checkpoint from which we started syncing.

shawn-zil:
I suspect that the gradual block production throughput degradation in #1970 might have something to do with the growing database indexes. So, if a validator does not hold the entire history, but only a certain number of blocks, the performance degradation might stabilize.

shawn-zil:
Implement this as a post-execution hook i.e. alongside other hooks. If the configuration value is set, ensure that only N number of blocks are present in the DB i.e. `DELETE FROM blocks WHERE height < current - N`.

shawn-zil:
Decision is also to delete transactions, receipts and blocks, when pruning.

shawn-zil:
Decision is to move the RPC call to `admin_blockRange()` instead.

shawn-zil:
@DrZoltanFazekas can I confirm that it is highly improbable for a fork to be > 15 blocks, and so it is safe to set a minimum prune interval of 30-blocks so that we never prune forks?

PR #1120, opened 2024-07-01
State checkpoints

theo-zil:
Adds the ability to export snapshots of the state at a given block, and for nodes to import that state and begin syncing from that height.

De-facto enables running pruned nodes, but doesn't add first-class support for this (see #1119).

**New configs**:
 * `do_snapshots`: enables exporting the snapshot file every checkpoint. *False by default*.
 * `load_checkpoint`: if `Some()`, will load and sync from a snapshot. `load_checkpoint.file` is the filepath, `load_checkpoint.hash` is the trusted blockhash against which the checkpoint in the file will be validated.
 * `blocks_per_epoch`: currently not directly used, will be the boundary for validator committee changes in the future. Added because checkpoints are specified in the number of epochs
 * `epochs_per_checkpoint`: together with `blocks_per_epoch`, determines the frequency of checkpoint snapshots.

Checkpoints are exported inside the node's datadir, at the path `/checkpoints/{block_number}`.

Note for infra: this means that to make snapshots available for testnet, we'll need to:
a) enable `do_snapshots` on at least one node; and
b) make the snapshot file from the latest block available for download.

Users will need to download the file, obtain the blockhash at that block height (e.g. with a block explorer) and pass the appropriate `load_checkpoint` config.


zilliqa/src/db.rs
```diff
@@ -235,12 +354,21 @@ impl Db {
         Ok(sqlite_tx.commit()?)
     }
 
-    pub fn put_canonical_block_number(&self, number: u64, hash: Hash) -> Result<()> {
-        self.block_store.lock().unwrap().execute("INSERT OR REPLACE INTO main_chain_canonical_blocks (height, block_hash) VALUES (?1, ?2)",
+    pub fn set_canonical_block_number_with_db_tx(
```
theo-zil, line 357: Idle thought: I considered reducing boilerplate by using a macro to generate all the `_with_db_tx` pairs of functions, but haven't spent the time on it. Might be worth to render the file more readable.

zilliqa/src/consensus.rs
```diff
@@ -1446,6 +1442,7 @@ impl Consensus {
         };
 
         // At genesis it could be fine not to have a qc block, so don't error.
+        // TODO: this should be an explicit check for genesis
```
JamesHinshelwood, line 1445: Raise an issue?

zilliqa/src/db.rs
```diff
@@ -221,6 +235,111 @@ impl Db {
...
+
+    pub fn load_trusted_checkpoint<P: AsRef<Path>>(
+        &self,
+        path: P,
+        hash: &Hash,
+        our_shard_id: u64,
+    ) -> Result<Block> {
+        let input = File::open(path)?;
+        let reader = BufReader::with_capacity(8192 * 1024, input); // 8 MiB read chunks
+        let trie_storage = Arc::new(self.state_trie()?); // TODO: what should we do if the trie DB already contains values?
```
JamesHinshelwood, line 256: Can you clarify what you mean by this?

zilliqa/src/db.rs
```diff
@@ -221,6 +235,111 @@ impl Db {
...
+        }
+
+        let our_max_view = self.get_latest_finalized_view()?.unwrap_or(0);
+        if block.view() <= our_max_view {
+            return Err(anyhow!("Snapshow is lower than our finalized view: snapshot is at {}, we are already at {}", block.view(), our_max_view));
+        }
+
+        // then decode state
+        for (idx, line) in lines.enumerate() {
+            let idx = idx + 2; // +1 because first line is just serialised block, +1 for 1-indexing
```
JamesHinshelwood, line 297: Didn't we already read the serialised block above this?

zilliqa/src/db.rs
```diff
@@ -584,6 +726,63 @@ impl Db {
     }
 }
 
+pub fn snapshot_block_with_state<P: AsRef<Path> + Debug>(
```
JamesHinshelwood, line 729: Could we document the snapshot file format separately? As it stands, the only way to work it out is by reading both the serialisation and deserialisation code.

zilliqa/src/db.rs
```diff
@@ -221,6 +235,111 @@ impl Db {
...
+    }
+
+    pub fn load_trusted_checkpoint<P: AsRef<Path>>(
+        &self,
+        path: P,
+        hash: &Hash,
+        our_shard_id: u64,
+    ) -> Result<Block> {
+        let input = File::open(path)?;
+        let reader = BufReader::with_capacity(8192 * 1024, input); // 8 MiB read chunks
```
JamesHinshelwood, line 255: Is there a reason we picked 8 MiB? The default is 8 KiB - Was that bad for some reason?

zilliqa/src/cfg.rs
```diff
@@ -52,6 +55,12 @@ pub struct NodeConfig {
     /// The location of persistence data. If not set, uses a temporary path.
     #[serde(default)]
     pub data_dir: Option<String>,
+    /// Persistence checkpoint to load.
+    #[serde(default)]
+    pub load_checkpoint: Option<Checkpoint>,
+    /// Whether to enable exporting checkpoint state snapshot files.
+    #[serde(default)]
+    pub do_snapshots: bool,
```
JamesHinshelwood, line 63: Is there a reason we're using both "checkpoint" and "snapshot"? Can we use just one of the words consistently?

JamesHinshelwood (dismissed):

zilliqa/src/db.rs
theo-zil: As in, if you're loading a checkpoint but you already have previous state in the DB. For example:
 * You used to run a node, it went out of sync, then you loaded a checkpoint to catch back up (this should be fine). Or...
 * You ran your node on another shard, then switched shards and are loading a checkpoint (maybe bad?)
 * You ran e.g. a testnet and then switched to mainnet and are syncing from a checkpoint without clearing the DB (maybe bad?)
 * You're doing weird stuff, like running a synced node, then maybe deleting/losing/corrupting your sqlite DB (so it no longer records your highest view) and then re-load an earlier checkpoint

Right now this PR just ignores it. Which I _think_ is probably fine assuming there are no hash collisions - I feel like the worst you'll have is a bloated storage with useless values in it. But is even that fine, or do we want to throw a warning saying "your key-value storage seems to have old values that may be useless in it, are you sure?" - since I'd expect the main usecase for checkpoints, I assume, would be to bring up brand new nodes with an empty key-value store.

zilliqa/src/cfg.rs
theo-zil: Good point. In theory, in my mind, "checkpoint" - short for - "weak subjectivity checkpoint" - is the height and trusted hash at which "snapshot" files are exported. Then, when starting a node, you download a snapshot file to be able to start your node from the corresponding w.s. checkpoint's height.

 In practice, I think you're right and it just ends up inconsistent. Any preference? If not I'll stick with "checkpoint"  (so we'll e.g. be exporting checkpoints, have checkpoint files, etc.).

zilliqa/src/db.rs
theo-zil: If we're potentially going to be reading/writing multi-GB files, I think a slightly larger batch size is warranted. The only tradeoff is memory (to my understanding at least...) and 8MiB seems like perfectly reasonable usage, and in return we will have 1000x fewer I/O syscalls when exporting and importing huge files.

zilliqa/src/db.rs
theo-zil: I've considered this but I feel like the only people who would want to manually edit them would also be able to read the code.
However since I've considered it and you think it's also useful then maybe worth doing after all. Any preference on where to put it? I'm thinking just sticking a new file under `docs/`.

zilliqa/src/db.rs
theo-zil: Yup, the `idx` is just a uint used for debugging/error messages when parsing. We pop off the first three lines, then start the loop so the iteration number = line number - 3.

However this is a holdover from when I only had the serialized block and nothing else - now I have it, the parent and the shard ID so it should actually be +4 here.

zilliqa/src/cfg.rs
JamesHinshelwood: Checkpoint everywhere sounds good to me :)

I agree that maybe you could call the height a "checkpoint" and the actual file a "snapshot", but I don't think we really need a separate word to describe the height.

zilliqa/src/db.rs
rrw-zilliqa: It's going to be very hard to review PRs that touch either ser or de code for checkpoints without knowing what the intended format is. docs/ works for me. See ^^^ for a suggestion that might help with this as well (you can then use the rust struct defn you give to serde as a spec).

zilliqa/src/db.rs
```diff
@@ -584,6 +726,63 @@ impl Db {
...
+        // now the account storage
+        for (storage_key, storage_val) in account_storage.iter() {
+            writer.write_all(hex::encode(storage_key).as_bytes())?;
+            writer.write_all(b":")?;
+            writer.write_all(hex::encode(storage_val).as_bytes())?;
+            writer.write_all(b",")?;
+        }
+        writer.write_all(b"\n")?;
+    }
+    writer.flush()?;
```
rrw-zilliqa, line 782: How do I tell when you've finished writing the snapshot? (If I were you, I'd write snapshots to a temporary filename, then rename when done - this ensures that the name does not become available until the contents are valid - you might also usefully write a hash of the snapshot either to the snapshot or to a separate file to protect against errors in transit).

rrw-zilliqa (dismissed):
I'm not sure I rate your chances of exporting mainnet persistence to a text file!

Also, we seem to write a lot of checkpoints, never clean them up, and I'm not sure how checkpoints get shared - does the user manually transport them from place to place? If so, why do we generate them automatically given that you're unlikely to want more than one in a blue moon .. ? If not, how do other nodes findn and ingest them (and who from)?

zilliqa/src/consensus.rs
```diff
@@ -627,23 +630,16 @@ impl Consensus {
...
         committee: &[NodePublicKey],
+        parent_block: &Block,
         view: u64,
         cosigned: &BitSlice,
     ) -> Result<()> {
         debug!("apply rewards in view {view}");
 
         let rewards_per_block =
             *self.config.consensus.rewards_per_hour / self.config.consensus.blocks_per_hour as u128;
-        let block = self.head_block();
```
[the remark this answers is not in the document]
theo-zil: It's more that the previous implementation was confusing (in the sense of "this seems obviously wrong, why is it this way") - I think my change should be logical enough. Unless Zoltan says otherwise!

zilliqa/src/db.rs
theo-zil: The use case is probably the first bullet point - you have some previous block, your node went down, you skip ahead using a checkpoint. If we disallow previous state, that effectively means you'd have to wipe your node and start fresh if it ever gets out of sync
Which may be fine anyway if people start running pruned nodes as a rule... So I don't have a strong opinion. Mind you the checkpoints only contain the actual state in-use at the checkpoint block height, so old state or otherwise unused bloated state won't propagate through them.

zilliqa/src/db.rs
```diff
@@ -221,6 +235,111 @@ impl Db {
...
+        path: P,
+        hash: &Hash,
+        our_shard_id: u64,
+    ) -> Result<Block> {
+        let input = File::open(path)?;
+        let reader = BufReader::with_capacity(8192 * 1024, input); // 8 MiB read chunks
+        let trie_storage = Arc::new(self.state_trie()?); // TODO: what should we do if the trie DB already contains values?
+        let mut state_trie = EthTrie::new(trie_storage.clone());
+
+        let mut lines = reader.lines();
```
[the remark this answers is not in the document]
theo-zil: I think I'm mostly worried about memory usage. I wouldn't wanna load e.g. 10-20GB of persistence into memory as one big string.
Unless there are streaming serde implementations? I confess I haven't looked into this, I just assumed most `serde_foo::from_str()`/`::from_file()` methods would load the entire file into memory.

A binary format might be a worthwhile improvement.

zilliqa/src/db.rs
```diff
@@ -221,6 +235,111 @@ impl Db {
...
+                ));
+            }
+            state_trie.insert(&hex::decode(account_hash)?, &serialized_account)?;
+        }
+        if state_trie.root_hash()? != block.state_root_hash().0 {
+            return Err(anyhow!("Invalid checkpoint file: state root hash mismatch"));
+        }
+
+        let block_ref = &block; // for moving into the closure
+        self.with_sqlite_tx(move |tx| {
```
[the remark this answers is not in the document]
theo-zil: Good point.
Now I _think_ it's not actually that bad because you'll fetch blocks from the network, realise you're out of date, and request a sync - and then you'll just magically find that the blocks your receive are in your database, so you update your "latest view", request the next batch, magically find you already had those blocks too, etc...
But I'm not certain this is _guaranteed_ to work without any weirdness or misbehaviour, so agree that it would be safer to avoid this scenario entirely.

zilliqa/src/db.rs
theo-zil: Excellent point, thanks!

zilliqa/src/consensus.rs
theo-zil: Actually on looking further, this isn't a hole in block validation, it just skips finalizing the block if we have insufficient data to do so. Since this can now happen not only on genesis but also when loading from a checkpoint, I've just updated the comments to be more clear; no issue necessary.

zilliqa/src/cfg.rs
```diff
@@ -172,6 +190,10 @@ pub struct ConsensusConfig {
     pub blocks_per_hour: u64,
     pub minimum_stake: Amount,
     pub eth_block_gas_limit: EvmGas,
+    #[serde(default = "blocks_per_epoch_default")]
+    pub blocks_per_epoch: u64,
+    #[serde(default = "epochs_per_checkpoint_default")]
+    pub epochs_per_checkpoint: u64,
```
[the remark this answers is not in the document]
theo-zil: I thought we already had an issue but didn't find one. Good call
#1135

zilliqa/src/consensus.rs
```diff
@@ -1503,6 +1503,33 @@ impl Consensus {
             }
         }
 
+        if block.number() % self.config.consensus.blocks_per_epoch == 0 && block.number() != 0 {
+            // TODO: handle epochs
```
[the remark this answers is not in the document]
theo-zil: #1140

zilliqa/src/db.rs
```diff
@@ -221,6 +235,111 @@ impl Db {
         Ok(())
     }
 
+    pub fn create_checkpoint_path(&self, height: u64) -> Result<Option<Box<Path>>> {
+        let Some(base_path) = &self.path else {
+            // If we don't have on-disk persistency, disable checkpoints too
+            return Ok(None);
```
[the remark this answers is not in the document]
theo-zil: Warning added, but now I'm thinking that maybe we shouldn't allow `do_checkpoints = true` at the same time as `data_dir = None`, as that will simply lead to unexpected behaviour.
We could add some validation to the config, or simply error out here if we try to export a checkpoint without having a datadir (ensuring the user isn't left wondering why the heck their node is running fine but not creating any checkpoints...)

On the other hand maybe ephemeral nodes aren't even an important enough usecase to warrant trying to support them carefully, I might be overthinking this.

zilliqa/src/consensus.rs
```diff
@@ -1503,6 +1503,33 @@ impl Consensus {
...
+            if self.config.do_snapshots
+                && block.number()
+                    % (self.config.consensus.blocks_per_epoch
+                        * self.config.consensus.epochs_per_checkpoint)
+                    == 0
+            {
+                if let Some(snapshot_path) = self.db.create_checkpoint_path(block.number())? {
```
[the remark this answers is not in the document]
theo-zil: I've now added old checkpoing cleanup - since you mention it, I don't see a good reason to keep _any_ old checkpoints, so I just clean the entire directory every time the new one is saved (after it's successfully saved, of course, so we always have at least one valid checkpoint!).

docs/checkpoints.md
```diff
@@ -0,0 +1,26 @@
+# State checkpoints
+If enabled in the config, every _epoch length * checkpoint interval_ blocks, the node will export a state checkpoint in its datadir. This snapshot can then be imported into a brand new node to serve as a checkpoint from which to start syncing (rather than using genesis), fast-tracking the sync.
+
+## Configuration
+ * `do_checkpoints`, *boolean (default `false`)*: If `true`, enables exporting the checkpoint files. If false, checkpoints will not be created.
+ * `consensus.epochs_per_checkpoint`, *u64 (default 24)*: If `do_checkpoints` is true, determines the frequency (in epochs) at which checkpoints will be exported. This is a consensus property and will eventually be obtained on-chain rather than being configurable.
```
JamesHinshelwood, line 6: > This is a consensus property and will eventually be obtained on-chain rather than being configurable.

I think we should remove this sentence. I'm not 100% confident that we want to move all of this configuration on-chain yet, so I'd rather not promise that we will.

docs/checkpoints.md
```diff
@@ -0,0 +1,26 @@
...
+ * `load_checkpoint`, *struct, optional (default `None`)*: If provided, on startup, the node will attempt to load the given checkpoint and start syncing from it. Fails if the node has a non-empty database, i.e. intended to be used only for brand new nodes without existing data.
```
JamesHinshelwood, line 7:
```suggestion
 * `load_checkpoint`, *struct, optional (default `None`)*: If provided, on startup, the node will attempt to load the given checkpoint and start syncing from it. The node will fail to start if this is set and the database is non-empty. This is intended to be used for new nodes.
```

docs/checkpoints.md
```diff
@@ -0,0 +1,26 @@
...
+   * `load_checkpoint.file`, *filepath*: The file to read the checkpoint from
+   * `load_checkpoint.hash`, *string*: The hex-encoded hash of the checkpoint block. This value is assumed to be trusted; the checkpoint contents are verified against this hash.
+
+## Directory
+Checkpoint files are saved inside the node's data directory, at the path `/checkpoints/{block_height}`.
+
+If the node does not have a data directory (i.e. is running on an ephemeral in-memory database), no checkpoints will be exported, regardless of the `do_checkpoints` parameter.
+
+Currently, old checkpoints are _not_ kept. Once a new checkpoint file is successfully created, any other files in the directory are deleted.
```
JamesHinshelwood, line 16: What motivated this choice? Given we are deferring responsibility for uploading/publishing the checkpoints to external tooling, I'd argue we should defer the responsibility of cleanup to the same tool. Otherwise, we run the possibility of the node deleting checkpoints from underneath the tool trying to upload them.

docs/checkpoints.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Checkpoint file format
+TBA
+Currently uses a text file format roughly as follows:
+line 1 - the checkpoint block
+line 2 - the parent block, allowing the node to establish a chain of 2 blocks for finalizing future blocks
+line 3 - the shard ID, for sanity checking
+all subsequent lines - the serialized state at the checkpoint block. The state consists of account tries, serialized at one account per line, in the format: `{database_hash}:{serialized_account}` followed by a `;` followed by every node in the account's storage in the format `{storage_key}:{storage_value}` separated by `,`
+
+Before the PR is merged, a version number will be added, and the format might move to binary
```
JamesHinshelwood, line 26: Will it?

docs/checkpoints.md
```diff
@@ -0,0 +1,26 @@
...
+## Directory
+Checkpoint files are saved inside the node's data directory, at the path `/checkpoints/{block_height}`.
+
+If the node does not have a data directory (i.e. is running on an ephemeral in-memory database), no checkpoints will be exported, regardless of the `do_checkpoints` parameter.
+
+Currently, old checkpoints are _not_ kept. Once a new checkpoint file is successfully created, any other files in the directory are deleted.
+
+## Checkpoint file format
+TBA
+Currently uses a text file format roughly as follows:
```
JamesHinshelwood, line 20: If we're documenting the format, I think we should do it non-roughly :stuck_out_tongue:

zilliqa/src/consensus.rs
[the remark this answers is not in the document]
JamesHinshelwood: As discussed on Slack, should we remove this change and fix the issue as part of #1143 instead?

zilliqa/src/crypto.rs
```diff
@@ -291,6 +290,28 @@ impl Hash {
     }
 }
 
+impl Serialize for Hash {
+    fn serialize<S>(&self, serializer: S) -> std::prelude::v1::Result<S::Ok, S::Error>
```
JamesHinshelwood, line 294:
```suggestion
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```

zilliqa/src/crypto.rs
```diff
@@ -291,6 +290,28 @@ impl Hash {
...
+    where
+        S: serde::Serializer,
+    {
+        hex::encode(self.0).serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for Hash {
+    fn deserialize<D>(deserializer: D) -> std::prelude::v1::Result<Self, D::Error>
```
JamesHinshelwood, line 303:
```suggestion
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
```

zilliqa/src/crypto.rs
```diff
@@ -291,6 +290,28 @@ impl Hash {
     }
 }
 
+impl Serialize for Hash {
+    fn serialize<S>(&self, serializer: S) -> std::prelude::v1::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        hex::encode(self.0).serialize(serializer)
```
JamesHinshelwood, line 298: Are we sure we want to serialise hashes as a hex-string rather than a byte array?

JamesHinshelwood (dismissed):
LGTM, but maybe we should resolve the 'TBA' in the docs before merging.

JamesHinshelwood (approved):

PR #1122, opened 2024-07-01
Select nodes to request blocks from more intelligently

### LINKED ISSUE #1097: Nodes should send direct messages to all nodes, not just committee members
Right now `get_random_peer` can only return a member of the committee.

If I was doing this, I would add another message channel for "send a message to a single node, but I don't care who that node is, just pick a random peer" and handle the selection at the `p2p_node` layer instead. This also removes the weird non-cryptographic RNG in `Consensus`.

JamesHinshelwood:
Previously, we requested missing blocks from a random node in the consensus committee. Now, we can request blocks from any node who has told us they have a block with at least that view.

In future, we can extend this logic by:
* Taking into account latency, bad responses, etc. when choosing peers.
* Extending the P2P protocol so that non-validator peers can share their 'best' block with each other, meaning we can download blocks from those non-validators.

Resolves #1097 since this was the only thing we used `get_random_peer` for.

### DIFF
--- a/zilliqa/src/block_store.rs
+++ b/zilliqa/src/block_store.rs
@@ -1,4 +1,4 @@
-use std::{cell::RefCell, num::NonZeroUsize, sync::Arc};
+use std::{cell::RefCell, collections::HashMap, num::NonZeroUsize, sync::Arc};
 
 use anyhow::{anyhow, Result};
 use libp2p::PeerId;
@@ -34,6 +34,8 @@ pub struct BlockStore {
     block_cache: RefCell<LruCache<Hash, Block>>,
     /// The maximum view of any proposal we have received, even if it is not part of our chain yet.
     highest_known_view: u64,
+    /// Information we keep about our peers' state.
+    peers: HashMap<PeerId, PeerInfo>,
     /// The maximum view of blocks we've sent a request for.
     requested_view: u64,
     /// The maximum number of blocks to send requests for at a time.
@@ -50,12 +52,19 @@ pub struct BlockStore {
     message_sender: MessageSender,
 }
 
+#[derive(Clone, Debug, Default)]
+struct PeerInfo {
+    highest_known_view: u64,
+    requested_blocks: u64,
+}
+
 impl BlockStore {
     pub fn new(config: &NodeConfig, db: Arc<Db>, message_sender: MessageSender) -> Result<Self> {
         Ok(BlockStore {
             db,
             block_cache: RefCell::new(LruCache::new(NonZeroUsize::new(5).unwrap())),
             highest_known_view: 0,
+            peers: HashMap::new(),
             requested_view: 0,
             max_blocks_in_flight: config.max_blocks_in_flight,
             batch_size: config.block_request_batch_size,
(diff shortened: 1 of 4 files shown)


zilliqa/src/block_store.rs
```diff
@@ -50,12 +52,19 @@ pub struct BlockStore {
     message_sender: MessageSender,
 }
 
+#[derive(Clone, Debug, Default)]
+struct PeerInfo {
+    highest_known_view: u64,
+    requested_blocks: u64,
```
rrw-zilliqa, line 58: I bet this is being reworked, so feel free to ignore me, but would be good to have comments here to say what gets stored in these (and how reliable it needs to be)

If it were me, I'd abstract this out - have a block_fetcher.rs and make peers a block_fetcher::Context - that makes it rather clearer where the logic is, and gives you more contained diffs when you inevitably make the data structure more complex.

Eventually,  you also want to age out this data so that misconceptions and errors correct themselves over time (in particular, eventually - actually quite soon, if Theo's PR is to be believed - highest_known_view will need to be joined by lowest_known_view, or more likely a Range .. )

zilliqa/src/block_store.rs
```diff
@@ -82,12 +87,27 @@ impl BlockStore {
...
         Ok(())
     }
 
-    pub fn request_missing_blocks(&mut self, random_peer: Option<PeerId>) -> Result<()> {
+    pub fn best_peer(&self, view: u64) -> Option<PeerId> {
+        let (best, _) = self
+            .peers
+            .iter()
+            .filter(|(_, peer)| peer.highest_known_view >= view)
+            .min_by_key(|(_, peer)| peer.requested_blocks)?;
```
rrw-zilliqa, line 106: Hopefully, "most" peer.requested_blocks will be -1, which means you'll hammer whichever one min_by_key finds first. Not important, but I thought I'd say.

rrw-zilliqa (approved):
This is perfectly safe and sensible:

 - I've not said a bunch of things here because @JamesHinshelwood and I went over them in a call yesterday. Let me know if you want them here for visibility.
 -  I think it would be good to start moving the (increasingly complex) block fetch logic out into a separate file.

Comments are all purely advice - I'd expect you to ignore them!

JamesHinshelwood:
> I think it would be good to start moving the (increasingly complex) block fetch logic out into a separate file.

How convinced are you if I argue that `block_store.rs` is that separate file? All it is responsible for is fetching blocks and storing those blocks and its <300 lines. IMO there isn't enough going on in `BlockStore` yet to justify another layer of abstraction.

zilliqa/src/block_store.rs
JamesHinshelwood: My next indeed changes this so I can retry failed requests - I will add extra details there.

zilliqa/src/block_store.rs
JamesHinshelwood: Hmm good point, it would be nice if we shuffled through all equally good peers. I'll raise an issue for this.

JamesHinshelwood: https://github.com/Zilliqa/zq2/issues/1133

Issue #1106, opened 2024-06-29
View timeouts take too long


DrZoltanFazekas:
We  tested shutting down one of the 4 validators. In views when the offline validator was the leader, it didn't propose a block, but the rest of the validators behaved according to the protocol: they sent new view messages to the next leader and the next leader proposed a block. However, the gap caused by the missing proposal in the failed views was much longer than expected the 5 sec timeout.

Here's a log of how long the gaps took:
```
view  block          miner                                      gas txns
-------------------------------------------------------------------------
56479 56461 06:48:36 0xcca93a2e1169caf02e515203d1539cc3f390890a 0 []
####################### view 56480 missing - 20 sec
56481 56461 06:48:56 0x9a21038511f75492bac85e3d6eae5b37c3fbbdca 0 []
####################### view 56482 missing - 80 sec
56483 56461 06:50:17 0xcca93a2e1169caf02e515203d1539cc3f390890a 0 []
56484 56462 06:50:18 0x3e422c617eb2880f7ac7be512755e0e103bb916b 0 []
56485 56463 06:50:19 0x3e422c617eb2880f7ac7be512755e0e103bb916b 0 []
56486 56464 06:50:20 0x9a21038511f75492bac85e3d6eae5b37c3fbbdca 0 []
56487 56465 06:50:21 0x9a21038511f75492bac85e3d6eae5b37c3fbbdca 0 []
56488 56466 06:50:23 0x3e422c617eb2880f7ac7be512755e0e103bb916b 0 []
56489 56467 06:50:24 0x9a21038511f75492bac85e3d6eae5b37c3fbbdca 0 []
56490 56468 06:50:25 0xcca93a2e1169caf02e515203d1539cc3f390890a 0 []
56491 56469 06:50:26 0x3e422c617eb2880f7ac7be512755e0e103bb916b 0 []
56492 56470 06:50:27 0xcca93a2e1169caf02e515203d1539cc3f390890a 0 []
56493 56471 06:50:29 0x9a21038511f75492bac85e3d6eae5b37c3fbbdca 0 []
####################### view 56495 missing - 60 sec
56496 56471 06:51:29 0xcca93a2e1169caf02e515203d1539cc3f390890a 0 []
####################### view 56497 missing - 160 sec
56498 56471 06:54:10 0x9a21038511f75492bac85e3d6eae5b37c3fbbdca 0 []
[...]

PR #1125, opened 2024-07-02
fix too long consensus timeouts

DrZoltanFazekas:
This will fix https://github.com/Zilliqa/zq2/issues/1106 which also describes why our timeouts take 4x longer than they should.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -450,7 +450,7 @@ impl Consensus {
             .expect("last timeout seems to be in the future...")
             .as_millis() as u64;
         let view_difference = self.view.get_view().saturating_sub(self.high_qc.view);
-        let exponential_backoff_timeout = consensus_timeout_ms * 2u64.pow(view_difference as u32);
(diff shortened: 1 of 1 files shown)


zilliqa/src/consensus.rs
```diff
@@ -450,7 +450,7 @@ impl Consensus {
             .expect("last timeout seems to be in the future...")
             .as_millis() as u64;
         let view_difference = self.view.get_view().saturating_sub(self.high_qc.view);
-        let exponential_backoff_timeout = consensus_timeout_ms * 2u64.pow(view_difference as u32);
+        let exponential_backoff_timeout = consensus_timeout_ms * 2u64.pow((view_difference as u32).saturating_sub(2));
```
JamesHinshelwood, line 453: If you can be bothered a comment on the line before which explains why we minus 2 would be useful :)

JamesHinshelwood (approved):

JamesHinshelwood (approved):

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):

PR #1136, opened 2024-07-04
Retry failed block requests

JamesHinshelwood:
This required quite a lot of additional machinery in order to inform the `BlockStore` of request failures, so it can retry.

Now, whenever we send a direct external message in `node.rs`, we remember a `request_id` which is returned to the caller. One level up, we keep a correspondence of our generated `request_id` to libp2p's request ID. When a request succeds or fails, we can look up the `request_id` in this map and pass it back down the stack to inform the initial caller of the result.

We've also changed the type which gets sent from the network layer to the `node.rs` layer from an `ExternalMessage` to a `Result<ExternalMessage, OutgoingMessageFailure>`. The `Err` half is how failed messages are passed down.

### DIFF
--- a/.github/workflows/docker-release.yaml
+++ b/.github/workflows/docker-release.yaml
@@ -5,7 +5,7 @@ on:
     branches:
       - 'release/**'
       - 'main'
-      - fix-sync
+      - dl-retry
     tags:        
       - v*
 
@@ -60,7 +60,7 @@ jobs:
       permissions:
         id-token: write
         contents: write
-      if: ${{ github.ref_name == 'main' }} || ${{ github.ref_name == 'fix-sync' }}
+      if: ${{ github.ref_name == 'main' }} || ${{ github.ref_name == 'dl-retry' }}
       runs-on: [ self-hosted, gcp]
       env:
         GCP_REGISTRY_DOMAIN: asia-docker.pkg.dev
--- a/zilliqa/src/block_store.rs
+++ b/zilliqa/src/block_store.rs
@@ -43,6 +43,8 @@ pub struct BlockStore {
     max_blocks_in_flight: u64,
     /// The maximum number of blocks to request at a time.
     batch_size: u64,
+    /// When a request to a peer fails, do not send another request to this peer for this amount of time.
+    failed_request_sleep_duration: Duration,
     /// Buffered block proposals, indexed by their parent hash. These are proposals we've received, whose parents we
     /// haven't yet seen. We want to be careful about buffering too many proposals. There is no guarantee or proof that
     /// any of them will eventually form our canonical chain. Therefore we limit the number of buffered proposals to
@@ -50,8 +52,8 @@ pub struct BlockStore {
     /// handle pending requests made by ourselves that arrive out-of-order, while also giving some extra space for
     /// newly created blocks that arrive while we are syncing.
     buffered: LruCache<Hash, Proposal>,
-    /// Requests we would like to sent, but haven't been able to (e.g. because we have no peers).
-    pending_requests: Vec<(u64, u64)>,
+    /// Requests we would like to send, but haven't been able to (e.g. because we have no peers).
+    unserviceable_requests: Vec<(u64, u64)>,
     message_sender: MessageSender,
 }
 
@@ -88,10 +90,11 @@ impl BlockStore {
             requested_view: 0,
             max_blocks_in_flight: config.max_blocks_in_flight,
             batch_size: config.block_request_batch_size,
+            failed_request_sleep_duration: config.failed_request_sleep_duration,
             buffered: LruCache::new(
                 NonZeroUsize::new(config.max_blocks_in_flight as usize + 100).unwrap(),
             ),
-            pending_requests: vec![],
+            unserviceable_requests: vec![],
             message_sender,
         })
     }
@@ -128,7 +131,7 @@ impl BlockStore {
             .filter(|(_, peer)| {
                 peer.last_request_failed_at
                     .and_then(|at| at.elapsed().ok())
-                    .map(|time_since| time_since > Duration::from_secs(10))
+                    .map(|time_since| time_since > self.failed_request_sleep_duration)
                     .unwrap_or(true)
             })
             .min_by_key(|(_, peer)| peer.requested_blocks)?;
@@ -150,8 +153,11 @@ impl BlockStore {
             self.requested_view = current_view;
         }
 
-        // Attempt to send pending requests if we can.
-        while let Some((from, to)) = self.pending_requests.pop() {
+        // Attempt to send pending requests if we can. Note that unserviceable requests are retried in LIFO order. This
+        // effectively means that requests for higher views are sent first. Also, we break early if a peer can't
+        // service the final request, despite the possibility that it could service a request for lower views. If this
+        // appears to be a problem in practice, we could use a `VecDeque` and retry requests in FIFO order.
+        while let Some((from, to)) = self.unserviceable_requests.pop() {
             if !self.request_blocks(from, to)? {
                 // Stop trying to send requests if no peers are available.
                 break;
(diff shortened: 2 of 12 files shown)


zilliqa/src/block_store.rs
```diff
@@ -103,6 +124,13 @@ impl BlockStore {
             .peers
             .iter()
             .filter(|(_, peer)| peer.highest_known_view >= view)
+            // If the last request failed, don't send requests to this peer for 10 seconds.
+            .filter(|(_, peer)| {
+                peer.last_request_failed_at
+                    .and_then(|at| at.elapsed().ok())
+                    .map(|time_since| time_since > Duration::from_secs(10))
```
rrw-zilliqa, line 131: Perhaps split this constant out into however you're managing constants these days? Having it embedded in the code makes it kinda hard to see.

zilliqa/src/block_store.rs
```diff
@@ -49,13 +50,32 @@ pub struct BlockStore {
     /// handle pending requests made by ourselves that arrive out-of-order, while also giving some extra space for
     /// newly created blocks that arrive while we are syncing.
     buffered: LruCache<Hash, Proposal>,
+    /// Requests we would like to sent, but haven't been able to (e.g. because we have no peers).
+    pending_requests: Vec<(u64, u64)>,
```
rrw-zilliqa, line 54: You're using the same name (pending) for requests we've sent to a peer (in PeerInfo) and requests you couldn't send. Perhaps change one of them to avoid confusion? unserviceable_requests?

zilliqa/src/block_store.rs
```diff
@@ -122,9 +150,22 @@ impl BlockStore {
             self.requested_view = current_view;
         }
 
+        // Attempt to send pending requests if we can.
+        while let Some((from, to)) = self.pending_requests.pop() {
+            if !self.request_blocks(from, to)? {
+                // Stop trying to send requests if no peers are available.
+                break;
+            }
```
rrw-zilliqa, line 158: How do we cope if no peers are available for an extended period and self.pending_requests starts to get big?
I guess it probably doesn't matter too much, since we will need all these blocks in the order in which they appear in the list in any event, and so pruning it from either end will do us no good. We might consider (perhaps a comment here?) throttling the number of these we send at once to avoid becoming a noisy neighbour when the network comes back and we have a long list of self.pending_requests...

zilliqa/src/block_store.rs
```diff
@@ -134,26 +175,36 @@ impl BlockStore {
...
 
         Ok(())
     }
 
+    /// Make a request for a range of blocks. Returns `true` if a request was made and `false` if the request had to be
+    /// buffered because no peers were available.
+    fn request_blocks(&mut self, from: u64, to: u64) -> Result<bool> {
+        let Some(peer) = self.best_peer(from) else {
+            warn!(from, "no peers to download missing blocks from");
+            self.pending_requests.push((from, to));
```
rrw-zilliqa, line 190: this has the effect that pending requests are retried in LIFO order. intentional? (I don't think it greatly matters, but very marginally, presumably more nodes have older blocks and older requests are likely to be for those older blocks?)

zilliqa/src/p2p_node.rs
```diff
@@ -306,27 +310,39 @@ impl P2pNode {
...
+                                let _ = self.swarm.behaviour_mut().kademlia.bootstrap();
+                            }
+
+
+                            if let Some((shard_id, request_id)) = self.pending_requests.remove(&request_id) {
+                                let error = OutgoingMessageFailure { peer, request_id, error };
+                                self.forward_external_message_to_node(&Self::shard_id_to_topic(shard_id).hash(), peer, Err(error))?;
+                            } else {
+                                return Err(anyhow!("request without id failed"));
+                            }
```
rrw-zilliqa, line 345: This deals with the case where you can't send a block request to a peer; how about if it never replies?

rrw-zilliqa (dismissed):
Generally LGTM -  thank you! - some comments inline.

zilliqa/src/p2p_node.rs
JamesHinshelwood: Sorry should have clarified this in the PR description. I have a future plan to alter how we request blocks which I think will fix this.

`libp2p_request_response` requires that every request has a response and will report a request failure if a response isn't received within a configurable amount of time. However, currently we aren't utilising this because of the way we've structured our messages:

**Current state**
A -> B: `libp2p.send_request(BlockRequest)`
B -> A: `libp2p.send_response(Ack)`
B -> A: `libp2p.send_request(BlockResponse)`
B -> A: `libp2p.send_response(Ack)`

**Desired state**
A -> B: `libp2p.send_request(BlockRequest)`
B -> A: `libp2p.send_response(BlockResponse)`

This will mean a node's failure to respond to a `BlockRequest` will result in an `OutgoingMessageFailure` (of type `OutboundFailure::Timeout`) and this error handling code will be called.

If you agree with this plan, I'll raise an issue to do this before merging this PR :)

zilliqa/src/block_store.rs
JamesHinshelwood: The length of `self.unserviceable_requests` is (implicitly) bounded by `self.max_blocks_in_flight`.

I do agree that being a noisy neighbour might be a problem - If we learn about a single node at first, that poor guy will have to service all of our requests. Will raise an issue on this. I think this should probably be based on changing `self.max_blocks_in_flight` to a per-peer limit instead?

zilliqa/src/block_store.rs
JamesHinshelwood: Correct. I figured it didn't matter since the length of `self.unserviceable_requests` is bounded and once a peer appears, we will send all the requests to it anyway (i.e. the list will be emptied before any of the responses are received).

However, you make a good point about block availability (i.e. its sad if we break early from our `self.unserviceable_requests.pop()` loop because the peer couldn't service a large block number when it could in fact service the lower numbers). I guess I could argue that this case will resolve itself quickly once we learn that said peer can now service a greater block number.

I will add a comment to clarify the current behaviour and suggest changing it in the future if we see problems?

zilliqa/src/p2p_node.rs
rrw-zilliqa: I agree with this plan, but check (in a separate issue, I guess?) that malicious nodes can't crash the network by causing a lot of messages to be sent to them and then vanishing, repeatedly, causing the calling node to run out of memory to store pending request records. It's one of those failure modes that is either trivial or impossible, so worth checking - if you're lucky, there is a max # pending messages in libp2p and you can set it high enough to make that situation impossible.

rrw-zilliqa:
PS - I think we've discussed this before, but it's safe for stuff to expire from the LRU cache, right?

.github/workflows/docker-release.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
       permissions:
         id-token: write
         contents: write
-      if: ${{ github.ref_name == 'main' }} || ${{ github.ref_name == 'fix-sync' }}
+      if: ${{ github.ref_name == 'main' }} || ${{ github.ref_name == 'dl-retry' }}
```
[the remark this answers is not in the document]
JamesHinshelwood: I blame @frankmeds :)

zilliqa/src/block_store.rs
```diff
@@ -42,20 +43,41 @@ pub struct BlockStore {
...
+    failed_request_sleep_duration: Duration,
     /// Buffered block proposals, indexed by their parent hash. These are proposals we've received, whose parents we
     /// haven't yet seen. We want to be careful about buffering too many proposals. There is no guarantee or proof that
     /// any of them will eventually form our canonical chain. Therefore we limit the number of buffered proposals to
     /// `max_blocks_in_flight + 100`. This number is chosen because it makes it probable we will always be able to
     /// handle pending requests made by ourselves that arrive out-of-order, while also giving some extra space for
     /// newly created blocks that arrive while we are syncing.
     buffered: LruCache<Hash, Proposal>,
+    /// Requests we would like to send, but haven't been able to (e.g. because we have no peers).
+    unserviceable_requests: Vec<(u64, u64)>,
```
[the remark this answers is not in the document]
JamesHinshelwood: Fair point, I should be using a `RangeInclusive` - I'll do this in the next PR

Issue #1142, opened 2024-07-05
Invalid view changes


DrZoltanFazekas:
We submitted 40 transactions to the network while polling blocks and logging the following information:
```
view height timestamp proposer gas_used transactions
----------------------------------------------------------------
3162 3156 13:30:51 0x7e5f4552091a69125d5dfcb7b8c2659029395bdf 0 0
3163 3157 13:30:51 0x1eff47bc3a10a45d4b230b5d10e37751fe6aa718 12856624 2
3164 3158 13:30:51 0x2b5ad5c4795c026514f8317c7a215e218dccd6cf 13835260 5
3165 3159 13:30:51 0x1eff47bc3a10a45d4b230b5d10e37751fe6aa718 13835260 5
####################### view 3167 missing
####################### view 3166 missing
####################### timestamp delta 6
3168 3160 13:30:57 0x1eff47bc3a10a45d4b230b5d10e37751fe6aa718 0 0
####################### view 3170 missing
####################### view 3169 missing
####################### timestamp delta 6
3171 3161 13:31:03 0x2b5ad5c4795c026514f8317c7a215e218dccd6cf 0 0
####################### view 3173 missing
####################### view 3172 missing
####################### timestamp delta 5
3174 3162 13:31:08 0x6813eb9362372eef6200f3b1dbc3f819671cba69 0 0
3175 3163 13:31:08 0x6813eb9362372eef6200f3b1dbc3f819671cba69 13835260 5
3176 3164 13:31:09 0x7e5f4552091a69125d5dfcb7b8c2659029395bdf 13835260 5
####################### view 3178 missing
####################### view 3177 missing
####################### timestamp delta 5
3179 3165 13:31:14 0x6813eb9362372eef6200f3b1dbc3f819671cba69 0 0
3180 3166 13:31:15 0x2b5ad5c4795c026514f8317c7a215e218dccd6cf 13835260 5
3181 3167 13:31:15 0x6813eb9362372eef6200f3b1dbc3f819671cba69 13835260 5
####################### view 3183 missing
####################### view 3182 missing
####################### timestamp delta 5
3184 3168 13:31:20 0x6813eb9362372eef6200f3b1dbc3f819671cba69 0 0
####################### view 3186 missing
####################### view 3185 missing
####################### timestamp delta 6
3187 3169 13:31:26 0x6813eb9362372eef6200f3b1dbc3f819671cba69 0 0
####################### view 3189 missing
####################### view 3188 missing
[...]

PR #1156, opened 2024-07-11
Too fast empty block creation

bzawisto:
This PR fixes (partially) - https://github.com/Zilliqa/zq2/issues/1142.

Having something in mempool doesn't mean this will be included in next block. Previously it led to a cases where delay was not respected for empty blocks.


zilliqa/src/pool.rs
```diff
@@ -256,6 +256,27 @@ impl TransactionPool {
...
         self.transactions.len()
     }
+
+    pub fn has_txn_for_next_block(&self) -> bool {
+        let mut ready = self.ready.clone();
+        while let Some(ReadyItem { tx_index, .. }) = ready.pop() {
+            // We don't include nonceless txns because the way we present results on API level requires having proper nonce
+            if let TxIndex::Intershard(_, _) = tx_index {
+                continue;
+            }
```
JamesHinshelwood, line 266: Are you sure this applies in this function? I would imagine that if there is an `Intershard` transaction in the pool, it would be ready for the next block and we should return true.

zilliqa/src/pool.rs
```diff
@@ -256,6 +256,27 @@ impl TransactionPool {
...
+
+            // A transaction might have been ready, but it might have gotten popped
+            // or the sender's nonce might have increased, making it invalid. In this case,
+            // we will have a stale reference would still exist in the heap.
+            //
+            let Some(_) = self.transactions.get(&tx_index) else {
+                continue;
+            };
+
+            return true;
```
JamesHinshelwood, line 276: Should we move this to a `fn pop_next_txn()` (or similar) since this logic and comment is repeated 3 times now?

zilliqa/src/pool.rs
```diff
@@ -256,6 +256,27 @@ impl TransactionPool {
     pub fn size(&self) -> usize {
         self.transactions.len()
     }
+
+    pub fn has_txn_for_next_block(&self) -> bool {
```
JamesHinshelwood, line 260: Maybe `has_txn_ready` would be better since it refers to the logic of the pool itself rather than blocks, which is something outside of this module. Maybe I'm nitpicking too much here though :)

zilliqa/src/consensus.rs
```diff
@@ -907,7 +907,7 @@ impl Consensus {
     }
 
     pub fn try_to_propose_new_block(&mut self) -> Result<Option<NetworkMessage>> {
-        if self.create_next_block_on_timeout {
+        if self.create_next_block_on_timeout && self.transaction_pool.has_txn_for_next_block() {
```
JamesHinshelwood, line 910: Some extra comments here might be useful. I got a bit confused working out what this function does and why it needed the extra condition.

zilliqa/src/pool.rs
DrZoltanFazekas: I agree, we should include intershard txns asap and not wait for other txns to be ready for the next block

JamesHinshelwood (approved):

Issue #1105, opened 2024-06-28
Correctly track block gas usage during block building


theo-zil:
Currently, block building happens in two steps:
 1. A transaction list is drained from the mempool, until the sum of the _gas limits_ for every transaction exceeds the block's gas limit.
 2. This list of transactions is executed; a second check - wholly redundant, as far as I can tell - ensures the actual gas usage is below the block gas limit (and returns an error if this fails)

This causes blocks to be under-filled, assuming the transaction gas limit is overestimated as a rule.

Here's what should happen instead:
 1. Transactions are drained one by one from the mempool and executed, until the actual gas usage exceeds the block's gas limit; the list of transactions before that point is the total transactions included in the block

For the last transaction, a potentially easy way to detect it would be to always pass `min(transaction.gas_limit, block_gas_remaining)` as the actual gas limit to the EVM for execution; then on out-of-gas errors, iff `block_gas_remaining < transaction.gas_limit` then re-insert that last transaction into the mempool (instead of discarding it as failed) and cease iteration.

PR #1157, opened 2024-07-11
Invalid view changes

bzawisto:
This PR fixes (partially) https://github.com/Zilliqa/zq2/issues/1142 and https://github.com/Zilliqa/zq2/issues/1105.

Txns are pulled one by one from mempool and included in the block based on their effective gas usage (not the max limit specified in txn).


zilliqa/src/pool.rs
```diff
@@ -124,11 +124,6 @@ impl TransactionPool {
                 continue;
             };
 
-            // If we've popped a nonced transaction, that may have made a subsequent one valid
-            if let Some(next) = tx_index.next().and_then(|idx| self.transactions.get(&idx)) {
-                self.ready.push(next.into());
-            }
-
```
JamesHinshelwood, line 131: @bzawisto Why did we remove this? I'm going to re-add it since I can't imagine this code working without it

JamesHinshelwood: Oh I see that you rely on the push in `mark_executed` instead. Will add a clarifying comment.

zilliqa/tests/it/main.rs
```diff
@@ -263,7 +263,7 @@ impl Network {
             consensus: ConsensusConfig {
                 genesis_deposits: genesis_deposits.clone(),
                 is_main: send_to_parent.is_none(),
-                consensus_timeout: Duration::from_secs(1),
+                consensus_timeout: Duration::from_secs(5),
```
JamesHinshelwood, line 266: @bzawisto Why these changes? Will leave them in but an explanation might be useful.

JamesHinshelwood (approved):

DrZoltanFazekas (approved):

PR #1158, opened 2024-07-11
 Optimise block proposal in consensus

joseph-zilliqa:
* Remove unnecessary conversions of NodePublicKey to and from bytes

* Add cache for getting block leaders


zilliqa/src/state.rs
```diff
@@ -28,6 +28,7 @@ use crate::{
 pub struct State {
     db: Arc<TrieStorage>,
     accounts: PatriciaTrie<TrieStorage>,
+    // pub stake_cache: RefCell<LruCache<Vec<u8>, NonZeroU128>>,
```
theo-zil, line 31: dev comment

zilliqa/src/exec.rs
```diff
@@ -678,9 +685,43 @@ impl State {
...
+            0,
+            BlockHeader::default(),
+        )?;
+
+        let stakers = contracts::deposit::GET_STAKERS
+            .decode_output(&stakers)
+            .unwrap()[0]
+            .clone()
+            .into_array()
+            .unwrap();
```
theo-zil, line 707: All of this seems duplicated with `get_stakers()`, and I think worth extracting into a helper function.

In fact, no, `get_stakers()` already calls `into_bytes()` inside its map, so I'd probably even have `get_stakers()` just call `get_stakers_bytes()` and re-iter the result constructing `NodePublicKey`s from the vec. This iterates one more time but I think that shouldn't be an issue since all performance-critical paths will probably use the `_bytes()` version anyway - and it should make the code much more concise.

zilliqa/src/consensus.rs
```diff
@@ -1903,37 +1953,80 @@ impl Consensus {
...
-        let public_key = *committee.get(index).unwrap();
+        let public_key = committee.get(index).unwrap();
 
         let peer_id = self
             .state
-            .get_peer_id(public_key)
+            .get_peer_id_from_public_key_bytes(public_key.clone())
             .unwrap()
             .context("Unable to get peer_id from staking contract!")
             .unwrap();
```
theo-zil, line 2027: This seems shared with `leader_bytes()` above. If I'm not missing any details, it should be possible to reduce `leader()` to something along the lines of
```rust
pub fn leader(&self, state: &State, view: u64) -> Validator {
    let (pub_key_bytes, peer_id) = self.leader_bytes(state, view);
    Validator {
        public_key: NodePublickey::from_bytes(pub_key).unwrap(),
        peer_id,
    }
}
```

zilliqa/src/consensus.rs
```diff
@@ -1903,37 +1953,80 @@ impl Consensus {
     }
 
     pub fn leader_at_block(&self, block: &Block, view: u64) -> Option<Validator> {
+        if let Some((cached_block_num, cached_view, leader)) = *self.leader_cache.borrow() {
+            if cached_block_num == block.number() && cached_view == view {
+                return Some(leader);
+            }
+        }
```
theo-zil, line 1960: Is it intentional that we don't update the cache here (and only inside `proposal()`)?
If so I'd consider renaming it to something like `latest_leader_cache`, since my first assumption upon seeing the field called `leader_cache` is that it would be used to store whatever the last accessed leader was, rather than just specifically the leader from the latest proposal.

theo-zil (commented):
A couple of minor comments, otherwise looks good - though I'm wondering whether an even more drastic refactor might be useful, where we define a new type along the lines of `NodePublicKeyUnchecked(pub Vec<u8>)`, only converting it to a BLS-based key in instances where we will need to sign/verify using it?
I think this would a) add some nominal type safety, where we're no longer passing around raw `Vec<u8>`s everywhere but rather wrap them in a named type; and b) might also reduce the amount of `_from_bytes` dual functions as most can simply operate on unchecked/raw keys and only a few select functions can take a checked/converted key.
Not entirely sure this is worthwhile, nor whether it can't be done in a subsequent PR anyway, but food for thought.

Also, as a matter of preference, I'd like a suffix such as `_raw` rather than `_from_bytes` as it's more concise, as a lot of names seem rather unwieldy. Compare `get_peer_id_from_public_key_bytes()` to `get_peer_id_from_raw()`. Though of course legibility comes first - I'm happy with the second but if there's disagreement then sticking with the first is fine.

joseph-zilliqa:
> A couple of minor comments, otherwise looks good - though I'm wondering whether an even more drastic refactor might be useful, where we define a new type along the lines of `NodePublicKeyUnchecked(pub Vec<u8>)`, only converting it to a BLS-based key in instances where we will need to sign/verify using it? I think this would a) add some nominal type safety, where we're no longer passing around raw `Vec<u8>`s everywhere but rather wrap them in a named type; and b) might also reduce the amount of `_from_bytes` dual functions as most can simply operate on unchecked/raw keys and only a few select functions can take a checked/converted key. Not entirely sure this is worthwhile, nor whether it can't be done in a subsequent PR anyway, but food for thought.
> 
> Also, as a matter of preference, I'd like a suffix such as `_raw` rather than `_from_bytes` as it's more concise, as a lot of names seem rather unwieldy. Compare `get_peer_id_from_public_key_bytes()` to `get_peer_id_from_raw()`. Though of course legibility comes first - I'm happy with the second but if there's disagreement then sticking with the first is fine.

I have added a container struct for `Vec<u8>` called `NodePublicKeyRaw` as you suggested, and renamed functions with the suffix `_from_bytes` and similar to `_raw`.

As there are many places that use the existing non raw functions I would suggest that opening another PR containing that refactoring would be better, however if you think that should be included in this PR I am happy to do that.

theo-zil (dismissed):
Nice, LGTM!
One minor comment but not a blocker, feel free to change or dismiss as you like.

theo-zil (approved):

Issue #943, opened 2024-05-18
zq2 does not return an error from GetBalance if the account does not exist


rrw-zilliqa:
```
curl -d '{
        "id": "1",
        "jsonrpc": "2.0",
        "method": "GetBalance",
        "params": ["2eefc4f453539e5ee732b49eb4792b268c2f3908"]
    }' -H "Content-Type: application/json" -X POST "https://api.zq2-devnet.zilliqa.com/"

{"jsonrpc":"2.0","result":{"balance":"0","nonce":0},"id":"1"}
```

vs

```
url -d '{
        "id": "1",
        "jsonrpc": "2.0",
        "method": "GetBalance",
        "params": ["2eefc4f453539e5ee732b49eb4792b268c2f3908"]
    }' -H "Content-Type: application/json" -X POST "https://api.devnet.zilliqa.com/"
{"error":{"code":-5,"data":null,"message":"Account is not created"},"id":"1","jsonrpc":"2.0"}
```

PR #1159, opened 2024-07-11
Fix GetBalance API call to return error if account does not exist

joseph-zilliqa:
* Fixes https://github.com/Zilliqa/zq2/issues/943

### DIFF
--- a/zilliqa/src/state.rs
+++ b/zilliqa/src/state.rs
@@ -179,6 +179,19 @@ impl State {
     /// Fetch an Account struct.
     /// Note: use get_account_storage to obtain a specific storage value.
     /// If modifying a raw account, ensure you call save_account afterwards.
+    /// Returns an error on failures to access the state tree, or decode the account; or none
+    /// if the account doesn't exist yet
+    pub fn get_account(&self, address: Address) -> Result<Option<Account>> {
+        Ok(self
+            .accounts
+            .get(&Self::account_key(address).0)?
+            .map(|bytes| bincode::deserialize::<Account>(&bytes))
+            .transpose()?)
+    }
+
+    /// Fetch an Account struct.
+    /// Note: use get_account_storage to obtain a specific storage value.
+    /// If modifying a raw account, ensure you call save_account afterwards.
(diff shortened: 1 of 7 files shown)


zilliqa/src/state.rs
```diff
@@ -176,12 +176,25 @@ impl State {
...
+            .transpose()?)
+    }
+
     /// Fetch an Account struct.
     /// Note: use get_account_storage to obtain a specific storage value.
     /// If modifying a raw account, ensure you call save_account afterwards.
     /// Returns an error on failures to access the state tree, or decode the account; or an empty
     /// account if one didn't exist yet
-    pub fn get_account(&self, address: Address) -> Result<Account> {
+    pub fn get_account_or_default(&self, address: Address) -> Result<Account> {
```
JamesHinshelwood, line 197: Its slightly a question of style, but I feel fairly strongly that the 'normal' and 'properly named' way to get an account should be the one which returns `Result<Account>`. i.e. I think `get_account` should return `Result<Account>`. This is because based on the semantics of the EVM, which we adopt, there is no such thing as an account existing or not - All `2^160` possible accounts always exist. It is simply an implementation-detail that 99% of them are the default account and thus there's no need to store them.

So, I'm apprehensive about exposing 'existance' if we can avoid it.

However of course the motivation for this issue is that ZQ1 _does_ expose that implementation detail ( :cry: ), but I would like just the `GetBalance` implementation to use a special way and everything else to stick with our pure non-optional account model.

FWIW, the execution model also needs to slightly break this abstraction, since it charges more gas for sending to non-existent accounts (because obviously it results in us having to store them). This works via the `State::has_account` method. So I wonder if an alternative to this PR might look like:

* Restore `get_account(...) -> Result<Account>` and use it everywhere.
* Add a call to `.has_account(...) -> Result<bool>` in the `GetBalance` implementation and early-return the error if it is false.

What do you think?

JamesHinshelwood (changes requested):

JamesHinshelwood (approved):

Issue #902, opened 2024-05-09
Calling `eth_getTransactionCount` with a block number of `pending` should include the number of transactions in the mempool


JamesHinshelwood:
This is how other implementations work and it avoids clients having to track the nonce of pending transactions themselves.

See #861 and #901.

PR #1172, opened 2024-07-15
Change eth_getTransactionCount to respect the pending block number 

joseph-zilliqa:
* Fixes https://github.com/Zilliqa/zq2/issues/902

* Add test for `eth_getTransactionCount` to verify that it respects the expected behavior when a block number of `pending` is passed

### DIFF
--- a/zilliqa/src/api/eth.rs
+++ b/zilliqa/src/api/eth.rs
@@ -292,6 +292,11 @@ fn get_transaction_count(params: Params, node: &Arc<Mutex<Node>>) -> Result<Stri
     expect_end_of_params(&mut params, 3, 3)?;
 
     let node = node.lock().unwrap();
+
+    if matches!(block_id, BlockId::Number(BlockNumberOrTag::Pending)) {
+        return Ok(node.txpool_content().pending.len().to_hex());
+    }
+
     let block = node.get_block(block_id)?;
     let block = build_errored_response_for_missing_block(block_id, block)?;
 
--- a/zilliqa/src/api/eth.rs
+++ b/zilliqa/src/api/eth.rs
@@ -299,10 +299,7 @@ fn get_transaction_count(params: Params, node: &Arc<Mutex<Node>>) -> Result<Stri
     let nonce = node.get_state(&block)?.get_account(address)?.nonce;
 
     if matches!(block_id, BlockId::Number(BlockNumberOrTag::Pending)) {
-        let pending_transaction_count =
-            node.pending_transaction_account_for_address(address) as u64;
-
-        Ok((nonce + pending_transaction_count).to_hex())
+        Ok(node.transaction_count_for_account(address).to_hex())
     } else {
         Ok(nonce.to_hex())
     }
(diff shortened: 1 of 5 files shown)


zilliqa/src/api/eth.rs
```diff
@@ -292,10 +292,24 @@ fn get_transaction_count(params: Params, node: &Arc<Mutex<Node>>) -> Result<Stri
...
     let block = node.get_block(block_id)?;
     let block = build_errored_response_for_missing_block(block_id, block)?;
 
-    Ok(node.get_state(&block)?.get_account(address)?.nonce.to_hex())
+    let nonce = node.get_state(&block)?.get_account(address)?.nonce;
+
+    if matches!(block_id, BlockId::Number(BlockNumberOrTag::Pending)) {
+        let pending_transaction_count = node
+            .txpool_content()
+            .pending
```
JamesHinshelwood, line 304: Annoyingly I think we need to be slightly smarter with this algorithm. If the pool contains many transactions from a signer, but they aren't contiguous with the signer's current nonce, then we shouldn't increment the nonce.

i.e.
account.nonce = 0
pool = [{ nonce = 1 }, { nonce = 2 }]
The current implementation would return `transaction_count = 2` when we actually want `transaction_count = 0`

zilliqa/src/api/eth.rs
```diff
@@ -292,10 +292,24 @@ fn get_transaction_count(params: Params, node: &Arc<Mutex<Node>>) -> Result<Stri
...
+
     let block = node.get_block(block_id)?;
     let block = build_errored_response_for_missing_block(block_id, block)?;
 
-    Ok(node.get_state(&block)?.get_account(address)?.nonce.to_hex())
+    let nonce = node.get_state(&block)?.get_account(address)?.nonce;
+
+    if matches!(block_id, BlockId::Number(BlockNumberOrTag::Pending)) {
+        let pending_transaction_count = node
+            .txpool_content()
```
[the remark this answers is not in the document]
joseph-zilliqa: I don't believe it would be very beneficial as there would be a load of duplicate code, with almost the same effect and some computation needs to happen in the `Consensus` anyway to ensure nonces are not too low. I have proposed a change moving the computation to `Consensus`. If you think that it should still be in `TransactionPool`, let me know and I will move all I can there.

zilliqa/src/node.rs
```diff
@@ -777,6 +777,10 @@ impl Node {
         self.consensus.txpool_content()
     }
 
+    pub fn transaction_count_for_account(&self, account: Address) -> u64 {
+        self.consensus.transaction_count_for_account(account)
```
JamesHinshelwood, line 781: Feel free to just make `transaction_pool` `pub` and use it directly - We do the same for other fields in `Consensus`.

zilliqa/src/consensus.rs
```diff
@@ -739,6 +739,13 @@ impl Consensus {
         content
     }
 
+    pub fn transaction_count_for_account(&self, account: Address) -> u64 {
+        let current_nonce = self.state.must_get_account(account).nonce;
+
+        self.transaction_pool
+            .transaction_count(account, current_nonce)
```
JamesHinshelwood, line 746: Bit of a nit, but I'd prefer if all these methods were called `pending_transaction_count`. If I see the name `transaction_count` my first guess would be that they returned the number of mined transactions from this account, which is specifically what these implementations don't do.

JamesHinshelwood (dismissed):
LGTM, just a couple of nits which you can ignore if you disagree with me :) Thanks!

JamesHinshelwood (approved):

Issue #993, opened 2024-05-29
Limit response size of `GetSmartContract(Sub)State`


JamesHinshelwood:
Querying the state of large contracts on ZQ1 causes performance problems/crashes. Some contracts have a lot of state (~50 MiB).

Add configuration to limit the size of the response from these APIs. Default it to on at a sensible level (1 MiB?).

rrw-zilliqa:
.. and remember to provide instructions on how to configure your zq2 node to disable the restriction :-)
we also probably want to implement functionality to allow you to work out what state should be queried by GetContractSubState() otherwise block explorers will need to understand Scilla to show you state, which might be more than I want to implement :-p

JamesHinshelwood:
Consider this blocked by #919 for now since that change will fundamentally affect how Scilla contract storage is stored and thus how we can account for its size.

PR #1174, opened 2024-07-15
Add a limit to GetSmartContractState return payload size

joseph-zilliqa:
* Fixes https://github.com/Zilliqa/zq2/issues/993

* Add a configurable limit to the payload size that can be returned by GetSmartContractState

* Set a default limit of 1024 KiB (1 MiB)


zilliqa/src/api/zil.rs
```diff
@@ -299,19 +299,60 @@ fn get_smart_contract_state(params: Params, node: &Arc<Mutex<Node>>) -> Result<V
...
+        let mut total_size = 0;
+
         let trie = state.get_account_trie(address)?;
         for (k, v) in trie.iter() {
-            let (var_name, indices) = split_storage_key(&k)?;
-            let mut var = result.entry(var_name);
+            let (mut var_name, indices) = split_storage_key(&k)?;
+            let mut var = result.entry(var_name.clone());
+
+            // Length of a key once encoded in JSON, quotes and escaped characters included
```
JamesHinshelwood, line 318: Given the complexity of calculating the length as we go, I wonder if it would be more sensible to instead limit the response size by total number of keys, rather than the serialised size? Maybe there are edge cases where values could be very large, but I'd imagine that's rare and we could always solve that with a per-value byte limit instead (which is much easier to check).

Another thought is that once we enable this limit, we will probably want to give clients a way of paginating through a contract's state which is larger than the limit. It seems more immediately natural that we would enable this pagination by letting them passing the number of keys to skip, which corresponds with a response size limited by number of entries too.

Happy to discuss though if you think otherwise? (and sorry, I think I am now contradicting an earlier opinion of mine :stuck_out_tongue: )

JamesHinshelwood (approved):

Issue #1171, opened 2024-07-15
`eth_getLogs` error on unknown blocks


DrZoltanFazekas:
When I call `eth_getLogs` with a future block range, I get:
```
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0", "id": 1, "method":"eth_getLogs","params": [{"address": "0xbA44BC29371E19117DA666B729A1c6e1b35DDb40", "fromBlock": "0xfffffffff", "toBlock": "0xfffffffff"}]}' https://api.zq2-prototestnet.zilliqa.com

{"jsonrpc":"2.0","error":{"code":-32603,"message":"Unhandled panic in RPC handler eth_getLogs"},"id":1}
```

The correct behavior (based on other EVM chains) is to return an empty set:
```
{"jsonrpc": "2.0", "id": 1, "result": []} 
```

PR #1176, opened 2024-07-16
Fix panic in eth_getLogs when called with future blocks

joseph-zilliqa:
* Fixes https://github.com/Zilliqa/zq2/issues/1171

* Return an empty list when eth_getLogs is called with future blocks, matching behaviour of other EVM based chains

### DIFF
--- a/zilliqa/src/api/eth.rs
+++ b/zilliqa/src/api/eth.rs
@@ -423,14 +423,21 @@ fn get_logs(params: Params, node: &Arc<Mutex<Node>>) -> Result<Vec<eth::Log>> {
             .get_block(block_hash)?
             .ok_or_else(|| anyhow!("block not found"))?))),
         (None, from, to) => {
-            let from = node
+            let Some(from) = node
                 .resolve_block_number(from.unwrap_or(BlockNumberOrTag::Latest))?
-                .unwrap()
-                .number();
-            let to = node
+                .as_ref()
+                .map(Block::number)
+            else {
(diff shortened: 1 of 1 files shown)


zilliqa/src/api/eth.rs
```diff
@@ -423,14 +423,21 @@ fn get_logs(params: Params, node: &Arc<Mutex<Node>>) -> Result<Vec<eth::Log>> {
...
+
+            let Some(to) = node
                 .resolve_block_number(to.unwrap_or(BlockNumberOrTag::Latest))?
-                .unwrap()
-                .number();
+                .as_ref()
+                .map(Block::number)
+            else {
+                return Ok(vec![]);
+            };
```
JamesHinshelwood, line 440: I'm not actually sure what the intended behaviour is (maybe we should check on an Ethereum testnet), but this will have the affect that if `to` is too large, I will always get an empty response which is not what I would immediately assume. I would hope instead that the block range is instead truncated to a valid range.

i.e. if the network was at block height `N` and I made a request for logs between `N - 10` and `N + 10`, I would assume I would receive the logs between `N - 10` and `N` rather than nothing.

Best to check another implementation or API I think?

joseph-zilliqa: Having just tested this I can confirm that what you suggested is correct, and have pushed changes to implement this behaviour.

JamesHinshelwood (approved):

PR #1182, opened 2024-07-18
Change storage of timestamps within the database to be a serialisation of Rust's Duration

joseph-zilliqa:
* Change storage of timestamps within the DB to use a serialisation of Rust's `std::time::Duration` since the unix epoch

* Change type of timestamps within the db to be `BLOB`


JamesHinshelwood (approved):
LGTM, thanks :)

PR #1185, opened 2024-07-19
Fix second location of apply_rewards happening after transaction execution

### LINKED ISSUE #1143: Nodes panic on removal of validator
We saw nodes panicking because they tried to unwrap `None` in https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/consensus.rs#L677
to apply rewards of a staker whom we removed from the validator set through a transaction calling `deposit::tempRemoveStaker()` in the current block. 

https://github.com/Zilliqa/zq2/pull/1120 changes the way we determine the leader in `apply_rewards()` . It was based on the previous block and not the current block's state before: https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/consensus.rs#L632-L634

With this change we would pick a different leader for the current block in case the actual block proposer were removed from the validator set in the current block. I think this is wrong since the actual proposer was selected from the validator set based on the previous block's state before the current block was proposed. The fact that the current block's proposer is removed from the validator set in the current block should only have an impact on the leader selection of the subsequent blocks. 

Related to that, there is another problem because we're trying to get the reward address of the leader in https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/consensus.rs#L640
and the reward addresses of the cosigners in https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/consensus.rs#L652 (which was the problem we ran into)
based on the current block's state, but their `pub_key` does not exist anymore in the deposit contract's current state since our transaction executed in the current block removed it.

As discussed, we would need to fetch the reward addresses based on the previous block's state in which the deposit contract had the respective `pub_key` before they were removed in the current block.

theo-zil:
Should now be the first state change that happens in every block. Fixes #1143.
Also adds a test for it.

### DIFF
--- a/zilliqa/src/crypto.rs
+++ b/zilliqa/src/crypto.rs
@@ -219,11 +219,11 @@ impl SecretKey {
         Self::from_bytes(&bytes_vec)
     }
 
-    fn as_bls(&self) -> bls_signatures::PrivateKey {
+    pub fn as_bls(&self) -> bls_signatures::PrivateKey {
         bls_signatures::PrivateKey::new(self.bytes)
     }
 
-    fn as_ecdsa(&self) -> k256::ecdsa::SigningKey {
+    pub fn as_ecdsa(&self) -> k256::ecdsa::SigningKey {
         // `SigningKey::from_bytes` can fail for two reasons:
         // 1. The bytes represent a zero integer. However, we validate this is not the case on construction.
         // 2. The bytes represent an integer less than the curve's modulus. However for ECDSA, the curve's order is
--- a/zilliqa/tests/it/main.rs
+++ b/zilliqa/tests/it/main.rs
@@ -2,6 +2,7 @@ use alloy_primitives::Address;
 use ethers::{
     abi::Tokenize,
     providers::{Middleware, PubsubClient},
+    types::TransactionRequest,
 };
 use primitive_types::U256;
 use serde_json::{value::RawValue, Value};
@@ -253,7 +254,7 @@ impl Network {
                     k.node_public_key(),
                     k.to_libp2p_keypair().public().to_peer_id(),
                     stake.into(),
-                    Address::random_with(rng.lock().unwrap().deref_mut()),
+                    Address::from_private_key(&k.as_ecdsa()),
                 )
             })
             .collect();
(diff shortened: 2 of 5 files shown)


zilliqa/src/crypto.rs
```diff
@@ -219,11 +219,11 @@ impl SecretKey {
         Self::from_bytes(&bytes_vec)
     }
 
-    fn as_bls(&self) -> bls_signatures::PrivateKey {
+    pub fn as_bls(&self) -> bls_signatures::PrivateKey {
```
theo-zil, line 222: Minor internal interface change - but I feel like if we have access to a `SecretKey` field in some scope, there's no real reason not to have these conversion methods public for it.

DrZoltanFazekas (approved):

zilliqa/src/crypto.rs
```diff
@@ -219,11 +219,11 @@ impl SecretKey {
...
         bls_signatures::PrivateKey::new(self.bytes)
     }
 
-    fn as_ecdsa(&self) -> k256::ecdsa::SigningKey {
+    pub fn as_ecdsa(&self) -> k256::ecdsa::SigningKey {
```
JamesHinshelwood, line 226: I was of the view that this method didn't really make sense. A validator's secret key isn't tied to any specific ECDSA key or EVM address that is meaningful.

zilliqa/tests/it/main.rs
```diff
@@ -253,7 +254,7 @@ impl Network {
                     k.node_public_key(),
                     k.to_libp2p_keypair().public().to_peer_id(),
                     stake.into(),
-                    Address::random_with(rng.lock().unwrap().deref_mut()),
+                    Address::from_private_key(&k.as_ecdsa()),
```
JamesHinshelwood, line 257: Why this change? I guess this relates to my previous point.

JamesHinshelwood: Oh I see - Because we need to send the removal transaction from the `rewardAddress` account. I'd slightly prefer for us to just generate a distinct ECDSA key for each `TestNode` and expose a `Wallet` from there, since it keeps the two keys distinct.

But I won't block if you think this is okay :)

JamesHinshelwood (approved):

zilliqa/tests/it/main.rs
theo-zil: That's very fair - I think I'll do it this way then. In the docker compose, the default config re-uses the private keys for the reward addresses so I kind of operated on the assumption that each node has a "natural" address, but I'm perfectly happy to decouple them.

However I think I'll do it in a separate PR, alongside cleaning up the ECDSA methods for SecretKey then.

PR #1187, opened 2024-07-22
Move Consensus::leader into the deposit contract

joseph-zilliqa:
* Fixes https://github.com/Zilliqa/zq2/issues/928

* Move implementation of `Consensus::leader` into the deposit contract

* Remove unnecessary `view` parameter from `Consensus::leader_at_block`

* Add implementation of Solidity's `prevrandao` as the Keccak256 hash of the current block's view


zilliqa/src/contracts/deposit.sol
```diff
@@ -10,6 +10,11 @@ struct Staker {
     bytes peerId;
 }
 
+struct Validator {
+    bytes publicKey;
+    bytes peerId;
+}
+
```
DrZoltanFazekas, line 17: I don't think we need the `Validator` struct, please remove it if not used

zilliqa/src/consensus.rs
```diff
@@ -1070,7 +1065,7 @@ impl Consensus {
 
     fn leader_for_view(&mut self, parent_hash: Hash, view: u64) -> Option<NodePublicKey> {
         if let Ok(Some(parent)) = self.get_block(&parent_hash) {
-            let leader = self.leader_at_block(&parent, view).unwrap();
+            let leader = self.leader_at_block(&parent).unwrap();
```
DrZoltanFazekas, line 1068: Shouldn't we rather use the `leaderAtView()` function of the `Deposit` contract and pass in the `view` argument? The `view` parameter of the `leader_for_view()` function could refer to an arbitrary view number, not necessarily the current view.

zilliqa/src/exec.rs
```diff
@@ -644,6 +647,26 @@ impl State {
         Ok(())
     }
 
+    pub fn leader_at_block(&self, block: &Block) -> Result<NodePublicKey> {
+        let data = contracts::deposit::LEADER.encode_input(&[])?;
```
DrZoltanFazekas, line 651: After looking at `Consensus::leader_at_block()` I trust this returns the leader based on the state of the `block` argument. I'm just wondering why `get_stakers_at_block()` uses a different way to return the stakers based on a given block's state.

zilliqa/src/consensus.rs
```diff
@@ -631,7 +626,7 @@ impl Consensus {
         let rewards_per_block =
             *self.config.consensus.rewards_per_hour / self.config.consensus.blocks_per_hour as u128;
 
-        let proposer = self.leader_at_block(parent_block, view).unwrap().public_key;
+        let proposer = self.leader_at_block(parent_block).unwrap().public_key;
```
DrZoltanFazekas, line 629: Shouldn't we rather use the `leaderAtView()` function of the `Deposit` contract and pass in the `view` argument? The `view` parameter of the `apply_rewards()` function could refer to an arbitrary view number, not necessarily the current view.

zilliqa/src/consensus.rs
```diff
@@ -1081,7 +1076,7 @@ impl Consensus {
                 return None;
             }
             let head_block = self.head_block();
-            let leader = self.leader_at_block(&head_block, view).unwrap();
+            let leader = self.leader_at_block(&head_block).unwrap();
```
DrZoltanFazekas, line 1079: Shouldn't we rather use the `leaderAtView()` function of the `Deposit` contract and pass in the `view` argument? The `view` parameter of the `leader_for_view()` function could refer to an arbitrary view number, not necessarily the current view.

zilliqa/src/consensus.rs
```diff
@@ -1582,7 +1577,7 @@ impl Consensus {
         }
 
         // Derive the proposer from the block's view
-        let proposer = self.leader_at_block(&parent, block.view()).unwrap();
+        let proposer = self.leader_at_block(&parent).unwrap();
```
DrZoltanFazekas, line 1580: We should use the `leaderAtView()` function of the `Deposit` contract and pass in the `view` argument, otherwise if we execute `leader_at_block()` on the `parent` the view number that determines the randomness in the `Deposit::leader()` function will be the `parent`'s view and not `block.view()`.

zilliqa/src/consensus.rs
```diff
@@ -1921,40 +1916,18 @@ impl Consensus {
         Ok(())
     }
 
-    pub fn leader_at_block(&self, block: &Block, view: u64) -> Option<Validator> {
+    pub fn leader_at_block(&self, block: &Block) -> Option<Validator> {
```
DrZoltanFazekas, line 1919: As suggested above, we should use the `leaderAtView()` function of the `Deposit` contract with the `view` parameter.

zilliqa/src/consensus.rs
```diff
@@ -1921,40 +1916,18 @@ impl Consensus {
...
         let Ok(state_at) = self.try_get_state_at(block.number()) else {
             return None;
         };
-        Some(self.leader(&state_at, view))
-    }
 
-    pub fn leader(&self, state: &State, view: u64) -> Validator {
-        let committee = state.get_stakers().unwrap();
+        let public_key = state_at.leader_at_block(block).unwrap();
```
DrZoltanFazekas, line 1924: Here we should pass in the `view` instead of the `block`, which is not used in `State::leader_at_block()`.

zilliqa/src/node.rs
```diff
@@ -684,11 +684,7 @@ impl Node {
         let parent = self
             .get_block(header.parent_hash)?
             .ok_or_else(|| anyhow!("missing parent: {}", header.parent_hash))?;
-        let proposer = self
-            .consensus
-            .leader_at_block(&parent, header.view)
-            .unwrap()
-            .public_key;
+        let proposer = self.consensus.leader_at_block(&parent).unwrap().public_key;
```
DrZoltanFazekas, line 687: We should use the `leaderAtView()` function of the `Deposit` contract and pass in the `view` argument, otherwise if we execute `leader_at_block()` on the `parent` the view number that determines the randomness in the `Deposit::leader()` function will be the `parent`'s view and not `header.view`.

DrZoltanFazekas (commented):
Most of my comments suggest calling `Deposit::leaderAtView()` with the `view` number that was used in the original Rust code instead of calling `Deposit::leader()` that uses the view number of the respective block to generate the randomness for the leader selection, which is in many cases incorrect.

DrZoltanFazekas (approved):
My comments from yesterday were addressed and I tested the branch on localdev

PR #1203, opened 2024-07-22
Add support for pending blocks

bzawisto:
Fixes: #863.

This is now implemented by cloning and using associated functions (with `_at` suffix).

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -20,10 +20,8 @@ use tracing::*;
 use crate::{
     block_store::BlockStore,
     blockhooks,
-    cfg::NodeConfig,
-    crypto::{verify_messages, Hash, NodePublicKey, NodePublicKeyRaw, NodeSignature, SecretKey},
     cfg::{ConsensusConfig, NodeConfig},
-    crypto::{verify_messages, Hash, NodePublicKey, NodeSignature, SecretKey},
+    crypto::{verify_messages, Hash, NodePublicKey, NodePublicKeyRaw, NodeSignature, SecretKey},
     db::Db,
     exec::TransactionApplyResult,
     inspector::{self, ScillaInspector, TouchedAddressInspector},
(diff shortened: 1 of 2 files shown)


zilliqa/src/consensus.rs
```diff
@@ -1056,11 +1109,61 @@ impl Consensus {
...
         }
         Ok(Some((proposal, broadcasted_transactions)))
     }
 
+    fn propose_new_block(&mut self) -> Result<Option<(Block, Vec<VerifiedTransaction>)>> {
+        let mut state = self.state.clone();
+        let mut tx_pool = self.transaction_pool.clone();
+        let mut votes = self.votes.clone();
+
+        let result = self.propose_new_block_at(&mut state, &mut tx_pool, &mut votes);
```
theo-zil, line 1122: I'm not super keen on cloning the pool and the votes on every proposal (cloning the state is cheap) - but the `&self` immutable borrow does make it difficult to just pass `&mut self.state, &mut self.transaction_pool, &mut self.votes` directly. Some of the usages of `&self` inside `propose_new_block()` can be eliminated, but it'd require 3-4 extra paremeters which gets ugly, so I'm not sure I have a better solution to suggest.

In a perfect world we would have a way to peek in the transaction pool immutably, similarly to how we can edit state and rollback painlessly, thus removing the need for `_at` in general (and mutating the votes is fine because it's just adding a default value if there isn't one). But I suppose this is a fine alternative for now, without having to redesign the transaction pool completely to allow that.

zilliqa/src/consensus.rs
```diff
@@ -1056,11 +1109,61 @@ impl Consensus {
...
+            let votes = (
+                vec![my_vote.signature()],
+                bitvec![u8, bitvec::order::Msb0; 1; 1],
+                0,
+                false,
+            );
+            v.insert(votes);
+        }
+
+        let result = self.propose_new_block_at(&mut state, &mut tx_pool, &mut votes)?;
```
theo-zil, line 1159: Haha, this is an `Option` so having it be named `result` made me do a double-take. Maybe naming it `proposal` will make it slightly more readable-at-first-glance?

zilliqa/src/consensus.rs
```diff
@@ -1056,11 +1109,61 @@ impl Consensus {
...
+        let result = self.propose_new_block_at(&mut state, &mut tx_pool, &mut votes);
+
+        self.state = state;
+        self.transaction_pool = tx_pool;
+        self.votes = votes;
+
+        result
+    }
+
+    pub fn propose_pending_block(&self) -> Result<Option<Block>> {
```
theo-zil, line 1131: Also very minor nit but maybe this could be named something like `get_pending_block()`? It doesn't even return a proper `Proposal` since it doesn't return the transactions. And it's a bit odd seeing the eth API make a call to a method called "propose a block"

theo-zil (changes requested):
One actual CR in `apply_rewards`, a few naming nits and a small discussion on structuring `propose_new_block` - if no better ideas crop up there (cc @JamesHinshelwood if you have any thoughts to give), then looks good to me

theo-zil (approved):
I think I'm actually happy to approve for now, and we can refactor this down the line if need be. I can't see a better way to do this without a bunch of work on the transaction pool.

Issue #929, opened 2024-05-16
Implement EVM filter APIs


JamesHinshelwood:

maxconway:
I've broken this down into separate issues:
- #2383 
- #2384
- #2385 
- #2386 
- #2387 

But I'll keep this issue around and close them all at once

PR #1210, opened 2024-07-24
Implement EVM filter APIs

joseph-zilliqa:
* Fixes https://github.com/Zilliqa/zq2/issues/929

* Implement EVM filter RPCs

* Add configuration options and defaults for time since filter was last used before it is automatically removed and maximum number of filters allowed to be registered at the same time

* Add tests for EVM filter APIs


zilliqa/src/api/types/eth.rs
```diff
@@ -383,27 +383,27 @@ fn m3_2048(bloom: &mut [u8; 256], data: &[u8]) {
     }
 }
 
-/// A type for representing null, a single item or an array of items.
-#[derive(Deserialize, Serialize)]
+/// A type for representing a single item or an array of items.
+#[derive(Debug, Deserialize, Serialize, Clone)]
 #[serde(untagged)]
 pub enum OneOrMany<T> {
-    Null,
```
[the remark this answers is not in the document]
JamesHinshelwood: I see we are using `Option<OneOrMany>` instead but if we're using that everywhere why not just use the `OneOrMany` with `Null`?

zilliqa/src/cfg.rs
```diff
@@ -70,6 +70,12 @@ pub struct NodeConfig {
     /// The maximum number of blocks to request in a single message when syncing.
     #[serde(default = "block_request_batch_size_default")]
     pub block_request_batch_size: u64,
+    /// Maximum amount of time since a filter was last used before it is automatically removed.
+    #[serde(default = "filter_expiry_default")]
+    pub filter_expiry: Duration,
+    /// Maximum amount of filters registered at once. Set to 0 for no limit.
+    #[serde(default = "max_filters_default")]
+    pub max_filters: u64,
```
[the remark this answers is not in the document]
JamesHinshelwood: More generally - Why do we have this limit at all? We don't have similar limits on other APIs (yet).

zilliqa/src/api/eth.rs
```diff
@@ -816,6 +766,377 @@ fn get_uncle(_: Params, _: &Arc<Mutex<Node>>) -> Result<Option<String>> {
     Ok(None)
 }
 
+#[derive(Deserialize, Default)]
+#[serde(default, rename_all = "camelCase")]
+struct NewFilterParams {
```
JamesHinshelwood, line 771: Should we consider using https://docs.rs/alloy/latest/alloy/rpc/types/eth/struct.Filter.html instead?

I also node that it supports a block hash as a parameter which I guess is part of the spec?

zilliqa/src/api/eth.rs
```diff
@@ -816,6 +766,377 @@ fn get_uncle(_: Params, _: &Arc<Mutex<Node>>) -> Result<Option<String>> {
...
+    // let first_block = node.resolve_block_number(from_block)?.unwrap().number();
+
+    let first_block = match params.from_block.unwrap_or(BlockNumberOrTag::Latest) {
+        BlockNumberOrTag::Number(n) => n,
+        tag => node.resolve_block_number(tag)?.unwrap().number(),
+    };
+
+    let mut rng = rand::thread_rng();
+
+    let mut id = [0u8; 32];
```
JamesHinshelwood, line 807: Do filter IDs really need to be 32 bytes long? What about just `rand::random::<u64>()`?

zilliqa/src/api/eth.rs
```diff
@@ -1,6 +1,9 @@
 //! The Ethereum API, as documented at <https://ethereum.org/en/developers/docs/apis/json-rpc>.
 
-use std::sync::{Arc, Mutex, MutexGuard};
+use std::{
+    sync::{Arc, Mutex, MutexGuard},
+    time::Instant,
```
JamesHinshelwood, line 5: Can we use `zilliqa::time::SystemTime` instead since we can control it in tests? (and this would let you write a test for filter expiration)

zilliqa/src/api/eth.rs
```diff
@@ -816,6 +766,377 @@ fn get_uncle(_: Params, _: &Arc<Mutex<Node>>) -> Result<Option<String>> {
...
+        + 1;
+
+    let mut rng = rand::thread_rng();
+
+    let mut id = [0u8; 32];
+    rng.fill(&mut id);
+
+    let id = B256::from_slice(&id);
+
+    node.filters.put(
```
JamesHinshelwood, line 889: Lots of the logic is duplicated between the 3 filter methods. Is it doable to de-duplicate at all?

zilliqa/src/api/eth.rs
```diff
@@ -816,6 +766,377 @@ fn get_uncle(_: Params, _: &Arc<Mutex<Node>>) -> Result<Option<String>> {
...
+            txn_hash,
+            block_number,
+            block_hash,
+        ))
+    });
+
+    logs.collect()
+}
+
+fn get_filter_changes(params: Params, node: &Arc<Mutex<Node>>) -> Result<serde_json::Value> {
```
JamesHinshelwood, line 967: Could consider returning https://docs.rs/alloy/latest/alloy/rpc/types/enum.FilterChanges.html to avoid having to convert to JSON here?

JamesHinshelwood (commented):
Sorry for the slow review and the resulting conflicts. I've made a few comments that we can discuss.

zilliqa/src/api/eth.rs
joseph-zilliqa: The block hash isn't mention in the spec (https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_newfilter) and that filter struct is used for a different api endpoint (https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getfilterlogs) that does include block hash as a parameter

zilliqa/src/api/eth.rs
joseph-zilliqa: This was what other Ethereum endpoints seemed to use for their filter ids

zilliqa/src/api/eth.rs
joseph-zilliqa: But it might just be better to use that struct anyway

zilliqa/src/cfg.rs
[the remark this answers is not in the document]
joseph-zilliqa: I was going along the lines of preventing DOS attacks by registering many filters, but I can remove the limit if it is unfounded

zilliqa/src/api/eth.rs
joseph-zilliqa: Having looked at it again I don't think it makes sense to have the block hash parameter, as that block would always be in the past and therefore this endpoint would not be suitable. Please correct me if I am wrong.

JamesHinshelwood:
Closing this for now, will re-create once we have time to work on it.

PR #1214, opened 2024-07-25
Remove ECDSA methods from SecretKey.

theo-zil:
The key is intended to be a signing key for the physical node - and not tied to any on-chain account. In the test framework and in z2, a separate key is now generated for each validator's on-chain account.

### DIFF
--- a/infra/tf/locals.tf
+++ b/infra/tf/locals.tf
@@ -5,7 +5,7 @@
 locals {
   binary_location      = "${path.module}/../../target/x86_64-unknown-linux-gnu/release/zilliqa"
   labels               = merge(var.labels, { "zq2-network" = var.network_name })
-  bootstrap_public_key = data.external.bootstrap_key_converted.result.bsl_public_key
+  bootstrap_public_key = data.external.bootstrap_key_converted.result.bls_public_key
   bootstrap_peer_id    = data.external.bootstrap_key_converted.result.peer_id
   genesis_address      = data.external.genesis_key_converted.result.address
   network_name         = element(split("/", data.google_compute_subnetwork.default.network), length(split("/", data.google_compute_subnetwork.default.network)) - 1)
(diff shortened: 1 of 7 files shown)


zilliqa/src/bin/convert-key.rs
```diff
@@ -18,10 +19,17 @@ fn main() -> Result<()> {
...
 
+    let key_bytes = secret_key.as_bytes();
+    let ecdsa_key =
+        k256::ecdsa::SigningKey::from_bytes(GenericArray::from_slice(&key_bytes)).unwrap();
+
+    let tx_pubkey = TransactionPublicKey::Ecdsa(k256::ecdsa::VerifyingKey::from(&ecdsa_key), true);
+
     let output = json!({
-        "public_key": secret_key.node_public_key(),
+        "bls_public_key": secret_key.node_public_key(),
```
JamesHinshelwood, line 29: We need to update `infra/tf/locals.tf` if we're changing these keys. cc @frankmeds in case there are other references.

JamesHinshelwood (changes requested):
LGTM, except we should fix the TF.

zilliqa/src/bin/convert-key.rs
frankmeds: good catch @JamesHinshelwood. It should be updated only in https://github.com/Zilliqa/zq2/blob/secretkey-small-cleanup/infra/tf/locals.tf#L8

theo-zil: Great catch, thanks!

infra/tf/locals.tf
```diff
@@ -5,8 +5,8 @@
 locals {
   binary_location      = "${path.module}/../../target/x86_64-unknown-linux-gnu/release/zilliqa"
   labels               = merge(var.labels, { "zq2-network" = var.network_name })
-  bootstrap_public_key = data.external.bootstrap_key_converted.result.public_key
+  bootstrap_public_key = data.external.bootstrap_key_converted.result.bsl_public_key
```
JamesHinshelwood, line 8:
```suggestion
  bootstrap_public_key = data.external.bootstrap_key_converted.result.bls_public_key
```

JamesHinshelwood (changes requested):

infra/tf/locals.tf
theo-zil: Thanks for the catch, whoops...

PR #1226, opened 2024-07-29
PopVerify pre-compile for proof-of-possession.

shawn-zil:
Used a custom pre-compile solution instead. A native Solidity solution would require additional 381-bit computations, primarily to decompress the points and compute y^2 = x^3 + 4, which would incur additional complexity by using large-integer libraries as the EVM data types support 256-bit computations natively only.

Added the use of the new blsful library because pop-prove and pop-verify require signatures with a different domain separation tag than that used for normal signatures. The existing bls-signatures library has a hard-coded DST for normal signatures only.


zilliqa/src/precompiles.rs
```diff
@@ -63,14 +69,85 @@ impl ERC20Precompile {
...
+        handler.1(&input[4..], gas_price, context)
+    }
 }
 
-impl ContextStatefulPrecompile<&State> for ERC20Precompile {
+pub(crate) struct PopVerifyPrecompile;
+
+// keep in-sync with zilliqa/src/contracts/deposit.sol
+impl PopVerifyPrecompile {
+    const POP_VERIFY_GAS_PRICE: u64 = 1_000_000u64; // FIXME: Gas Price?
```
shawn-zil, line 109: I'm not sure what is a good number to use as the gas price. So, I arbitrarily put in 1M.

theo-zil (approved):
LGTM - though this will probably have conflicts with #1160, and I'd rather that one be merged first since it's been open longer.
However I haven't touched the precompile code myself yet so I'll leave it up to you and James to see which way you want to resolve the conflicts.

theo-zil (changes requested):
Forgot to CR - the ECDSA methods in crypto.rs don't need to be re-added

zilliqa/src/precompiles.rs
JamesHinshelwood: Happy to leave like this for now and discuss with @DrZoltanFazekas soon :)

JamesHinshelwood (commented):
We should also update `deposit_stake` in `z2/src/validators.rs` to send the PoP

zilliqa/src/contracts/deposit.sol
```diff
@@ -82,22 +93,58 @@ contract Deposit {
...
-        revert("call must come from a reward address corresponding to a staker with more than 10% stake");
+        revert(
+            "call must come from a reward address corresponding to a staker with more than 10% stake"
+        );
     }
 
-    function deposit(bytes calldata blsPubKey, bytes calldata peerId, bytes calldata /* signature */, address rewardAddress) public payable {
+    // keep in-sync with zilliqa/src/precompiles.rs
+    function _popVerify(
+        bytes memory pubkey,
```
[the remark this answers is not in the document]
shawn-zil: The pubkey is required for verification. It just so happens that it's also the message. So, I think it's better to leave it as pubkey.

zilliqa/src/crypto.rs
```diff
@@ -250,6 +251,20 @@ impl SecretKey {
...
+
+    pub fn as_ecdsa(&self) -> k256::ecdsa::SigningKey {
+        // `SigningKey::from_bytes` can fail for two reasons:
+        // 1. The bytes represent a zero integer. However, we validate this is not the case on construction.
+        // 2. The bytes represent an integer less than the curve's modulus. However for ECDSA, the curve's order is
+        //    equal to its modulus, so this is impossible.
+        // Therefore, it is safe to unwrap here.
+        k256::ecdsa::SigningKey::from_bytes(&self.bytes.into()).unwrap()
+    }
+
```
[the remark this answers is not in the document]
shawn-zil: Will remove.

zilliqa/src/contracts/deposit.sol
```diff
@@ -82,22 +93,58 @@ contract Deposit {
...
+            hex"bfd24965", // bytes4(keccak256("popVerify(bytes,bytes)"))
+            signature,
+            pubkey
+        );
+        uint inputLength = input.length;
+        bytes memory output = new bytes(32);
+        bool success;
+        assembly {
+            success := staticcall(
+                gas(), // FIXME: Gas?
```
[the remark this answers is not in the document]
shawn-zil: Will remove.

z2/src/validators.rs
```diff
@@ -164,6 +170,17 @@ fn hex_string_to_u8_20(hex_str: &str) -> Result<[u8; 20], &'static str> {
     Ok(array)
 }
 
+pub fn pop_prove(secret_key: &str) -> Result<()> {
```
shawn-zil, line 173: Not sure where is best to compute the POP signature.
1. Add a new `z2 pop` command that computes the pop-signature given a secret-key; or
2. Alter the `deposit` command to take a secret-key instead, and compute both public-key and pop-signatures from it; or
3. Provide an external tool/script that computes the pop-signature given a secret-key; or
4. All the above.

shawn-zil:
> Forgot to CR - the ECDSA methods in crypto.rs don't need to be re-added

Removed.

theo-zil (approved):

z2/src/validators.rs
JamesHinshelwood: @chetan-zilliqa Any opinions? I can see arguments for both 1 or 2.

chetan-zilliqa: Okay, so with option 1, the PoP signature is created separately, and the validator operator must pass it as a parameter to the z2 deposit command. In option 2, both the public-key and pop-signature would be generated internally by the tool, eliminating the need to pass the --public-key parameter to z2 deposit. This simplifies the process for the node operator by removing an additional step. Option 2 seems more efficient to me. However, it appears that z2 deposit currently accepts the secret key as an argument, so it might not require any changes. Apologies if I've misunderstood or caused any confusion

JamesHinshelwood (dismissed):

z2/src/validators.rs
shawn-zil: @chetan-zilliqa for option 2, the `z2 deposit` command currently only accepts the public-key. So, it would require a change to take the private-key as an input.

As for options 3 & 1, it might also be feasible to modify the `convert-key` tool to dump out the public-key and pop-signature instead.

Either way, it's a small change. So, just let me know what's the preference.

shawn-zil: @chetan-zilliqa I understand what you mean now with #1234 but that is still in a draft stage. I guess that if you intend to ask users to supply an actual `--private-key` to the `z2 deposit` command, then I guess you can adapt the code from `convert-key.rs` to generate the `pop-signature` as well.

chetan-zilliqa: It's ok @shawn-zil . Yes. I am open to it.

chetan-zilliqa (approved):

PR #1227, opened 2024-07-29
Attempt to remove the contracts.json from git diff linecounts.

theo-zil:
It's a generated file so can effectively be treated as binary - specifically, we pretty much never need an explicit line-by-line diff. Thus this is set in the .gitattributes file. Hopefully, this should make diffs on pull requests display relevant linecounts, rather than exploding into the tens of thousands whenever contracts are recompiled and thus hiding the true change size.

If this works, this could also be added to Cargo.lock files, though their impact on diffs is usually minor so I left that for now.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+zilliqa/src/contracts/compiled.json -diff


JamesHinshelwood (approved):
FWIW I'm in favour of keeping the diff for `Cargo.lock` because:
1. Its content isn't a deterministic function of the rest of the repository (like `compiled.json` is) - i.e. there are multiple valid `Cargo.lock`s at a time
2. It is occasionally useful as a reviewer to see when dependency updates are sneaking in when we wouldn't expect them to.

PR #1232, opened 2024-07-30
Clean up alloy dependencies

JamesHinshelwood:
Instead of depending on the alloy subcrates individually, we just depend on the supercrate and enable the features we need. This means dependency updates only touch one version. Thanks @yaron-zilliqa for pointing out that this was possible.


theo-zil (approved):
nice

theo-zil (approved):

PR #1234, opened 2024-07-30
Rename wallet variable to private key

chetan-zilliqa:


JamesHinshelwood (approved):
We should update `deposit.md` too and any other docs that mention this.

chetan-zilliqa:
> We should update `deposit.md` too and any other docs that mention this.

Yes, I will test it against my local network and also modify the doc.

Issue #1027, opened 2024-06-06
GetSmartContractCode() doesn't return EVM code and eth_getCode() doesn't return Scilla code


rrw-zilliqa:
I believe the Zilliqa 1 behaviour for returning code (in EthRpcMethods.cpp, LookupServer.cpp and JSONConversion.cpp) is:

 * GetSmartContractCode() returns scilla code, or EVM code prefixed with the characters 'E' 'V' 'M'
 * eth_getCode() and eth_getTransactionXXX() returns EVM code if the contract is EVM, and Scilla code if not
 * GetTransaction() returns scilla code or EVM code prefixed with 'E' 'V' 'M'

(basically, code is stored as either Scilla or EVM code prefixed with 'E' 'V' 'M' and the eth_XXX() methods strip 'E' 'V' 'M' if they find it).

zq2 doesn't do this, though admittedly only a small number of applications are likely to care.  We should either implement the old behaviour or note in the documentation what the new behaviour is.

PR #1235, opened 2024-07-30
Change GetSmartContractCode to return EVM code

joseph-zilliqa:
* Fixes https://github.com/Zilliqa/zq2/issues/1027

* Return EVM code from `GetSmartContractCode` by encoding the bytecode as hexadecimal

### DIFF
--- a/zilliqa/src/api/zil.rs
+++ b/zilliqa/src/api/zil.rs
@@ -342,12 +342,12 @@ fn get_smart_contract_code(params: Params, node: &Arc<Mutex<Node>>) -> Result<Va
         .get_state(&block)?
         .get_account(smart_contract_address)?;
 
-    let (code, language) = match account.code {
+    let (code, type_) = match account.code {
         Code::Evm(ref bytes) => (hex::encode(bytes), "evm".to_string()),
         Code::Scilla { code, .. } => (code, "scilla".to_string()),
     };
 
-    Ok(json!({ "code": code, "lanuage": language }))
+    Ok(json!({ "code": code, "type": type_ }))
 }
 
 fn get_smart_contract_init(params: Params, node: &Arc<Mutex<Node>>) -> Result<Value> {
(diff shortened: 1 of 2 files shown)


zilliqa/src/api/zil.rs
```diff
@@ -341,8 +342,13 @@ fn get_smart_contract_code(params: Params, node: &Arc<Mutex<Node>>) -> Result<Va
         .get_state(&block)?
         .get_account(smart_contract_address)?;
 
-    let Some((code, _)) = account.code.scilla_code_and_init_data() else {
-        return Err(anyhow!("Address not contract address"));
+    let code = match account.code {
+        Code::Evm(ref bytes) => {
+            "EVM".to_string()
+                + &String::from_utf16(&bytes.into_iter().map(|v| *v as u16).collect::<Vec<_>>())
```
JamesHinshelwood, line 348: Why utf16? Surely this will fail most of the time. `bytes` is usually not going to be a valid UTF-16 string.

Should we instead use `hex::encode`?

JamesHinshelwood (changes requested):

zilliqa/src/api/zil.rs
joseph-zilliqa: ZQ1 returns this directly as a string, the only way I could get a result that is similar is by encoding each byte as a utf16 code point which allows all values of one byte whereas utf8 does not allow all one byte values

JamesHinshelwood: Oh yeah
```
curl -X POST https://api.zilliqa.com/ -H 'content-type: application/json' -d '{"jsonrpc":"2.0","id":"1","method":"GetSmartContractCode","params":["e30161F32A019d876F082d9FF13ed451a03A2086"]}'
```
The current response is super cursed :)

I think we should make a breaking change here and return EVM code as a hex string. I don't think there's any need to prefix them with `"EVM"` either.

We should document this in `docs/api/zilliqa/getsmartcontractcode.doc.md`

joseph-zilliqa: Surely there still needs to be a way to differentiate scilla code from evm here? Is it worth adding an extra field to the output such as `language` with `scilla` or `evm`?

theo-zil: To be pedantic, EVM isn't a language per se, so I personally would call the field `runtime` rather than `language`

JamesHinshelwood: `runtime`, `engine`, `type` all seem decent to me. I agree that `language` is misleading.

PR #1265, opened 2024-08-06
Migrate bls crates to blsful

shawn-zil:
Description in the linked ticket.

### DIFF
--- a/zilliqa/src/crypto.rs
+++ b/zilliqa/src/crypto.rs
@@ -8,9 +8,13 @@ use std::fmt::Display;
 
 use alloy::primitives::{Address, B256};
 use anyhow::{anyhow, Result};
-use bls12_381::{G1Projective, G2Affine};
-use bls_signatures::Serialize as BlsSerialize;
-use blsful::Bls12381G2Impl;
+use blsful::{
+    inner_types::{group::prime::PrimeCurveAffine, G2Projective},
+    vsss_rs::ShareIdentifier,
+    AggregateSignature, Bls12381G2, Bls12381G2Impl, MultiPublicKey, MultiSignature, PublicKey,
+    Signature,
+};
+use itertools::Itertools;
 use k256::ecdsa::{signature::hazmat::PrehashVerifier, Signature as EcdsaSignature, VerifyingKey};
 use serde::{
     de::{self, Unexpected},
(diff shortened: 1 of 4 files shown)


JamesHinshelwood:
Performance of processing blocks looks miles better with this PR, nice one :) A `-59.179%` change in processing time per block and `+144.97%` improvement in throughput.

process-blocks-both

(Blame criterion for generating a misleading graph where the axes don't start at zero :stuck_out_tongue:)

zilliqa/src/crypto.rs
```diff
@@ -20,20 +22,44 @@ use sha3::{Digest, Keccak256};
...
+        NodeSignature(Signature::<Bls12381G2Impl>::Basic(
+            blsful::inner_types::G2Projective::identity(),
+        ))
     }
 
     pub fn from_bytes(bytes: &[u8]) -> Result<NodeSignature> {
-        Ok(NodeSignature(bls_signatures::Signature::from_bytes(bytes)?))
+        // Default to Basic signatures - it's the normal signature.
+        Ok(NodeSignature(Signature::<Bls12381G2Impl>::Basic(
+            blsful::inner_types::G2Projective::from_compressed(bytes.try_into()?)
```
JamesHinshelwood, line 38: Just to double check - You verified that `bls_signatures::Signature::from_bytes` was treating the bytes as a compressed signature?

zilliqa/src/crypto.rs
```diff
@@ -20,20 +22,44 @@ use sha3::{Digest, Keccak256};
...
+                .expect("blst_p2_uncompress() error"),
```
JamesHinshelwood, line 39: Should we propagate the error instead? `from_bytes` is returning a `Result`, but this implementation never returns an `Err`. It seems like `from_compressed` will fail if `!p.is_on_curve()` or `!p.is_torsion_free()` and potentially for more reasons if the underlying C library doesn't return success.

zilliqa/src/crypto.rs
```diff
@@ -20,20 +22,44 @@ use sha3::{Digest, Keccak256};
...
-        let signatures: Vec<_> = signatures.iter().map(|s| s.0).collect();
-        Ok(NodeSignature(bls_signatures::aggregate(&signatures)?))
+        let signatures = signatures.iter().map(|s| s.0).collect_vec();
+
+        // IETF standards say N >= 1
+        // Handles single case where N == 1, as AggregateSignature::from_signatures() only handles N > 1.
+        if signatures.len() < 2 {
+            let g = blsful::inner_types::G2Projective::identity();
+            return Ok(NodeSignature(match signatures[0] {
+                Signature::Basic(s) => Signature::Basic(g + s),
```
JamesHinshelwood, line 51: Surely `identity + s = s`? Why the extra step?

zilliqa/src/crypto.rs
```diff
@@ -20,20 +22,44 @@ use sha3::{Digest, Keccak256};
...
+        )))
     }
 
     pub fn aggregate(signatures: &[NodeSignature]) -> Result<NodeSignature> {
-        let signatures: Vec<_> = signatures.iter().map(|s| s.0).collect();
-        Ok(NodeSignature(bls_signatures::aggregate(&signatures)?))
+        let signatures = signatures.iter().map(|s| s.0).collect_vec();
+
+        // IETF standards say N >= 1
+        // Handles single case where N == 1, as AggregateSignature::from_signatures() only handles N > 1.
```
JamesHinshelwood, line 47: https://github.com/hyperledger-labs/agora-blsful/issues/10 - Feel free to reference here in case it gets fixed.

JamesHinshelwood (dismissed):
LGTM

zilliqa/src/crypto.rs
shawn-zil: I followed bls-signatures, which used a fold() with the offset of g. But you're right, it produces the same outcome. I'll remove the extra step.

zilliqa/src/crypto.rs
shawn-zil: Yes, there's a check in `bls_signatures::Signature::from_bytes`:
```
    if raw.len() != G2_COMPRESSED_SIZE {
        return Err(Error::SizeMismatch);
    }
```

shawn-zil: But I don't seem to find a length check in the underlying library, so I've added length checks in ours.

JamesHinshelwood (approved):

PR #1271, opened 2024-08-08
fix: z2 join should work with chain known data

frankmeds:


z2/src/utils.rs
```diff
@@ -54,3 +54,15 @@ pub fn string_from_path(in_path: &Path) -> Result<String> {
         .ok_or(anyhow!("Cannot convert path to string"))?
         .to_string())
 }
+
+/// Get local public IP
+pub async fn get_public_ip() -> Result<String> {
+    let output: zqutils::commands::CommandOutput = zqutils::commands::CommandBuilder::new()
+        .silent()
+        .cmd("curl", &["-4", "ipconfig.io"])
```
JamesHinshelwood, line 62: This will break for anyone with a non-standard network set up. We should at least document that this is required and we will infer the external IP automatically so node operators are aware.

JamesHinshelwood (approved):

PR #1285, opened 2024-08-13
Integrate Autonat

### LINKED ISSUE #1101: Consider using `libp2p-autonat` to let nodes automatically work out their external address
Right now we rely on manual configuration to tell a node its external address.

maxconway:
- Plus docker compose file and config file used for testing
- Using autonat v1 since v2 didn't seem super mature yet

Closes #1101

The docker compose file is used for testing the network without mDNS. I actually just commented out mDNS for testing, we could try to block it in the compose file if we need that capability in future.

### DIFF
--- a/z2/Cargo.toml
+++ b/z2/Cargo.toml
@@ -43,7 +43,7 @@ itertools = "0.13.0"
 jsonrpsee = {version = "0.22.4", features = ["client"]}
 k256 = "0.13.3"
 lazy_static = "1.5.0"
-libp2p = {version = "0.54.0", features = ["identify", "autonat"]}
+libp2p = {version = "0.54.0", features = ["identify"]}
 octocrab = "0.39.0"
 primitive-types = "0.12.2"
 prost = "0.13.1"
--- a/z2/Cargo.toml
+++ b/z2/Cargo.toml
@@ -43,7 +43,7 @@ itertools = "0.13.0"
 jsonrpsee = {version = "0.22.4", features = ["client"]}
 k256 = "0.13.3"
 lazy_static = "1.5.0"
-libp2p = {version = "0.54.0", features = ["identify"]}
+libp2p = {version = "0.54.0", features = ["identify", "autonat"]}
 octocrab = "0.39.0"
 primitive-types = "0.12.2"
 prost = "0.13.1"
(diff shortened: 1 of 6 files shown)


z2/Cargo.toml
```diff
@@ -43,7 +43,7 @@ itertools = "0.13.0"
 jsonrpsee = {version = "0.22.4", features = ["client"]}
 k256 = "0.13.3"
 lazy_static = "1.5.0"
-libp2p = {version = "0.54.0", features = ["identify"]}
+libp2p = {version = "0.54.0", features = ["identify", "autonat"]}
```
JamesHinshelwood, line 46: Isn't this activating the feature in the wrong `Cargo.toml`? Should this move to `zilliqa/Cargo.toml`?

I guess this still works, as long as you `cargo build` at the root of the workspace, since the features get unified anyway.

zilliqa/src/p2p_node.rs
```diff
@@ -112,6 +114,17 @@ impl P2pNode {
             )
             .map_err(|e| anyhow!(e))?,
             mdns: mdns::Behaviour::new(Default::default(), peer_id)?,
+            autonat: autonat::Behaviour::new(
+                peer_id,
+                autonat::Config {
```
JamesHinshelwood, line 119: Is there a good reason we don't just use the default? If there is, can we add the reason(s) in a comment?

JamesHinshelwood (dismissed):

z2/Cargo.toml
maxconway: Yep, my bad, I'll switch that over

zilliqa/src/p2p_node.rs
```diff
@@ -112,6 +114,19 @@ impl P2pNode {
...
+                    // Config changes to speed up autonat initialization.
+                    retry_interval: Duration::from_secs(10),
+                    refresh_interval: Duration::from_secs(30),
+                    boot_delay: Duration::from_secs(5),
+                    throttle_server_period: Duration::ZERO,
+                    // Don't attempt to reach IPs that are known to be at a private.
+                    only_global_ips: false,
```
JamesHinshelwood, line 126: Doesn't your comment mean the opposite of the configuration?

If `only_global_ips` is `false`, I assume that means we allow private IPs, but your comment says the opposite?

Am I confused?

JamesHinshelwood (approved):

zilliqa/src/p2p_node.rs
maxconway: No, I think I am. I'm going to remove it and set it to default, I suspect it's doing nothing anyway

JamesHinshelwood (approved):

Issue #1297, opened 2024-08-14
Improve validator address verification in deposit contract


theo-zil:
Currently, we implement a standard proof-of-possession on validator registration, to ensure that the caller submitting the transaction has access to the private key corresponding to the `blsPubKey` parameter. This is important to avoid a malicious actor from signing up somebody else's public key without their permission, which can cause issues with BLS aggregate signatures: https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html

However, the standard PoP implementation is simply a signature of the public key on itself, which has no protection against replay attacks. In `deposit.sol`, we never delete staker data after first registration which avoids the possibility of naive replay - trying to re-register with the same public key and PoP will fail since the transaction will not match the known staker address. But there are still two vulnerabilities for us:
 - The transaction can be front-ran while in the mempool, since the PoP will have been revealed; a malicious control address will then be registered with the staker, instead of the staker's real address.
 - The same staker registration transaction will be valid on different chains - in particular, on different shards (but this also applies to e.g. testnet vs. mainnet). An attacker can then obtain the PoP from a staker's registration transaction on one shard, and successfully bypass the PoP requirement when registering on other shards.

To fix this, the signature provided should not be simply a basic PoP; instead, it should sign over the `from` address of the transaction (making a front-run attempt from a different account invalid) and the shard/chain id (to avoid replays on other chains).

Issue #1135, opened 2024-07-03
Move consensus config on-chain (into shard contract)


theo-zil:
Every shard has a shard contract at a hard-coded address (the main shard has a `ShardRegistry is Shard` contract). New validators joining that shard should be able to bootstrap using the values from that contract.

This is dependent on either light nodes, or at least some kind of fast sync (obtain checkpoint using external trust -> set config to values from state in checkpoint -> go from there).

If full on-chain bootstrapping is implemented, most or all of the `ConsensusConfig` struct will no longer need to be specified on startup (as the node will fetch those values automatically), and can be removed from all config files.

PR #1298, opened 2024-08-14
Validator epochs/ZIP25

theo-zil:
Preliminary draft of the new validator state machine, implementing epochs.

- The validator committee no longer changes except at epoch boundary blocks.
- Joining or leaving the committee is now subject to a waiting queue of no less than 1 epoch (and currently, no more than 2). Thus, at any given epoch, it is possible to mostly (see #1338) know the validator set for the next epoch.
- Epoch changes are handled by a `tick_epoch()` function in the deposit contract.

An important consequence of all this is that our consensus now depends on an EVM call succeeding. Inline comment on this in the code below.

Further work out of scope of this PR:
- Fill out missing integration tests - #1341
- Implement unsuspend, and consensus-based suspend - #1337 
- Epoch length tracking in EVM - part of #1135. (Currently, the epoch length is hardcoded in `deposit.sol`.)
- Improving staker address and pubkey validation on registration - #1297 
- Allowing validators to change their registered addresses - #1301 
- Adjust slashing and suspension queues - #1338
- Consider implementing churn limits on the validator set - #1300 
- Improving epoch change transaction tracking - #1299 
- Slashing mechanisms - currently, this reuses the previous mechanism where any validator with >=10% of total stake can slash any other validator. Additionally, slashed funds are not kept track of (effectively, burned).
- Add an unbonding period to unstaked funds - #1328 
- Allow reusing instaked-but-not-yet-withdrawn funds - #1329 
- Not directly necessary, but somewhat related: block number tracking - #1339


zilliqa/src/consensus.rs
```diff
@@ -1053,6 +1053,16 @@ impl Consensus {
 
         let node_config = &self.config;
 
+        if self.block_is_first_in_epoch(parent.number() + 1) {
+            // Here we ignore any errors and proceed with block creation, under the philosophy that
+            // it is better to have the network run but with broken epochs, than it would be to
+            // crash the entire network on a failing epoch transition.
```
theo-zil, line 1059: Our consensus now depends on successful EVM calls - and not just read-only checks as before, but a complex write transaction.

This raises the question of what to do if the `tick_epoch()` call fails for any reason. Should we assume that it must never fail, and crash the node if it does (potentially crashing the network if a buggy state is reached)? The approach taken here is to soft-ignore errors; the network then will remain alive, but with a static validator set since epoch changes will be broken. Measures can then be taken to upgrade to fix the root cause of the issue (e.g. upgrading the deposit contract) to restore normal functionality.

But, especially right now in the early testnet phases, perhaps hard-crashing the network on any bugs might be preferable to discover any issues immediately.

zilliqa/src/consensus.rs
```diff
@@ -1053,6 +1053,16 @@ impl Consensus {
...
+            match state.tick_epoch() {
+                Ok(()) => (),
+                Err(e) => error!("Unable to transition the epoch - EVM error: {e}"),
+            };
+        }
```
theo-zil, line 1064: Our consensus now depends on successful EVM calls - and not just read-only checks as before, but a complex write transaction.

This raises the question of what to do if the `tick_epoch()` call fails for any reason. Should we assume that it must never fail, and crash the node if it does (potentially crashing the network if a buggy state is reached)? The approach taken here is to soft-ignore errors; the network then will remain alive, but with a static validator set since epoch changes will be broken. Measures can then be taken to upgrade to fix the root cause of the issue (e.g. upgrading the deposit contract) to restore normal functionality.

But, especially right now in the early testnet phases, perhaps hard-crashing the network on any bugs might be preferable to discover any issues immediately.

zilliqa/src/contracts/deposit.sol
```diff
@@ -1,26 +1,138 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 pragma solidity ^0.8.20;
 
+// Implementation of a circular queue/buffer of validator public keys
```
DrZoltanFazekas, line 4: I wonder if we could use simply dynamic arrays of type `Value[]` instead, with elements being pushed during the epoch and popped until the array is empty when `tickEpoch()` is called.

zilliqa/src/contracts/deposit.sol
```diff
@@ -138,61 +207,238 @@ contract Deposit {
...
+        committeeKeys.pop();
+    }
+
+
+    function currentEpoch() public view returns (uint64) {
+        return uint64(block.number / blocksPerEpoch); // TODO - take the value from the shard contract rather than storing it locally
+    }
+
+    function currentQueueIndex() private view returns (uint8) {
+        return uint8(currentEpoch() % WAIT_EPOCHS);
```
DrZoltanFazekas, line 334: would work if the `WAIT_PERIOD` were the same for staking and unstaking, but for the latter we'll need a longer unbonding period

zilliqa/src/contracts/deposit.sol
```diff
@@ -138,61 +207,238 @@ contract Deposit {
...
         require(pop, "rogue key check");
 
-        uint256 keyIndex = _stakersMap[blsPubKey].keyIndex;
-        if (keyIndex == 0) {
-            // The staker will be at index `_stakerKeys.length`. We also need to add 1 to avoid the 0 sentinel value.
-            _stakersMap[blsPubKey].keyIndex = _stakerKeys.length + 1;
-            _stakerKeys.push(blsPubKey);
+        // If staker info isn't stored yet, do so
+        // peerId is guaranteed to be non-zero for all stakers
+        if (_stakersMap[blsPubKey].controlAddress == address(0)) {
```
DrZoltanFazekas, line 218: due to this condition the `deposit()` function now can't be used for updating the validators `rewardAddress` i.e. we should provide an extra function

zilliqa/src/contracts/deposit.sol
```diff
@@ -138,61 +207,238 @@ contract Deposit {
...
+            // Remove the staker from the total stake and the active committee
+            totalStake -= staker.stakedBalance;
+            // Only remove the staker from the committee if we know it's in it currently
+            // There are edge cases where it may not be necessary - e.g. if a withdrawal is processed on the same epoch as a slash/suspend, it will already have been removed just above
+            if (staker.keyIndex >= 0) {
+                swapRemoveValidator(uint(staker.keyIndex));
+            }
+
+            // Apply penalty, if any (might be 0 e.g. for voluntary suspension)
+            // TODO: collect these rewards somewhere? Right now they're effectively burned
```
DrZoltanFazekas, line 386: a fraction of the slashing penalty shall be paid as whistleblower reward to whoever submitted the slashing evidence, the rest must be burned back to the zero account

zilliqa/src/contracts/deposit.sol
```diff
@@ -138,61 +207,238 @@ contract Deposit {
...
+    function tickEpoch() public {
+        // if current epoch has already been processed, exit
+        if (headProcessedEpoch >= currentEpoch()) {
+            return;
+        }
+
+        uint8 queueIndex = currentQueueIndex();
+
+        // Process pending queue for this epoch
+        for (uint i = 0; i < pendingQueues[queueIndex].length(); ++i) {
```
DrZoltanFazekas, line 346: Malicious validators could mount a DoS attack by calling `depositTopup()` a lot of times with small values so that `tickEpoch()` would need to consume more than the block limit iterating through the `pendingQueue`

zilliqa/src/contracts/deposit.sol
```diff
@@ -138,61 +207,238 @@ contract Deposit {
...
         bool pop = _popVerify(blsPubKey, signature);
         require(pop, "rogue key check");
 
-        uint256 keyIndex = _stakersMap[blsPubKey].keyIndex;
-        if (keyIndex == 0) {
-            // The staker will be at index `_stakerKeys.length`. We also need to add 1 to avoid the 0 sentinel value.
-            _stakersMap[blsPubKey].keyIndex = _stakerKeys.length + 1;
-            _stakerKeys.push(blsPubKey);
+        // If staker info isn't stored yet, do so
+        // peerId is guaranteed to be non-zero for all stakers
```
DrZoltanFazekas, line 217: Is it or do have have to check, and even if it is, why is the comment relevant here?

zilliqa/src/contracts/deposit.sol
```diff
@@ -138,61 +207,238 @@ contract Deposit {
...
+        // next epoch boundary (a fast-track compared to simply withdrawing all stake, which would take up to 2 epochs),
+        // but would remain liable for liveness until then.
+        _stakersMap[blsPubKey].status = Status.Suspended;
+
+        Value storage newSuspension = kickQueue.pushBack();
+        newSuspension.key = blsPubKey;
+        // no associated amount
+    }
+
+    function eject(
```
DrZoltanFazekas, line 292: this is the successor of `tempRemoveStaker()` and will be removed once penalties are fully implemented, right?

DrZoltanFazekas (commented):
I was only reviewing the deposit contract, will continue with the rest tomorrow

PR #1335, opened 2024-08-22
Factorise the PopVerify() gas-price.

shawn-zil:
@DrZoltanFazekas need your input on a suitable gas price.


JamesHinshelwood:
Moved to #1879

PR #1350, opened 2024-08-26
Support CHAINID, BLOCKNUMBER, and TIMESTAMP in scilla contracts.

saeed-zil:


evm_scilla_js_tests/test/scilla/BlockNumber.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  it("Deploy BlockNumber contract @block-1", async () => {
+    expect(contract.address).to.be.properAddress;
+  });
+
+  it("Call BlockNumber contract -  EventBlockNumber @block-1", async () => {
+    let result = (await hre.zilliqaSetup.zilliqa.blockchain.getNumTxBlocks()).result;
+    const tx = await contract.EventBlockNumber();
+    expect(tx).to.have.eventLogWithParams("BlockNumber", {value: result});
```
JamesHinshelwood, line 22: Should we assert `BlockNumber == tx.receipt.block_number` ?

evm_scilla_js_tests/test/scilla/Timestamp.ts
```diff
@@ -12,7 +12,7 @@ xdescribe("Scilla timestamp #parallel", () => {
     const blockCount = await ethers.provider.getBlockNumber();
     const blockTimestamp = (await ethers.provider.getBlock(blockCount)).timestamp;
     const tx = await contract.EventTimestamp(blockCount);
-    const timestamp = Number(tx.getReceipt().event_logs[0].params[0].value.arguments[0]);
+    const timestamp = Number(JSON.parse(tx.receipt.event_logs[0].params[0].value).arguments[0]);
```
JamesHinshelwood, line 15: Can you compare the responses for ZQ1 and ZQ2 before we merge this change? The fact that we needed to add a `JSON.parse` suggests we might have changed the format? (double encoding?)

zilliqa/src/exec.rs
```diff
@@ -1013,6 +1020,10 @@ impl PendingState {
         load_account(&self.pre_state, &mut self.new_state, address)
     }
 
+    pub fn network_id(&self) -> u64 {
+        self.pre_state.chain_id - 0x8000
```
JamesHinshelwood, line 1024: Can we move `zilliqa::api::zil::network_id` to a central place and use it both in the API layer and here? I don't want to have two sources of truth for the `- 0x8000` logic.

zilliqa/src/scilla.rs
```diff
@@ -775,7 +787,19 @@ impl ActiveCall {
...
+    fn fetch_blockchain_info(&self, name: String, _args: String) -> Result<String> {
+        match name.as_str() {
+            "CHAINID" => Ok(self.state.network_id().to_string()),
+            "BLOCKNUMBER" => Ok(self.current_block.number.to_string()),
+            "TIMESTAMP" => Ok(self
+                .current_block
+                .timestamp
+                .duration_since(SystemTime::UNIX_EPOCH)?
+                .as_micros()
+                .to_string()),
```
JamesHinshelwood, line 799: Double checking that Scilla expects microseconds here?

zilliqa/src/scilla.rs
```diff
@@ -552,9 +557,10 @@ impl StateServer {
                     return Err(err("no active call"));
                 };
 
-                active_call.fetch_blockchain_info(query_name, query_args);
-
-                Ok::<_, ErrorObject>(())
+                match active_call.fetch_blockchain_info(query_name, query_args) {
+                    Ok(value) => Ok(Value::Array(vec![true.into(), value.into()])),
```
JamesHinshelwood, line 561: When should this return `false`?

JamesHinshelwood (approved):
LGTM, thanks - Just some minor comments.

zilliqa/src/scilla.rs
chetan-zilliqa: Actually, the `BLOCKNUMBER`, `BLOCKHASH` and `TIMESTAMP` refers to the value of second argument passed in the function and not the current block. so we would need to interpret `_args` and read the persistence to read those parameters.Here is the zq1 implementation - https://github.com/Zilliqa/Zilliqa/blob/ba6dc3d1cd08de09ea9f543340c983d83ecdfde7/src/libScilla/ScillaIPCServer.cpp#L516

chetan-zilliqa (changes requested):

evm_scilla_js_tests/test/scilla/BlockNumber.ts
saeed-zil: `expect(tx).to.have.eventLogWithParams("BlockNumber", {value: result});` does the job, but I'll rename `result` to something more meaningful.

evm_scilla_js_tests/test/scilla/Timestamp.ts
saeed-zil: Good point! :) I was also suspicious of this new requirement.

zilliqa/src/scilla.rs
saeed-zil: @chetan-zilliqa Thanks! Good catch! :) But the `BLOCKNUMBER` seems to have zero-param

chetan-zilliqa: Yeah, you are correct. `BLOCKNUMBER` is the current block number..

evm_scilla_js_tests/test/scilla/Timestamp.ts
saeed-zil: I created #1380 for it. I know the cause and I have a solution but it needs more tests so let's merge this ticket first then I'll raise another PR.

chetan-zilliqa (approved):

z2/src/converter.rs
JamesHinshelwood, line 1: cc @bzawisto

zilliqa/src/node.rs
```diff
@@ -131,6 +131,26 @@ pub struct Node {
...
+        ChainId {
+            eth: eth_chain_id,
+            zil: Self::zil_chain_id(eth_chain_id),
+        }
+    }
+
+    pub fn zil_chain_id(eth_chain_id: u64) -> u64 {
+        eth_chain_id - 0x8000
+    }
 }
```
JamesHinshelwood, line 154: Since we're always setting the `zil` chain ID to `eth - 0x8000` can we instead do
```suggestion
#[derive(Debug)]
pub struct ChainId {
    pub eth: u64,
}

impl ChainId {
    pub fn new(eth_chain_id: u64) -> Self {
        ChainId {
            eth: eth_chain_id,
        }
    }

    pub fn zil(&self) -> u64 {
        self.eth - 0x8000
    }
}
```
?

zilliqa/src/exec.rs
```diff
@@ -1070,6 +1071,18 @@ impl PendingState {
         }
     }
 
+    pub fn get_zil_chain_id(&self) -> u64 {
```
JamesHinshelwood, line 1074: Tiny nit but I prefer 
```suggestion
    pub fn zil_chain_id(&self) -> u64 {
```
:)

zilliqa/src/block_store.rs
```diff
@@ -241,6 +261,10 @@ impl BlockStore {
         self.get_block(hash)
     }
 
+    pub fn get_highest_block_number(&self) -> Result<Option<u64>> {
```
JamesHinshelwood, line 264: Can you replace the uses of `db.get_highest_block_number()` with `block_store.get_highest_block_number()` if we are adding this. Some are withing `BlockStore` itself, but I guess we have some in `Consensus` too.

zilliqa/src/node.rs
```diff
@@ -131,6 +131,26 @@ pub struct Node {
     message_sender: MessageSender,
     reset_timeout: UnboundedSender<Duration>,
     pub consensus: Consensus,
+    pub chain_id: ChainId,
+}
+
+#[derive(Debug)]
```
JamesHinshelwood, line 137: Please impl `Copy` and avoid `&ChainId`. Its only 128 bits :)

zilliqa/src/scilla.rs
```diff
@@ -181,6 +182,7 @@ impl Scilla {
         }
     }
 
+    #[allow(clippy::too_many_arguments)]
```
JamesHinshelwood, line 185: Is this needed? You haven't added any arguments.

zilliqa/src/scilla.rs
```diff
@@ -568,9 +570,10 @@ impl StateServer {
                     return Err(err("no active call"));
                 };
 
-                active_call.fetch_blockchain_info(query_name, query_args);
-
-                Ok::<_, ErrorObject>(())
+                match active_call.fetch_blockchain_info(query_name, query_args) {
+                    Ok((success, value)) => Ok(Value::Array(vec![success.into(), value.into()])),
```
JamesHinshelwood, line 574: Small nit: Can we rename `success` to something like `present`? If `fetch_blockchain_info` returns `Ok((false, _))` it **succeeded**, but the block queried wasn't present.

zilliqa/src/state.rs
```diff
@@ -44,20 +46,25 @@ pub struct State {
     scilla_lib_dir: String,
     pub block_gas_limit: EvmGas,
     pub gas_price: u128,
+    pub zil_chain_id: u64,
```
JamesHinshelwood, line 49: Why not store a `ChainId`? Then we should also remove the `chain_id` argument from some of the execution functions in `exec.rs` (e.g. `apply_transaction_evm`) and instead use `self.chain_id.eth`?

JamesHinshelwood (commented):
LGTM, thanks! I made a few minor comments. Feel free to disagree with me and I'll press approve instead :)

JamesHinshelwood (approved):

Issue #1241, opened 2024-07-31
Support bech32 addresses for some APIs


chetan-zilliqa:
In ZQ1, the following APIs can accept either a hexadecimal address or a bech32 address with a `zil` HRP. Currently, we only support hex addresses.

- `GetBalance`
- `GetSmartContractState`
- `GetSmartContractSubState`
- `GetSmartContractInit`
- `GetSmartContractCode`
- `GetSmartContracts`

PR #1356, opened 2024-08-27
Accept bech32 address in some APIs

joseph-zilliqa:
- Fixes https://github.com/Zilliqa/zq2/issues/1241

- Accept bech32 address in APIs:
  - `GetBalance`
  - `GetSmartContractState`
  - `GetSmartContractSubState`
  - `GetSmartContractInit`
  - `GetSmartContractCode`
  - `GetSmartContracts`

### DIFF
--- a/zilliqa/src/api/zil.rs
+++ b/zilliqa/src/api/zil.rs
@@ -60,6 +60,35 @@ pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
 }
 
 #[derive(Deserialize)]
+#[serde(transparent)]
+struct ZilAddress {
+    #[serde(deserialize_with = "deserialize_zil_address")]
+    inner: Address,
+}
+
+impl From<ZilAddress> for Address {
+    fn from(value: ZilAddress) -> Self {
+        value.inner
+    }
+}
+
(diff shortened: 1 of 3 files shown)


zilliqa/src/api/zil.rs
```diff
@@ -59,6 +59,35 @@ pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
...
+    }
+}
+
+fn deserialize_zil_address<'de, D>(deserializer: D) -> Result<Address, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+
+    let addr = if s.starts_with("zil") {
```
JamesHinshelwood, line 81: Small nit: I'd prefer if the logic went: "Try to parse with `bech32::decode`. If it succeeds, check the HRP is `zil`. If it fails, try to parse as a hex string".

I guess it is equivalent to this, since if the string starts with `zil` then that must be the HRP, but it feels slightly less robust?

JamesHinshelwood (approved):
LGTM thanks, one comment you might want to consider or ignore :)

JamesHinshelwood (approved):

Issue #1292, opened 2024-08-14
Support _codehash in scilla contracts


saeed-zil:
Test(s) to enable after support:

* https://github.com/Zilliqa/zq2/blob/main/evm_scilla_js_tests/test/scilla/Codehash.ts#L6

PR #1358, opened 2024-08-27
Add support for _codehash in scilla contracts

joseph-zilliqa:
* Fixes https://github.com/Zilliqa/zq2/issues/1292

* Add support for _codehash in scilla contracts

* Enable test scilla/Codehash.ts


JamesHinshelwood:
cc @saeed-zil @chetan-zilliqa 

Is there something to add at the execution side as well as the API side?

chetan-zilliqa (approved):
We are good.There is no api that returns the code hash.

zilliqa/src/scilla.rs
```diff
@@ -715,7 +718,19 @@ impl ActiveCall {
                 Ok(Some((val, "ByStr20".to_owned())))
             }
             "_codehash" => {
-                todo!()
+                let code_bytes = match &account.account.code {
+                    Code::Evm(bytes) => bytes.clone(),
```
JamesHinshelwood, line 722: Double check the behaviour in ZQ1, but I'd guess we should treat EVM accounts as if they had no Scilla code.

joseph-zilliqa: It appears that ZQ1 has the same behaviour

JamesHinshelwood (approved):
cc @saeed-zil for potential conflicts with #1350

JamesHinshelwood (approved):
Thanks. Sorry for one last comment - Can we add an entry at the top of `CHANGELOG.md` documenting this change?

JamesHinshelwood (approved):

Issue #1352, opened 2024-08-26
Global variables in EVM refer to the current and not to the parent block


DrZoltanFazekas:
We pass `parent.header` as the block context for executing all EVM transactions. In particular for e.g. `block.number`, this will return the previous/parent block's number, rather than the currently-being-executed block's number.

This and other properties of the `block` accessible as global variables specified in 
https://docs.soliditylang.org/en/latest/units-and-global-variables.html#block-and-transaction-properties must refer to the block in which the currently executed transaction was included.

PR #1369, opened 2024-08-28
Execute txns with current block rather than their parent

bzawisto:
Fixes: https://github.com/Zilliqa/zq2/issues/1352


shawn-zil (approved):
LGTM. It may need a rebase.

JamesHinshelwood (approved):

Issue #1375, opened 2024-08-29
nodes panic when checking incoming blocks


DrZoltanFazekas:
We observed this on devnet on several API nodes. The log says:
```
message: "called `Option::unwrap()` on a `None` value"
panic.column: 68
panic.file: "zilliqa/src/consensus.rs"
panic.line: 1731
```

It was the unwrap in `consensus.rs` on line 1731:

```
        // Derive the proposer from the block's view
        let proposer = self.leader_at_block(&parent, block.view()).unwrap();
```
since `leader_at_block_raw()` returned `None` on line 2115-2117 because the parent `block` was missing:

```
        let Ok(state_at) = self.try_get_state_at(block.number()) else {
            return None;
        };

```
which is strange because we would have returned an error on line 1710 before we even attempt to get the `proposer` on line 1731:

```
        let Some(parent) = self
            .get_block(&block.parent_hash())
            .map_err(|e| (e, false))?
        else {
            warn!(
                "Missing parent block while trying to check validity of block {}",
                block.number()
            );
            return Err((MissingBlockError::from(block.parent_hash()).into(), true));
        };
```

JamesHinshelwood:
Caused by #1056

PR #1381, opened 2024-08-30
Fix nodes panic when checking incoming blocks

joseph-zilliqa:
* FIxes https://github.com/Zilliqa/zq2/issues/1375


joseph-zilliqa:
I'm not sure if this is the best way to fix this error

JamesHinshelwood:
This happens because we make inconsistent changes to the DB, so this won't fix it.

PR #1390, opened 2024-09-02
Implement `GetNumPeers` API endpoint

maxconway:


zilliqa/src/block_store.rs
```diff
@@ -299,4 +299,8 @@ impl BlockStore {
             .entry(peer)
             .or_insert_with(|| PeerInfo::new(capacity))
     }
+
+    pub fn get_num_peers(&self) -> usize {
+        self.peers.len()
```
JamesHinshelwood, line 304: I think we should get the number of peers from the libp2p `Swarm` instead. The `BlockStore` will only remember a subset of these peers (specifically those who we receive blocks from) and won't evict peers if they are disconnected.

maxconway: What's the best way to get from a node::Node to a libp2p swarm? As far as I can see libp2p swarms are only used in p2p_node.rs, but I can't see any way of getting from a node::Node to a p2p_node::P2pNode. (generate_child_config looks like it's involved in the opposite conversion)

I can try an implement an into function but it feels like there's probably something I'm missing here

maxconway: As far as I can see the brute force way to do this is to bring a reference to the swarm through P2pNode, through NodeLauncher, and then into Node. Then we'll be able to query it. That'll also require a Debug implementation for Swarm<Behaviour>. Does that sound right?

maxconway: I've pulled the number of peers right through now. Does it look right?

zilliqa/src/block_store.rs
```diff
@@ -299,4 +299,8 @@ impl BlockStore {
             .entry(peer)
             .or_insert_with(|| PeerInfo::new(capacity))
     }
+
+    pub fn get_num_peers(&self) -> usize {
```
JamesHinshelwood, line 303: Small nit but I think this is now unused?

JamesHinshelwood (approved):
LGTM, thanks! :)

zilliqa/src/block_store.rs
maxconway: Fixed

JamesHinshelwood (approved):

PR #1419, opened 2024-09-09
Avoid voting for a new view if we aren't in the commitee

### LINKED ISSUE #1399: Non-validators shouldn't view change or send `NewView`s
Currently nodes will increment their view and send a `NewView`, even if they aren't a validator. This makes no sense and we shouldn't do it.

JamesHinshelwood:
Fixes #1399

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -409,11 +409,6 @@ impl Consensus {
 
         trace!("Considering view change: view: {} time since: {} timeout: {} last known view: {} last hash: {}", self.view.get_view(), time_since_last_view_change, exponential_backoff_timeout, self.high_qc.view, head_block.hash());
 
-        let view_difference = self.view.get_view().saturating_sub(self.high_qc.view);
-        let consensus_timeout_ms = self.config.consensus.consensus_timeout.as_millis() as u64;
-        let next_exponential_backoff_timeout =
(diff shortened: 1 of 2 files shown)


zilliqa/src/consensus.rs
```diff
@@ -414,6 +414,19 @@ impl Consensus {
         let next_exponential_backoff_timeout =
             consensus_timeout_ms * 2u64.pow((view_difference + 1) as u32);
```
DrZoltanFazekas, line 415: I forgot to fix this when I changed the formula in `fn get_consensus_timeout_params(`). The correst formula to be used here is: `consensus_timeout_ms * 2u64.pow((view_difference + 1 as u32).saturating_sub(2));`

DrZoltanFazekas: Maybe move the 3 lines from above, setting `view_difference`, `consensus_timeout_ms` and `next_exponential_backoff_timeout` right above the `!info()` statement below, since the 3 variables are only used there

DrZoltanFazekas (changes requested):

zilliqa/src/consensus.rs
JamesHinshelwood: Good points, thank you.

DrZoltanFazekas (approved):

PR #1423, opened 2024-09-10
fix: _balance always returns zero in scilla transition calls.

saeed-zil:


zilliqa/src/scilla.rs
```diff
@@ -275,7 +275,7 @@ impl Scilla {
             "-gaslimit".to_owned(),
             gas_limit.to_string(),
             "-balance".to_owned(),
-            value.to_string(),
+            contract_balance.to_string(),
```
JamesHinshelwood, line 278: Woops, my mistake for mixing this up. Can we double check the OCaml source to confirm for certain that it is expecting the txn sender's balance here?

saeed-zil: The related test passes after this change, but surely I'll double check.

saeed-zil: @JamesHinshelwood It expects to receive the calling contract's balance here. Here's the zq1 code:

https://github.com/Zilliqa/Zilliqa/blob/f0bcbb017b18e5f16f82321374b11cfc25504882/src/libData/AccountStore/AccountStoreSC.cpp#L746

saeed-zil: and here:
https://github.com/Zilliqa/Zilliqa/blob/f0bcbb017b18e5f16f82321374b11cfc25504882/src/libData/AccountStore/AccountStoreSC.cpp#L1620

The txn sender's balance is included in the `message`:

```rust
                let next_message = json!({
                    "_tag": message.tag,
                    "_amount": message.amount.to_string(),
                    "_sender": format!("{to_addr:#x}"),
                    "_origin": format!("{from_addr:#x}"),
                    "params": message.params,
                });
```

saeed-zil: and it's sent to `run` as `-imessage`:

```rust
        let args = vec![
            "-init".to_owned(),
            serde_json::to_string(&init)?,
            "-ipcaddress".to_owned(),
            self.state_server_addr(),
            "-imessage".to_owned(),
            serde_json::to_string(&msg)?,
            "-i".to_owned(),
            code.to_owned(),
            "-gaslimit".to_owned(),
            gas_limit.to_string(),
            "-balance".to_owned(),
            contract_balance.to_string(),
            "-libdir".to_owned(),
            self.lib_dir.clone(),
            "-jsonerrors".to_owned(),
            "-pplit".to_owned(),
            "true".to_owned(),
        ];
 ```

File: evm_scilla_js_tests/contracts/scilla/SendZil.scilla [hunk 2]
```diff
@@ -36,10 +36,12 @@ transition fundUser (user : ByStr20, amount : Uint128)
   send msgs
 end
 
-transition fundContract (contract_address : ByStr20, amount : Uint128)
-  msg = { _tag : "acceptZil" ; _recipient : contract_address ; _amount : amount };
+transition fundContracts (contract_address1 : ByStr20, amount1 : Uint128, contract_address2 : ByStr20, amount2 : Uint128)
+  msg1 = { _tag : "acceptZil" ; _recipient : contract_address1 ; _amount : amount1 };
+  msg2 = { _tag : "acceptZil" ; _recipient : contract_address2 ; _amount : amount2 };
```
JamesHinshelwood, line 41: Do we also have a test for sending to a contract which doesn't have a corresponding transition? (in which case the amount should not be sent?)

JamesHinshelwood (dismissed):

evm_scilla_js_tests/contracts/scilla/SendZil.scilla
saeed-zil: No, We didn't. But now we do :D

JamesHinshelwood (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Unreleased changes, in reverse chronological order. New entries are added at the
 - [#1359](https://github.com/Zilliqa/zq2/pull/1359): Add tool to convert ZQ1 persistence into a ZQ2 data directory.
 - [#1419](https://github.com/Zilliqa/zq2/pull/1419): Nodes outside the consensus commitee will no longer vote for a new view if a proposal is missed.
 - [#1393](https://github.com/Zilliqa/zq2/pull/1393): Begin assembling a block proposal before a supermajority of votes is received.
+- [#1423](https://github.com/Zilliqa/zq2/pull/1423): Fix invalid value returned by calling `_balance` in a scilla transition.
```
JamesHinshelwood, line 13: Please add this under `Unreleased`

JamesHinshelwood (changes requested):

CHANGELOG.md
saeed-zil: Oh, my bad! I didn't see that date!

PR #1433, opened 2024-09-11
Extended logging to track missing txns

bzawisto:
As title says.

### DIFF
--- a/zilliqa/src/pool.rs
+++ b/zilliqa/src/pool.rs
@@ -213,7 +213,7 @@ impl TransactionPool {
             // keeping the same nonce. So for those, it will always discard the new (identical)
             // one.
             if ReadyItem::from(existing_txn) >= ReadyItem::from(&txn) {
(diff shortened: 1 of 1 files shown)


zilliqa/src/pool.rs
```diff
@@ -211,6 +213,7 @@ impl TransactionPool {
             // keeping the same nonce. So for those, it will always discard the new (identical)
             // one.
             if ReadyItem::from(existing_txn) >= ReadyItem::from(&txn) {
+                debug!("Received txn with the same nonce but lower gas. Txn hash: {:?}, from: {:?}, nonce: {:?}, gas_limit: {:?}", txn.hash, txn.signer, txn.tx.nonce(), txn.tx.gas_limit());
                 return false;
```
DrZoltanFazekas, line 217: I guess you wanted to log the gas price and not the gas limit. The higher gas price makes a transaction with the same nonce better, not the higher gas limit...

DrZoltanFazekas (changes requested):

PR #1439, opened 2024-09-12
Support importing and calling external libraries in scilla transitions.

saeed-zil:


saeed-zil:
After merging https://github.com/Zilliqa/zq2/pull/1423 this PR should be changed to merge to `main` instead of `fix-balance` branch.

zilliqa/src/exec.rs
```diff
@@ -1368,11 +1418,18 @@ fn scilla_create(
         ));
     };
 
+    let init_data = ContractInit::new(init_data)?;
+
+    // We need to cache external libraries used in the current contract. Scilla checker needs to import them to check the contract.
```
JamesHinshelwood, line 1423: If I understand correctly - All the executed external libraries will live persistently in `ext_libs_cache_dir`?

Do we need to do this? What if instead we didn't cache anything and just populated the directory with the specific libraries that are required when we are running the transaction? 

I'd like to avoid adding caches if we can avoid it, since it is extra complexity and room for more bugs to appear (e.g. outdated contracts in the cache, name collisions, etc.)

JamesHinshelwood (commented):
LGTM mostly - A few points worth discussing.

zilliqa/src/cfg.rs
```diff
@@ -260,10 +269,17 @@ pub fn scilla_address_default() -> String {
...
 
-pub fn scilla_lib_dir_default() -> String {
+pub fn scilla_stdlib_dir_default() -> String {
     String::from("/scilla/0/_build/default/src/stdlib/")
 }
 
+pub fn scilla_ext_libs_cache_folder_default() -> ScillaExtLibsCacheFolder {
+    ScillaExtLibsCacheFolder {
+        on_host: String::from("/tmp"),
+        on_docker: String::from("/scilla_ext_libs"),
```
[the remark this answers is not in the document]
rrw-zilliqa: Because the containerised version of Scilla needs to know where to look for libraries and its filesystem looks different to the host's.

(and z2 does indeed run Scilla natively - once this PR is merged, I'll add a hack - because obviously every node needs a different directory or they will conflict).

zilliqa/src/exec.rs
rrw-zilliqa: On the other hand, doing so might lead to significant slowdowns in execution speed (admittedly, not as significant as executing Scilla itself!)

zilliqa/src/exec.rs
```diff
@@ -1323,13 +1328,49 @@ impl From<Account> for PendingAccount {
...
+            } => {
+                let contract_init = ContractInit::new(serde_json::from_str(init_data)?)?;
+                if !contract_init.is_library()? {
+                    return Err(anyhow!(
+                        "Impossible to load a non-library contract as a Scilla library."
+                    ));
+                }
+
+                let file_path = ext_libs_path.join(&lib.name);
+                fs::write(&file_path, code).with_context(|| {
```
rrw-zilliqa, line 1357: FWIW, this code:

 - will race against anything else running at the time (I think we're safe, though, since there is only one other thing that could be - the scilla interpreter - and that isn't running when this code is executed). Might be worth a comment though.
 -  Doesn't atomically write the file (again, I think we're safe-ish unless disk space runs out or something).
 - Allows a user who can create a library with the appropriate name (`../../../fish/soup.txt`) to overwrite any file on the node's filesystem to which the node's account has access.

I'd think it was worth a comment on why the last one (in particular!) isn't a threat ...

zilliqa/src/exec.rs
JamesHinshelwood: I'd prefer we implement the simple version first and add a cache if we prove that loading a contract's libraries is a bottleneck. My hunch is that it won't be, since we already need to spend a bunch of time writing other files (such as the contract we're actually executing) at transaction execution time anyway.

saeed-zil: @JamesHinshelwood The name might be misleading. My intention is not to cache a library contract for future transactions, except for the current one. I don’t verify the existence of the external library file before writing it to the cache directory, ensuring the code is always up-to-date. To address this, I can remove the external library contract files after each transaction execution.

@rrw-zilliqa To be stateless, we need to touch the Scilla HTTP server too, don’t we?

zilliqa/src/cfg.rs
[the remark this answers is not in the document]
saeed-zil: I believe the best approach is the one suggested by @rrw-zilliqa: adopting a stateless design. This involves sending the external libraries to the Scilla HTTP server along with the contract’s code. However, for now, we're not going to touch the Scilla server so we need to inform the Scilla server where to locate the external libraries. We can’t use the same configuration for both the dockerized version of the Scilla server and the one executed locally on the zq2 server.

zilliqa/src/state.rs
```diff
@@ -300,6 +307,108 @@ impl Default for Account {
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct ScillaTypedVariable {
+    pub vname: String,
+    pub value: ScillaVariableValue,
+    pub r#type: String,
+}
+
+#[derive(Serialize, Debug, Clone, Deserialize)]
+#[serde(untagged)]
+pub enum ScillaVariableValue {
```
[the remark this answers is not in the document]
saeed-zil: Yes, but this code:

zilliqa/src/exec.rs
rrw-zilliqa: @saeed-zil  - yes, sadly.

If it were me, I'd zap the whole directory after every contract execution - in fact, I'd probably go one further and create a new directory for each contract execution, and then clean them up after myself - this allows you to have a debugging "keep" flag so that you can see what library was executed with each contract (which is important because you might have a single txn with several contract executions each of which uses a different library with the same name).

It also means that your cleanup code is `tokio::fs::remove_dir()` , which is safer than writing it yourself (and saves you worrying about dotfiles, illegal filenames, half-written files which were interrupted by a node crash, etc. etc.)

zilliqa/src/exec.rs
```diff
@@ -4,13 +4,14 @@ use std::{
     collections::{hash_map::Entry, BTreeMap, HashMap},
     error::Error,
     fmt::{self, Display, Formatter},
-    mem,
+    fs, mem,
```
rrw-zilliqa, line 7: Sure you want to use `std::fs` and not `tokio::fs` ? Will block processing if you use blocking ops from `std`

zilliqa/src/exec.rs
```diff
@@ -1368,11 +1424,18 @@ fn scilla_create(
...
+    cache_external_libraries(
+        &state.pre_state,
+        ext_libs_cache_dir,
+        &init_data.external_libraries()?,
+    )?;
     let check_output = match scilla.check_contract(&txn.code, gas, &init_data)? {
         Ok(o) => o,
         Err(e) => {
             warn!(?e, "transaction failed");
-            let gas = gas.min(e.gas_remaining);
```
rrw-zilliqa, line 1375: This has the effect of not charging for transactions which fail to check . I think we ought to charge for this ? @JamesHinshelwood ?

zilliqa/src/exec.rs
```diff
@@ -1323,13 +1327,56 @@ impl From<Account> for PendingAccount {
...
+}
+
 fn scilla_create(
     mut state: PendingState,
     scilla: MutexGuard<'_, Scilla>,
     from_addr: Address,
     txn: TxZilliqa,
     current_block: BlockHeader,
     mut inspector: impl ScillaInspector,
+    ext_libs_cache_dir: &str,
```
rrw-zilliqa, line 1379: How are external libraries cleaned up after execution?

rrw-zilliqa (commented):
Afraid I have a question about path canonicalisation and another about cleanup ,and might be nice if you could comment on the gas issue too? - other than that, happy to approve ...

zilliqa/src/exec.rs
saeed-zil: The problem is if the check fails, the returned error, `e`, doesn't have gas_remaining:

```
CheckError { errors: [Error { start_location: Location { line: 2 }, error_message: "Ends in an error in state: 343.\n" }] }
```

saeed-zil: We need to charge a fixed value.

zilliqa/src/exec.rs
saeed-zil: I need to make all the functions in the call hierarchy `async` 🥲

saeed-zil:
Recently, I realized that we need to cache external libraries not only for contract check but also for contract deployment and transition execution

zilliqa/src/exec.rs
saeed-zil: I had to use `scopeguard` crate to ensure they'll be deleted even in case of an error.

zilliqa/src/exec.rs
JamesHinshelwood: Agree with @rrw-zilliqa - This code used to charge gas in the `let gas = gas.min(e.gas_remaining);` line and now we don't - Why? Clearly the Scilla server used to return `gas_remaining` if that code worked, so what changed?

rrw-zilliqa (dismissed):
Looks good to me :-)

I'll submit a follow-up for z2

zilliqa/src/exec.rs
saeed-zil: As I said, if the check fails the return value doesn't contain the gas_remaining. I'll revert my changes for now and will file another issue.

rrw-zilliqa (approved):
Still looks good to me :-)

Issue #1056, opened 2024-06-18
Ensure database can't be left in an inconsistent state


JamesHinshelwood:
We insert a block, then its transaction receipts. Sometimes that can leave the DB in an inconsistent state. Either use a transaction for both or move the block insert after.

Hopefully addressed by:

- [x] #1445 
- [x] #1449 
- [x] #1469

JamesHinshelwood:
When restarting one of the proto-testnet nodes we observed another DB consistency problem:

```
thread 'tokio-runtime-worker' panicked at zilliqa/src/consensus.rs:1998:52:
called `Result::unwrap()` on an `Err` value: database error: trie error: missing node

Caused by:
    trie error: missing node

Stack backtrace:
   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at ./usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.86/src/backtrace.rs:27:14
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at ./rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/result.rs:1964:27
   2: zilliqa::exec::<impl zilliqa::state::State>::apply_transaction_evm
             at ./zilliqa/zilliqa/src/exec.rs:457:17
   3: zilliqa::exec::<impl zilliqa::state::State>::call_contract
             at ./zilliqa/zilliqa/src/exec.rs:945:51
   4: zilliqa::exec::<impl zilliqa::state::State>::leader_raw
             at ./zilliqa/zilliqa/src/exec.rs:657:22
   5: zilliqa::consensus::Consensus::leader_at_block_raw
             at ./zilliqa/zilliqa/src/consensus.rs:1998:26
   6: zilliqa::consensus::Consensus::leader_at_block
             at ./zilliqa/zilliqa/src/consensus.rs:1982:9
   7: zilliqa::consensus::Consensus::check_block
             at ./zilliqa/zilliqa/src/consensus.rs:1614:24
   8: zilliqa::consensus::Consensus::proposal
             at ./zilliqa/zilliqa/src/consensus.rs:498:15
   9: zilliqa::consensus::Consensus::receive_block
             at ./zilliqa/zilliqa/src/consensus.rs:1748:22
  10: zilliqa::node::Node::handle_block_response
[...]

PR #1449, opened 2024-09-12
Store a block's canonicality directly in the `blocks` table

JamesHinshelwood:
One of the causes of #1056 was that we could add a block to `blocks` but not add the `number` -> `hash` mapping in `main_chain_canonical_blocks`. This meant the block existed if you looked it up via the hash or view, but didn't if looked up via its number.

We could fix this by wrapping both mutations in a database operation, but I think its cleaner to just add an `is_canonical` column to `blocks`. We maintain the invariant that, for all canonical blocks, the block `number` is unique. This is easy to maintain because blocks always start as canonical and only ever become non-canonical once (i.e. a block can't become canonical again later).


zilliqa/src/db.rs
```diff
@@ -226,10 +231,8 @@ impl Db {
                 gas_used INTEGER NOT NULL,
                 gas_limit INTEGER NOT NULL,
                 qc BLOB NOT NULL,
-                agg BLOB);
-            CREATE TABLE IF NOT EXISTS main_chain_canonical_blocks (
-                height INTEGER NOT NULL PRIMARY KEY,
-                block_hash TEXT NOT NULL REFERENCES blocks (block_hash));
+                agg BLOB,
+                is_canonical BOOLEAN NOT NULL);
```
JamesHinshelwood, line 235: Don't merge this yet - It makes a backwards-incompatible change to the database which we don't have good support for yet (e.g. DB migrations). We should discuss how to deal with this.

rrw-zilliqa (approved):
Looks good to me :-)

PR #1451, opened 2024-09-13
Include failed transaction in a block

bzawisto:
Failed transaction are now included in blocks. Additionally, balance checks shouldn't be done during gas estimation.

### DIFF
--- a/zilliqa/src/exec.rs
+++ b/zilliqa/src/exec.rs
@@ -523,7 +523,7 @@ impl State {
                 };
                 let code = match code {
                     EvmCode(code) => mem::take(code),
-                    _ => bail!("Evm account contains on-evm type of code!"),
+                    _ => bail!("Evm account contains non-evm code type!"),
                 };
 
                 // The only updated fields are nonce and balance, rest can be set to default (means: no update)


zilliqa/src/consensus.rs
```diff
@@ -1211,43 +1201,72 @@ impl Consensus {
...
+                                    tx.hash,
+                                    tx_index_in_block,
+                                    self.config.consensus.eth_block_gas_limit - gas_left,
+                                )),
+                            )
+                        } else {
+                            (true, None)
+                        }
+                    }
+                    Err(_) => {
```
JamesHinshelwood, line 1233: I'm not convinced this branch makes sense. If `apply_transaction_at` fails for some catastrophic reason like the DB being inaccessible, then I don't think it makes sense to add add a failed transaction receipt - We should just propagate the error up and eventually panic.

If application-level errors (e.g. not enough funds) and system-level errors are both getting muddled into the `Err` half of the `Result` of `apply_transaction_at` then I think we should fix that lower down the stack. Or in other words, `State::apply_transaction` should always return `Ok(TransactionApplyResult)` if we expect it to produce a mined transaction and a receipt.

In theory I think we could make this change by matching on the result of `evm.transact()` and mapping some of the `EvmError` variants to `Ok(<failed transaction receipt>)` and keep others as an `Err`?

zilliqa/src/exec.rs
```diff
@@ -497,9 +496,66 @@ impl State {
...
+                    EvmCode(_) => B256::ZERO,
+                    _ => KECCAK_EMPTY,
+                };
+                let code = match code {
+                    EvmCode(code) => mem::take(code),
+                    _ => bail!("Evm account contains non-evm code type!"),
+                };
+
+                // The only updated fields are nonce and balance, rest can be set to default (means: no update)
+                let affected_acc = revm::primitives::Account {
```
JamesHinshelwood, line 530: This feels like we are re-implementing things that should really be handled inside `revm`. I'm quite worried about doing this because when we did it before (when we depended on sputnik) we got lots of the edge cases wrong.

Just to summarise my understanding of the root issue here - The problem is that transactions get dropped from the transaction pool if we try to execute them and after executing, realise that there is not enough space left in the block. Could we not instead solve that issue by being more strict on the condition by which we consider executing a transaction - Instead of (optimistically) executing the next transaction and checking it fits, what if we checked the maximum gas spend of a transaction before executing it and stop building the block if it doesn't fit. That way, we only execute a transaction once we are certain it can be included in the block. `reth` does this and I don't think its a problem from the user's perspective if their transaction with an overly large gas limit gets included one block later than it theoretically could have.

Am I making sense? Happy to have a chat to discuss this instead if you prefer.

cc @DrZoltanFazekas

DrZoltanFazekas: I preproduced the case on the Holesky testnet, which we can consider a "reference" of correct behaviour. First I reduced the balance of my account `0xd819ffce7a58b1e835c25617db7b46a00888b013` to `0.000008323450116360 ether`. Then I sent 10 simple transfer transactions knowing that after 5 of them my account will run out of gas. The first 5 transactions were mined in the same block, you can see them as the 5 topmost transactions in the list: https://holesky.etherscan.io/address/0xd819ffce7a58b1e835c25617db7b46a00888b013. I also got the receipts for those 5 transactions. Then my script got stuck waiting for further receipts and timed out minutes later with:
```
AggregateError: 
    at internalConnectMultiple (node:net:1118:18)
    at internalConnectMultiple (node:net:1186:5)
    at Timeout.internalConnectMultipleTimeout (node:net:1712:5)
    at listOnTimeout (node:internal/timers:583:11)
    at processTimers (node:internal/timers:519:7)
```

In the Holesky RPC node's mempool there was no `pending` or `queued` transaction from my account, so the other 5 transactions for which there wasn't enough balance left must have been dropped. 

Note the difference between Holesky and our prototestnet: on the prototestnet the validators would drop the other 5 transactions, but they would stay in the API nodes' mempool as `pending` (forever).

Issue #1322, opened 2024-08-19
Block time increases with the number of txns regardless of how much gas consumed


DrZoltanFazekas:
In case of larger transactions with 1.3m gas each, we can fit only 65 of them in a block. The average block time increases to 4sec. Examples in blocks `23,487` to `23,501`.

Somewhat smaller transactions consuming 516k gas each allow us to squeeze up to 150 txns in a block, but the block time increases to 5-6 sec. See blocks `12,107` to `12,111`. 

When I further decreased the gas used per txn to 165k, we squeezed up to 509 txns into a block, but ended up with a block time of 12 sec, e.g. block `12,905` and `12,906`, or a timeout:
```
  view height timestmp                                   proposer      gas txn highqc
-------------------------------------------------------------------------------------
 12903  12903 10:52:25 0x58a7b212d07838bb466e6fd88b17cc97505397e1 34588470 210  12902 5s since last block
 12904         timeout 0xc6f65e983f921dfea53e2a11659a3c2347aa530f    after   5s
 12905  12903 10:52:35 0x58a7b212d07838bb466e6fd88b17cc97505397e1        0   0  12902 10s since last block
```

Investigate

1. how a 12sec block time was possible given that the leader should stop adding further transactions to the block to be proposed when it gets close to the timeout: https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/consensus.rs#L1105-L1106

1. why the block time increases so dramatically with the number of txns even though the total gas consumed is the same, compare e.g. block `23,487` (next block after 3 sec) with block `12,904` (next block after 12 sec), both consuming 84m gas, but the first one containing only 65 txns while the second one 509 txns.

Both small and large transactions (in terms of gas consumption) are doing the same - computing the Fibonacci sequence and storing it in a mapping - the only difference is the number of iterations (the `num` argument):
```
contract Fibonacci {
	uint256 public constant MAX = 2 ** 255;
	uint256 public index;
	mapping(uint256 => uint256) public fib;
	function compute(uint256 num) public {
		fib[0] = 0;
		fib[1] = 1;
		for(uint i = 2; i <= num; i++) {
			fib[i] = fib[i-1] + fib[i-2];
			index = i > index ? i : index;
			// we continue computing Fibonacci numbers up to
			// num, unless we know the next number will overflow,
			// in which case of which we start over  
			if (fib[i] >= MAX) {
				compute(num - i);
				break;
			}
		}	
	}
}
```

[...]

PR #1464, opened 2024-09-17
Do Not Clone Txn Pool

shawn-zil:


zilliqa/src/consensus.rs
```diff
@@ -155,7 +150,7 @@ pub struct Consensus {
     /// The persistence database
     db: Arc<Db>,
     /// Actions that act on newly created blocks
-    transaction_pool: TransactionPool,
+    transaction_pool: RefCell<TransactionPool>,
```
JamesHinshelwood, line 153: `RefCell<TransactionPool>` seems like a bit of a smell to me. If two people call `get_pending_block` at the same time (which the `&self` signature suggests should be possible), the `borrow_mut` call will panic as both try to take exclusive access over the `transaction_pool`. Its probably true that this doesn't happen in practice now, since the API layer accesses `Consensus` through a `Mutex`, but relying on this fact for correctness feels fragile. Its probably not far in our future that we need to move the locking further down the call-stack to improve API performance, at which point this assumption would break and the compiler won't tell us about the impending panics.

I guess our fundamental problem is that to serve 'pending' blocks we need to mutate the transaction pool to construct that pending block. Theoretically I can see a few solutions:
* Clone the stuff we need to mutate (what we do in `main` currently)
* Use a `RefCell`/`Mutex` to check/enforce exclusive access to the pending block (what this PR does)
* Change a bunch of signatures from `&self` to `&mut self` to enforce exclusive access to the pending block.
* Change when we assemble the 'pending' block - e.g. if we filled the block with 'ready' transactions as soon as those transactions arrived, theoretically we wouldn't need to mutate anything when reading the pending block.
* Change `TransactionPool`'s data structures to let it return the list of ready transactions without a `&mut self`. I haven't thought exactly how we could do this, but there is nothing special about the current `BinaryHeap` it uses for transactions and I could imagine a sorted `Vec` working just as well?

Is my understanding correct?

shawn-zil: 1. Cloning the stuff we need - yes, it would move the issue to the `get_pending_block` call e.g. when that end-point gets spammed. It is also why the current PR is written the odd way it is i.e. mutate and return/restore the txn pool i.e. trading memory with compute.

2. Yes.

3. It was the first thing that I tried, but the impact of this had a pretty big blast radius, touching lots of code, simply to allow the `get_pending_block` call to run. So, I decided to take a different route - inner mutation.

4. This is something that we could consider doing, but it would effectively require re-architecting the consensus process again.

5. The BinaryHeap uses an underlying Vec as-is. I could give this a try to see how well it works. Let me rethink this.

zilliqa/src/pool.rs
```diff
@@ -132,6 +132,43 @@ impl TransactionPool {
         }
     }
 
+    /// Returns a list of txns that are pending for inclusion in the next block
+    pub fn pending_hashes(&self) -> Vec<Hash> {
+        let mut pending_hash = Vec::<Hash>::new();
+        let mut pending_set = HashSet::new();
+        let mut ready = self.ready.clone();
```
JamesHinshelwood, line 139: Haven't we kept the transaction pool cloning here but hidden it away? @bzawisto might have a good point that if we use a different data structure here we can remove the clone (as ZQ1 seems to manage?)?

shawn-zil: In this case, it's just cloning the ready heap, which is relatively small as compared to the actual list of transactions. But if the consensus is to rewrite the entire pool using a different data structure, then I'll take a look into that.

JamesHinshelwood: If @bzawisto doesn't disagree, I think it would be fine to merge this for now and look at changing the data structure afterwards?

shawn-zil: That's good too. @bzawisto if you have no objections, could you approve this PR?

shawn-zil:
Also impacts #1322

Issue #1364, opened 2024-08-28
Checkpoints break the syncing algorithm in `BlockStore`


JamesHinshelwood:
Currently block store keeps a map of information about each peer:
https://github.com/Zilliqa/zq2/blob/e5b4d088efaff6e565569e90ca2aa8c5bc8e1d84/zilliqa/src/block_store.rs#L39

This information includes the greatest view of any block that we've seen from the peer:
https://github.com/Zilliqa/zq2/blob/e5b4d088efaff6e565569e90ca2aa8c5bc8e1d84/zilliqa/src/block_store.rs#L63

Currently this is only filled when we receive a proposal from a peer, but in future we want to add extra network messages by which peers can share their greatest view:
https://github.com/Zilliqa/zq2/blob/e5b4d088efaff6e565569e90ca2aa8c5bc8e1d84/zilliqa/src/block_store.rs#L262

We use this information to avoid making requests to peers for blocks they don't actually have:
https://github.com/Zilliqa/zq2/blob/e5b4d088efaff6e565569e90ca2aa8c5bc8e1d84/zilliqa/src/block_store.rs#L129

However, now we have checkpoints which mean nodes can be started from a checkpoint at block `n`, rather than always syncing from genesis. Validators can be started from checkpoints and propose blocks at views greater than `n`. This makes our assumption that a node who proposed `m` can provide us with blocks `0..=m` no longer true.

So we need to fix our block syncing algorithm to deal with this weaker property. I don't know how yet, lets discuss :)

PR #1494, opened 2024-09-23
Persistent networks and checkpoint support in z2

rrw-zilliqa:
(feat) The ability to specify network shapes, save them across z2 runs, and start them in groups.

(feat) The ability to load checkpoints from z2

Relates to #1364


z2/src/setup.rs
```diff
@@ -48,73 +65,164 @@ pub struct Setup {
...
+            } else {
+                Config::from_spec(val, base_port)?
+            }
+        } else if let Some(val2) = loaded_config {
+            println!("Starting previously saved config in {config_dir}");
+            val2
+        } else {
+            // Set up a default network
+            println!(">> No network specified or loaded; using default 4-node network for legacy reasons.");
+            //config = loaded_config.ok_or(anyhow!(
```
86667, line 183: can be removed?

z2/src/node_spec.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub is_validator: bool,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+pub struct Composition {
+    pub nodes: HashMap<u64, NodeDesc>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+pub struct NodeSpec {
```
86667, line 19: I wonder if `NetworkSpec` or `NodesSpec` is more clear here

rrw-zilliqa: I'll leave that to the language lawyers - when those who care enough have decided what they want, very happy to implement it :-)

z2/src/setup.rs
86667: nope, i was talking about the commented out code :+1:

86667:
Nice! Will be handy.

It works well on initialising a new network but has trouble when a config already exists. I get:

```
WARNING: You've specified a network configuration; we'll ignore this and take the config from the existing loaded configuration file
```

but it does later attempt to read a node that doesnt exist, so the nodespec config is infact being used as well as the current config. 

I will approve anyway as its an improvement nevertheless

rrw-zilliqa:
> Nice! Will be handy.
> 
> It works well on initialising a new network but has trouble when a config already exists. I get:
> 
> ```
> WARNING: You've specified a network configuration; we'll ignore this and take the config from the existing loaded configuration file
> ```
> 
> but it does later attempt to read a node that doesnt exist, so the nodespec config is infact being used as well as the current config.
> 
> I will approve anyway as its an improvement nevertheless (once its out of draft)

Thanks! I'll rerequest when the clilppy checks have passed - in the meantime, can  you give me a set of commands to reproduce your issue and I'll fix it? It seems like something I ought to fix ...

86667 (approved):

PR #1501, opened 2024-09-24
Add zero account as source/sink of rewards/gas respectively

86667:
This PR implements the burning mechanism described in [ZIP-9](https://github.com/Zilliqa/ZIP/blob/master/zips/zip-9.md).

- Added a new env var `total_native_token_supply` which defines the total amount of native token in the network. Stake deposits and genesis accounts (defined in `genesis_deposits` and `genesis_accounts` respectively) are taken from this `total_native_token_supply`  value and the remaining amount is assigned to the zero address (`0x0`)
- Rewards are removed from the zero account balance after being added to validator's balances at block execution
- All transaction gas fees are added back to the zero account as a burn mechanism


zilliqa/src/consensus.rs
```diff
@@ -695,20 +695,6 @@ impl Consensus {
     /// Note that the algorithm below is mentioned in cfg.rs - if you change the way
     /// rewards are calculated, please change the comments in the configuration structure there.
     fn apply_rewards(
-        &mut self,
```
86667, line 698: Refactor here removing redundent fn. Removed `apply_rewards()` and renamed `apply_rewards_raw()` to `apply_rewards()`

z2/src/docgen.rs
```diff
@@ -384,6 +385,7 @@ pub fn get_implemented_jsonrpc_methods() -> Result<HashMap<ApiMethod, PageStatus
             scilla_address: scilla_address_default(),
             blocks_per_epoch: 3600,
             epochs_per_checkpoint: 24,
+            total_native_token_supply: total_native_token_supply_default(),
```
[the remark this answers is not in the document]
JamesHinshelwood: I'm happy with "native token" :+1:

zilliqa/src/cfg.rs
```diff
@@ -188,6 +192,20 @@ impl<'de> Deserialize<'de> for Amount {
...
+        Amount::from(*self + *other)
+    }
+}
+
+impl Sub for Amount {
+    type Output = Self;
+    fn sub(self, other: Self) -> Self {
+        Amount::from(*self - *other)
+    }
+}
```
JamesHinshelwood, line 207: Maybe a bit of a nitpick, but I think I'd prefer if we just 'unwrapped' the `Amount` in `state.rs` straight away to get a `u128` and did the computation on that. In my head `Amount` is an awkward workaround because the TOML crate we use doesn't support large integers and it should only exist at the outer-most layer.

I'm not massively opinionated though, so feel free to ignore me if you disagree :)

JamesHinshelwood (dismissed):
LGTM, thanks! Feel free to address here or in a follow-up.

zilliqa/src/consensus.rs
```diff
@@ -767,21 +753,30 @@ impl Consensus {
...
+                let reward = (U256::from(rewards_per_block / 2) * U256::from(stake)
+                    / U256::from(total_cosigner_stake))
+                .to::<u128>();
+                at_state.mutate_account(cosigner, |a| a.balance += reward)?;
+                total_rewards_issued += reward;
             }
         }
 
+        // ZIP-9: Fund rewards amount from zero account
+        at_state.mutate_account(Address::ZERO, |a| a.balance -= total_rewards_issued)?;
```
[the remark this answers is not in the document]
86667: Yes good point. Updated to be safe on all usages of `mutate_account`

zilliqa/src/cfg.rs
86667: That makes sense. I wasnt sure about why the `Amount` struct was necessary. removed

JamesHinshelwood (approved):

Issue #1400, opened 2024-09-04
Validators ignore blocks less than their current view


JamesHinshelwood:
Currently, if a validator times out and increments `self.view` it will then ignore any block proposals received which are before `self.view`. It is correct that we shouldn't try to vote on proposals which are before our `self.view`, but we should still store them in case the rest of the network chooses that block as the canonical chain. We can maintain a condition of ignoring proposals if they are less than `self.finalised_view`, since these are blocks which are definitely not canonical.

rrw-zilliqa:
A natural way of doing this is to keep them in block_store's (upcoming!) block response cache, where all the other blocks which we don't recognise (yet) as being canonical sit.

86667:
Turns out that this issue fixes itself by re-requesting blocks which were previously rejected due to having views less than current. 

@rrw-zilliqa's [PR](https://github.com/Zilliqa/zq2/pull/1495) adding a block_cache provides a mechanism to resolve this inefficiency. I will pick this up again once it has been merged.

PR #1505, opened 2024-09-25
Store block in cache when view is smaller than current

86667:
Currently we reject blocks whose `view <= self.view`. This means that nodes will reject then re-request the missing block later on. 

Instead, here we add outdated blocks to the `block_store`'s cache to be used later or trimmed out.

Note that we still reject blocks whose `view <= self.finalised_view()`.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -647,7 +647,32 @@ impl Consensus {
         trace!("checking if block view {} is safe", block.view());
 
         // If the proposed block is safe, vote for it and advance to the next round.
-        if self.check_safe_block(&block, during_sync)? {
+        if self.check_safe_block(&block)? {
+            // If the proposed block is safe but outdated then add to block cache - we may need it later
+            let outdated = block.view() < self.view.get_view();
+            let process_immediately = !outdated || during_sync;
+            if !process_immediately {
+                trace!(
+                    "proposal is outdated: {} < {} but may be useful in the future, buffering",
+                    block.view(),
+                    self.view.get_view()
+                );
+                self.block_store.buffer_proposal(
+                    from,
+                    Proposal::from_parts_with_hashes(
+                        block,
+                        transactions
+                            .into_iter()
+                            .map(|tx| {
+                                let hash = tx.calculate_hash();
+                                (tx, hash)
+                            })
+                            .collect(),
+                    ),
+                )?;
+                return Ok(None);
+            }
+
             trace!(
                 "block view {} number {} aka {} is safe",
                 block.view(),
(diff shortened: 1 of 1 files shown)


86667:
Ive struggled to find a nice way to test this. Any suggestions?

rrw-zilliqa:
> Ive struggled to find a nice way to test this. Any suggestions?

I'd invent a debug API that sends proposals, then stop a node, erase its data, restart it and tell the debug API to tell another node to send an old proposal, which should now be stored - so you can then check in the db that it has been. Of course, with my #1364 changes you'd need to disable block fetching as well. Worth checking (and I think it's true) that insert_block() is idempotent, otherwise races will cause it to fail ...

86667:
This PR now has completely different changes to those reviewed before. The new changes are the outcome of this issue's findings - https://github.com/Zilliqa/zq2/issues/1400#issuecomment-2390887756

zilliqa/src/consensus.rs
```diff
@@ -647,7 +635,20 @@ impl Consensus {
...
+            // If the proposed block is safe but outdated then add to block cache - we may need it later
+            let outdated = block.view() < self.view.get_view();
+            let process_immediately = !outdated || during_sync;
+            if !process_immediately {
+                trace!(
+                    "proposal is outdated: {} < {} but may be useful in the future, buffering",
+                    block.view(),
+                    self.view.get_view()
+                );
+                self.block_store.buffer_proposal(from, proposal)?;
```
JamesHinshelwood, line 648: nit 
```suggestion
                self.block_store.buffer_proposal(from, Proposal::from_parts(block, transactions))?;
```
would avoid the clone at the start.

zilliqa/src/consensus.rs
```diff
@@ -1947,20 +1948,17 @@ impl Consensus {
...
     }
 
-    fn check_safe_block(&mut self, proposal: &Block, during_sync: bool) -> Result<bool> {
+    fn check_safe_block(&mut self, proposal: &Block) -> Result<bool> {
         let Some(qc_block) = self.get_block(&proposal.parent_hash())? else {
             trace!("could not get qc for block: {}", proposal.parent_hash());
             return Ok(false);
         };
-        // We don't vote on blocks older than our view
-        let outdated = proposal.view() < self.view.get_view();
```
JamesHinshelwood, line 1956: @DrZoltanFazekas Could you double-check these changes? I think they make sense but you are the expert :)

DrZoltanFazekas: At [https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/consensus.rs#L739-L742](https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/consensus.rs#L739-L742) we would vote for the outdated proposal and we shouldn't. Most likely we would send the `Vote` to the wrong leader as it's the leader of `self.view.get_view() + 1` but if it happens to be the same leader to whom we sent a `NewView` message when we timed out in `proposal.view()`, we get caught equivocating.

86667: thanks, it is a little tricky to follow. The aim here is to simply move the outdated check outside of the `check_safe_block()` fn

DrZoltanFazekas (commented):
@86667 I wonder if we ever get to the point of storing an outdated proposal since we reject it based on the condition at [https://github.com/Zilliqa/zq2/blob/1400-validators-ignore-blocks-less-than-their-current-view/zilliqa/src/consensus.rs#L594](https://github.com/Zilliqa/zq2/blob/1400-validators-ignore-blocks-less-than-their-current-view/zilliqa/src/consensus.rs#L594)

Instead of `block.view() <= head_block.header.view` maybe we should reject if `block.view() <= self.finalized_view`

86667:
> @86667 I wonder if we ever get to the point of storing an outdated proposal since we reject it based on the condition at https://github.com/Zilliqa/zq2/blob/1400-validators-ignore-blocks-less-than-their-current-view/zilliqa/src/consensus.rs#L594
> 
> Instead of `block.view() <= head_block.header.view` maybe we should reject if `block.view() <= self.finalized_view`

The particular changes here are for the case of our head being behind whilst proposals are arriving which are ahead of our head (our head < proposal) but behind our view (our view > proposal view). We will need these ahead proposals once we have eventually caught up our head.

Happy to look into rejecting blocks with `block.view() <= head_block.header.view`  but I think that is out of scope of this PR. I would want to understand the implications of storing many blocks at the same height on the `deal_with_fork()` code and block cache.

PR #1511, opened 2024-09-29
Add more API tests

maxconway:
The main intention here was to create an example test for testing transactions (`test_simulate_transactions`), since I thought something fishy was going on with the transaction tests. While I was at it I added a few more tests using the same methodology. I also identified that the strange behaviour I'd noted was due to 0 vs 1 based page numbers in `get_txns_for_tx_block_ex`, and I've now fixed the test to respect that.

- `get_txns_for_tx_block_0`
- `get_txns_for_tx_block_ex_1`
- `test_simulate_transactions`
- `get_tx_rate_1`
- `get_tx_block_rate_1`
- Some removals of duplicated asserts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 Unreleased changes, in reverse chronological order. New entries are added at the top of this list.
 
-- [#1511](https://github.com/Zilliqa/zq2/pull/1511): Add more API tests.
 - [#1449](https://github.com/Zilliqa/zq2/pull/1449): Restructure how blocks are stored to prevent database inconsistencies.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 Unreleased changes, in reverse chronological order. New entries are added at the top of this list.
 
+- [#1511](https://github.com/Zilliqa/zq2/pull/1511): Add more API tests.
```
JamesHinshelwood, line 7:
```suggestion
```
I've usually been leaving non-user-visible changes out of the changelog. But happy to do otherwise if others think we should track them.

JamesHinshelwood (approved):
LGTM

JamesHinshelwood (approved):

PR #1527, opened 2024-10-01
Store state trie in SQLite

JamesHinshelwood:
This means that writes of blocks and writes to state can be kept in sync, because the write barriers that SQLite provides will apply to both forms of data. This should eliminate inconsistencies between the two data stores.

Note that I also renamed the SQL database from `blockdata.db` to `db.sqlite3`.


shawn-zil (approved):

shawn-zil (dismissed):

86667:
Does this mean that a node running with this commit will have to start from genesis/checkpoint and rebuild its state?

86667 (approved):

PR #1530, opened 2024-10-01
Fix checkpoints: State at checkpoint should contain current block data as well as parent block data

### LINKED ISSUE #1430: Checkpoints should contain the state of the previous block
Current a checkpoint at block `x` contains:
* Serialised block `x`
* Serialised block `x-1`
* The state trie at block `x` (pointed to by the state root hash in block `x`)

However, this results in a state that isn't able to answer some questions about block `x`, such as who the proposer was. This is because the consensus committee for block `x` is defined by the state at block `x-1`.

So instead, we should store the state trie at block `x-1` and when loading a checkpoint, replay block `x` to construct a state which includes both `x-1` and `x`.

86667:
Closes #1430

### DIFF
--- a/zilliqa/src/db.rs
+++ b/zilliqa/src/db.rs
@@ -340,7 +340,7 @@ impl Db {
 
         if !trie_storage.db.is_empty() || self.get_highest_block_number()?.is_some() {
             // If checkpointed block already exists then assume checkpoint load already complete. Return None
-            if let Some(_) = self.get_block_by_hash(block.hash())? {
+            if self.get_block_by_hash(block.hash())?.is_some() {
                 return Ok(None);
             }
             // This may not be strictly necessary, as in theory old values will, at worst, be orphaned
@@ -400,7 +400,7 @@ impl Db {
 
         let parent_ref: &Block = &parent; // for moving into the closure
         self.with_sqlite_tx(move |tx| {
-            self.insert_block_with_db_tx(tx, &parent_ref)?;
+            self.insert_block_with_db_tx(tx, parent_ref)?;
             self.set_latest_finalized_view_with_db_tx(tx, parent_ref.view())?;
             self.set_high_qc_with_db_tx(tx, block.header.qc)?;
             Ok(())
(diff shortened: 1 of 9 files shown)


zilliqa/src/db.rs
```diff
@@ -271,12 +271,15 @@ impl Db {
...
+    /// Return None if checkpoint already loaded
     pub fn load_trusted_checkpoint<P: AsRef<Path>>(
         &self,
         path: P,
         hash: &Hash,
         our_shard_id: u64,
-    ) -> Result<Block> {
+    ) -> Result<Option<(Block, Vec<SignedTransaction>, Block)>> {
         // For now, only support a single version: you want to load the latest checkpoint, anyway.
         const SUPPORTED_VERSION: u32 = 1;
```
JamesHinshelwood, line 284: Shall we bump this to version 2 since we're now including transactions?

I don't care about backwards compatibility, but it might be helpful to ensure we don't try to load an older checkpoint with this new version of the code and interpret some other random bytes as transactions or similar.

zilliqa/tests/it/persistence.rs
```diff
@@ -235,11 +244,28 @@ async fn checkpoints_test(mut network: Network) {
...
+        .unwrap();
+    assert_eq!(block.transactions.len(), 1);
+
+    let block_from_checkpoint = new_node_wallet
+        .get_block(latest_block_number)
+        .await
+        .unwrap()
+        .unwrap();
+    assert_eq!(block.transactions, block_from_checkpoint.transactions);
+    // Check access to previous block state via fetching author of current block
```
JamesHinshelwood, line 265: Just double checking that this test fails in `main`?

JamesHinshelwood (approved):
LGTM, but I think we should update the version.

Also can we change `docs/checkpoints.md` to reflect the new checkpoint format?

zilliqa/tests/it/persistence.rs
86667: yes, the line

```rust
assert_eq!(block.author, block_from_checkpoint.author);
```
was the test harness as it fails for when state of parent doesnt exist in trie

zilliqa/src/db.rs
86667: cool updated

JamesHinshelwood (approved):
Thanks!

JamesHinshelwood (approved):

PR #1559, opened 2024-10-03
Implement and test all of the solidity block and transaction properties.

saeed-zil:


evm_scilla_js_tests/test/BlockProperties.ts
```diff
@@ -0,0 +1,65 @@
...
+    const contractChainId = await contract.getChainId();
+    const network = await ethers.provider.getNetwork();
+
+    expect(contractChainId).to.equal(network.chainId);
+  });
+
+  it("should return the correct coinbase", async function () {
+    const contractCoinbase = await contract.getCoinbase();
+
+    expect(contractCoinbase).to.equal("0x0000000000000000000000000000000000000000");
```
JamesHinshelwood, line 46: This is the incorrect coinbase. Can we either amend the test to state this isn't implemented yet or implement it?

evm_scilla_js_tests/test/BlockProperties.ts
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  it("should return the correct gas limit", async function () {
+    const contractGasLimit = await contract.getGasLimit();
+    const currentBlock = await ethers.provider.getBlock("latest");
+
+    expect(contractGasLimit).to.equal(currentBlock.gasLimit);
+  });
+
+  it("should emit blockhash correctly", async function () {
```
JamesHinshelwood, line 56: Can we add test cases for `blockhash` with the current block number and with a block number more than 256 blocks in the past? I believe both of these should return zero. Maybe a future block number test too for good measure?

evm_scilla_js_tests/test/BlockProperties.ts
```diff
@@ -0,0 +1,65 @@
...
+  it("should return the correct block number", async function () {
+    const contractBlockNumber = await contract.getBlockNumber();
+    const currentBlockNumber = await ethers.provider.getBlockNumber();
+
+    expect(contractBlockNumber).to.equal(currentBlockNumber);
+  });
+
+  it("should return the correct block timestamp", async function () {
+    const contractBlockTimestamp = await contract.getTimestamp();
+    const currentBlock = await ethers.provider.getBlock("latest");
```
JamesHinshelwood, line 25: Shouldn't we specify a fixed block number in all of these tests and pass the same one for both the `eth_call` and the `getBlock`? Otherwise, the network might move forward a block between these 2 lines causing flaky tests?

evm_scilla_js_tests/test/BlockProperties.ts
saeed-zil: test got disabled for now.

zilliqa/src/exec.rs
```diff
@@ -356,11 +356,19 @@ impl DatabaseRef for &State {
...
     fn block_hash_ref(&self, number: u64) -> Result<B256, Self::Error> {
-        Ok(self
-            .block_store
-            .get_block_by_number(number)?
-            .map(|block| B256::new(block.hash().0))
-            .unwrap_or_default())
+        if let Some(current_block_number) = self.block_store.get_highest_block_number()? {
+            match current_block_number.checked_sub(number) {
+                None => Ok(B256::default()), // current_block_number - number < 0
+                Some(diff) if diff > 256 => Ok(B256::default()), // current_block_number - number > 256
```
JamesHinshelwood, line 362: When I checked, it seemed like `revm` implemented this check already - Did you have to add this to get the test to pass or is it just an extra check?

JamesHinshelwood (approved):

zilliqa/src/exec.rs
```diff
@@ -356,11 +356,19 @@ impl DatabaseRef for &State {
     }
 
     fn block_hash_ref(&self, number: u64) -> Result<B256, Self::Error> {
-        Ok(self
-            .block_store
-            .get_block_by_number(number)?
-            .map(|block| B256::new(block.hash().0))
-            .unwrap_or_default())
+        if let Some(current_block_number) = self.block_store.get_highest_block_number()? {
```
JamesHinshelwood, line 359: We shouldn't use `get_highest_block_number` here. We might be executing a transaction on a much older block for (e.g.) replay.

Instead, we should use the block number passed into the `TxEnv`. I believe this is what `revm` does by default (see `fn block_hash` in `context.rs`), so maybe just trust it and remove the check entirely?

zilliqa/src/exec.rs
saeed-zil: The reason I added these checks was test failures. But let double check. Maybe I was wrong.

JamesHinshelwood (approved):

JamesHinshelwood:
Thanks!

PR #1563, opened 2024-10-04
Reduce time taken to collect VerifiedTransactions

shawn-zil:
The original `execute_block()` collected verified transactions transactions from the Proposal block by re-verifying each transaction in the Proposal block. This triggers the `recover_signer()` library call for each transaction in the Proposal. It then fills up any missing ones from the mempool by cloning.

This PR flips this behaviour around to collect verified transactions first from the mempool, which already contains pre-verified transactions. It then fills up any missing ones from the Proposal block by only re-verifying those missing from the mempool.

This speeds things up by several orders of magnitude, when the transactions are already pre-verified and present in the mempool, with the worst case scenario being similar in timing to the previous behaviour e.g.

before
```
zilliqanode1      | 2024-10-04T01:31:05.720473Z  INFO zilliqa::consensus: 2635: 95 ELAPSED_VER 323.269721ms
zilliqanode0      | 2024-10-04T01:31:05.730972Z  INFO zilliqa::consensus: 2635: 95 ELAPSED_VER 323.800224ms
zilliqanode3      | 2024-10-04T01:31:05.756535Z  INFO zilliqa::consensus: 2635: 95 ELAPSED_VER 323.695312ms
zilliqanode2      | 2024-10-04T01:31:05.763893Z  INFO zilliqa::consensus: 2635: 95 ELAPSED_VER 325.994903ms
```

after
```
zilliqanode1      | 2024-10-04T02:17:40.488820Z  INFO zilliqa::consensus: 2653: 96 ELAPSED_INS 192.081µs
zilliqanode2      | 2024-10-04T02:17:40.510038Z  INFO zilliqa::consensus: 2653: 96 ELAPSED_INS 197.46µs
zilliqanode3      | 2024-10-04T02:17:40.519616Z  INFO zilliqa::consensus: 2653: 96 ELAPSED_INS 208.291µs
zilliqanode0      | 2024-10-04T02:17:40.790467Z  INFO zilliqa::consensus: 2653: 96 ELAPSED_INS 327.814331ms
```

The time taken to collect about ~95 VerifiedTransactions is about 325ms across all nodes (before) and is about 200us (if already present in the pool) to ~325ms (if not present in the pool).

As this is a small change that makes a measurable impact, with the worst case similar as before, I thought it would be good to merge this one first.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -2632,24 +2632,28 @@ impl Consensus {
         let mut verified_txns = Vec::new();
 
         // We re-inject any missing Intershard transactions (or really, any missing
-        // transactions) from our mempool. If any txs are unavailable either in the
+        // transactions) from our mempool. If any txs are unavailable in both the
         // message or locally, the proposal cannot be applied
         let now = std::time::Instant::now();
         for (idx, tx_hash) in block.transactions.iter().enumerate() {
-            // Attempt to insert verified txns from pool.
-            if let Some(local_tx) = self.transaction_pool.pop_transaction(*tx_hash) {
-                verified_txns.insert(idx, local_tx);
-                continue;
+            // Prefer to insert verified txn from pool. This is faster.
+            let txn = match self.transaction_pool.pop_transaction(*tx_hash) {
+                Some(txn) if txn.hash == *tx_hash => txn,
+                _ => match transactions
+                    .get(idx)
+                    .and_then(|sig_tx| sig_tx.clone().verify().ok())
+                {
+                    // Otherwise, recover txn from proposal. This is slower.
+                    Some(txn) if txn.hash == *tx_hash => txn,
+                    _ => {
+                        warn!("Proposal {} at view {} referenced a transaction {} that was neither included in the broadcast nor found locally - cannot apply block", block.hash(), block.view(), tx_hash);
+                        return Ok(());
+                    }
+                },
             };
-            // And recover missing txns from block
-            if let Some(block_tx) = transactions.get(idx) {
-                verified_txns.insert(idx, block_tx.clone().verify()?);
-                continue;
-            };
-
-            warn!("Proposal {} at view {} referenced a transaction {} that was neither included in the broadcast nor found locally - cannot apply block", block.hash(), block.view(), tx_hash);
-            return Ok(());
+            verified_txns.insert(idx, txn);
         }
+
         info!("{} ELAPSED_INS {:?}", len, now.elapsed());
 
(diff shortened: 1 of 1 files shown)


zilliqa/src/consensus.rs
```diff
@@ -2627,29 +2627,35 @@ impl Consensus {
...
-            } else {
-                let Some(local_tx) = self.transaction_pool.pop_transaction(*tx_hash) else {
-                    warn!("Proposal {} at view {} referenced a transaction {} that was neither included in the broadcast nor found locally - cannot apply block", block.hash(), block.view(), tx_hash);
-                    return Ok(());
-                };
-                transactions.insert(idx, local_tx);
-            }
+            // Prefer to insert verified txn from pool. This is faster.
+            let txn = match self.transaction_pool.pop_transaction(*tx_hash) {
+                Some(txn) if txn.hash == *tx_hash => txn,
```
JamesHinshelwood, line 2638: Surely this `if` guard is unfalsifiable if `pop_transaction` has a sane implementation?

zilliqa/src/consensus.rs
```diff
@@ -2627,29 +2627,35 @@ impl Consensus {
...
+                _ => match transactions
+                    .get(idx)
+                    .and_then(|sig_tx| sig_tx.clone().verify().ok())
```
JamesHinshelwood, line 2641: This `.ok()` converts verification failures into "transaction not found failures" which may be confusing. Can we return an `Err` instead. `.transpose()?` might do the trick.

zilliqa/src/consensus.rs
```diff
@@ -2627,29 +2627,35 @@ impl Consensus {
...
+                {
+                    // Otherwise, recover txn from proposal. This is slower.
+                    Some(txn) if txn.hash == *tx_hash => txn,
+                    _ => {
+                        warn!("Proposal {} at view {} referenced a transaction {} that was neither included in the broadcast nor found locally - cannot apply block", block.hash(), block.view(), tx_hash);
+                        return Ok(());
+                    }
+                },
+            };
+            verified_txns.insert(idx, txn);
```
JamesHinshelwood, line 2651: If I understand correctly, its always true that `idx == verified_txns.len()` (i.e. we're always adding to the end of the list) which makes this equivalent to `verified_txns.push(txn);`. Can we just push and remove the `idx` to make things clearer?

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):

shawn-zil:
Also impacts #1322

PR #1572, opened 2024-10-07
Add `GetRecentTransactions` API.

maxconway:
- plus two tests
- First commit was named "Add GetNumDSBlocks API.", but that was a typo, this is really for GetRecentTransactions

### DIFF
--- a/zilliqa/src/api/zil.rs
+++ b/zilliqa/src/api/zil.rs
@@ -904,7 +904,8 @@ fn get_recent_transactions(
     let node = node.lock().unwrap();
     let mut block_number = node.get_chain_tip();
     let mut txns = Vec::new();
-    while block_number > 0 && txns.len() < 100 {
+    let mut blocks_searched = 0;
+    while block_number > 0 && txns.len() < 100 && blocks_searched < 100 {
         let block = match node
             .consensus
             .block_store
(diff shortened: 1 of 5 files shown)


zilliqa/src/api/zil.rs
```diff
@@ -894,3 +896,34 @@ fn get_num_ds_blocks(_params: Params, node: &Arc<Mutex<Node>>) -> Result<String>
...
 }
+
+fn get_recent_transactions(
+    _params: Params,
+    node: &Arc<Mutex<Node>>,
+) -> Result<RecentTransactionsResponse> {
+    let node = node.lock().unwrap();
+    let mut block_number = node.get_chain_tip();
+    let mut txns = Vec::new();
+    while block_number > 0 && txns.len() < 100 {
```
JamesHinshelwood, line 907: Can we add a limiter for the number of blocks too? I worry about a network with many empty blocks (like the current proto-testnet) in which this loop will take a long time to terminate. 100 blocks seems reasonable to me, with the justification that transactions aren't really 'recent' if they occurred ages ago?

JamesHinshelwood (approved):

zilliqa/src/api/zil.rs
maxconway: Yep sounds good will do

maxconway: Done. The docs are also updated to match

JamesHinshelwood (approved):

Issue #1218, opened 2024-07-26
Don't create blocks faster than 1sec


DrZoltanFazekas:
Currently, we wait for 1sec before we create an empty block if there is no transaction in the mempool that is ready to be included in the block. However, we don't wait in case there is a transaction and create the block immediately. If the next transaction appears in the mempool right after the block has been proposed, we create another block immediately. This can lead to situations where multiple blocks are created within 1sec. We want the leader to wait until the 1sec blocktime expires before they broadcast the proposal. Note that during this 1-sec period, we should immediately add transactions that show up in the mempool to the block in preparation, and not wait 1 sec and then start adding/executing transactions.

PR #1573, opened 2024-10-07
Do not immediately propose block upon receiving tx

86667:
Closes https://github.com/Zilliqa/zq2/issues/1218

The overall goal is to stop blocks being created and broadcast immediately upon receive of a transaction. 

- Create `early_proposal` block proposal as soon as any Votes arrive l
- As new transactions come in update `early_proposal` to include them
- Broadcast proposal only after `empty_block_timeout` seconds have passed since last view

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -1052,7 +1052,7 @@ impl Consensus {
                         .empty_block_timeout
                         .saturating_sub(Duration::from_millis(time_since_last_view_change)),
                 )?;
-                trace!("Empty transaction pool, will create new block on timeout");
+                trace!("will propose new block {} on timeout", self.view.get_view());
             }
         } else {
             self.votes.insert(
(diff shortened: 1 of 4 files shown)


zilliqa/src/consensus.rs
```diff
@@ -1037,36 +1042,33 @@ impl Consensus {
             if supermajority_reached {
                 // We propose new block immediately if there's something in mempool or it's the first view
                 // Otherwise the block will be proposed on timeout
-                if self.view.get_view() == 1 || self.transaction_pool.has_txn_ready() {
+                if self.view.get_view() == 1 {
                     return self.propose_new_block();
-                } else {
```
86667, line 1042: simple change here with large diff - remove else block

shawn-zil (commented):
This is a nice feature that should allow new transactions to be continuously added to the early proposal, until the deadline is reached. I would like to see how it impacts things when spammed with transactions. Either case, it also simplifies the code and makes it easier to understand.

However, it's currently failing CI. So, that should be fixed.

JamesHinshelwood (dismissed):
LGTM, thanks for simplifying :)

JamesHinshelwood (approved):

Issue #1502, opened 2024-09-24
Help plz: Invalid signature in simple transaction (golang)


PurrProof:
I've took CreateTransaction golang example from [here](https://dev.zilliqa.com/zilliqa2/api/apis/api/zilliqa/CreateTransaction/)

Here is slightly modified version with better logging, increased gas limit etc:
[main.txt](https://github.com/user-attachments/files/17118115/main.txt)

On the ZQ2 side there is an `invalid signature` error:

```
2024-09-24T17:05:55.222307Z  INFO zilliqa::node: 382: seen new txn Zilliqa { tx: TxZilliqa { chain_id: 1, nonce: 1, gas_price: ZilAmount(2000000016), gas_limit: ScillaGas(50000), to_addr: 0x4baf5fada8e5db92c3d3242618c5b47133ae003c, amount: ZilAmount(10000000), code: "", data: "" }, key: PublicKey { point: AffinePoint { x: FieldElement(FieldElementImpl { value: FieldElement5x52([3388391445012197, 1329098899749562, 1619017321969550, 2206526636033769, 217722622591469]), magnitude: 1, normalized: true }), y: FieldElement(FieldElementImpl { value: FieldElement5x52([1180503181878570, 1780274363765302, 1808606147638834, 3434437250924948, 29555431482941]), magnitude: 1, normalized: true }), infinity: 0 } }, sig: ecdsa::Signature<Secp256k1>(C85D8726480271D0264DBC05A2AAEA256BAD4D355112261D3A6B370244D825F03BE0C450BC38362AB2C0DF221EFE867C7A625380B390EDA59AD1D00B6BBBF6A2) } hash=bcc0c88f2fdd75fe4a0229129a35ef8fa75d9fb01962a4b05e6ee2d232b3506b
2024-09-24T17:05:55.230325Z ERROR zilliqa::api::zil: 40: e=invalid signature
```

Test program output:

```shell
Private Key: 0000000000000000000000000000000000000000000000000000000000000002

Wallet initialized and private key added.

Public Key: 02c6047f9441ed7d6d3045406e95c07cd85c778e4b8cef3ca7abac09b95c709ee5

Address: 8d393a22e4476ff8212de13fe1939de2a236f0a7

Provider initialized with host: http://172.17.0.1:4201

Balance and Nonce:
{
    "balance": "5000000000000000",
    "nonce": 0
}

Minimum Gas Price: 2000000016

Transaction Version (packed): 65537
[...]

PR #1592, opened 2024-10-08
(fix) Check for 0 nonce so as to avoid a panic!() later

rrw-zilliqa:
relates to #1502

### DIFF
--- a/zilliqa/src/api/zil.rs
+++ b/zilliqa/src/api/zil.rs
@@ -175,6 +175,12 @@ fn create_transaction(
 
     // TODO: Perform some initial validation of the transaction
 
+    // If we don't trap this here, it will later cause the -1 in
+    // transaction::get_nonce() to panic.
+    if transaction.nonce == 0 {
+        return Err(anyhow!("Invalid nonce (0)"));
+    }
+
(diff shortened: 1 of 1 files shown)


86667 (approved):

zilliqa/src/api/zil.rs
```diff
@@ -175,6 +175,12 @@ fn create_transaction(
 
     // TODO: Perform some initial validation of the transaction
 
+    // If we don't trap this here, it will later cause the -1 in
+    // transaction::get_nonce() to panic.
+    if transaction.nonce == 0 {
```
JamesHinshelwood, line 180: I wonder if we should make `nonce: NonZeroU64` in `TxZilliqa`?

JamesHinshelwood (approved):

zilliqa/src/api/zil.rs
rrw-zilliqa: I think that's probably a good idea ... I can issue a PR if you like?

JamesHinshelwood: Happy to do it separately, since this change is good in its own right :)

PR #1606, opened 2024-10-09
feat: Set block timestamp to be time of proposal rather than time of first Vote arriving

86667:
I feel this is a more useful `timestamp`

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -1153,7 +1153,7 @@ impl Consensus {
             proposal.header.transactions_root_hash,
             proposal.header.receipts_root_hash,
             proposal.transactions,
-            proposal.header.timestamp,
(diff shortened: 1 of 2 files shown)


zilliqa/src/consensus.rs
```diff
@@ -1153,7 +1153,7 @@ impl Consensus {
             proposal.header.transactions_root_hash,
             proposal.header.receipts_root_hash,
             proposal.transactions,
-            proposal.header.timestamp,
+            SystemTime::max(SystemTime::now(), parent_block.header.timestamp),
```
JamesHinshelwood, line 1156: This timestamp in the block should match the timestamp read by contracts in this transaction via `block.timestamp`. This is the timestamp we pass when executing transactions via `executed_block_header`.

86667: Thanks, I saw it passed in to tx processing but assumed tests would have caught if it broke.

Closing

PR #1620, opened 2024-10-11
Standardise proposal creation process 

### LINKED ISSUE #1320: wait for txns before proposing an empty block with an AggQC
In case of blocks proposed from `Vote`s (i.e. they have a QC attached), if there are no transactions in the mempool that we can include in the block, we wait before we propose an empty block: https://github.com/Zilliqa/zq2/blob/main/zilliqa/src/consensus.rs#L943-L961

In case of blocks proposed from `NewView` messages (i.e. they have an AggQC attached), this behaviour hasn't been implemented (yet).

86667:
Have leaders use the same proposal creation process whether via `vote()` (proposer for direct-chain with QC) or `new_view()` (proposer after previous block missed with AggQC). This closes #1320.


JamesHinshelwood (dismissed):
This is much clearer, thank you :)

86667:
@JamesHinshelwood Thanks. This is ready for review now. 

There is lots of `state` cloning going on which im not sure is necessary - will raise a separate PR with changes if so

JamesHinshelwood (approved):
LGTM, thank you :)

Cloning `State` is generally okay, since the underlying DB is wrapped in an `Arc` so really you're just cloning the few extra bits we store along side it - The one thing to be cautious of there is not to make a bunch of pending change to the state, without calculating the `root_hash()` (thus keeping those changes in memory) and then cloning those. This is another reason I want to have:
* One struct for state which we store and clone around
* Another struct for the transient state on which we execute transactions and eventually get a root hash from

PR #1629, opened 2024-10-14
Implement GetNumTransactions

### LINKED ISSUE #1541: GetNumTransactions
### TODO list

- [x] Write tests
- [x] Draft implementation
- [ ] Finalize and merge

maxconway:
- Closes #1541


JamesHinshelwood (approved):
This might have edge-cases around forks where a block containing a transaction was mined, but the block was later reverted. However, I think those will be transient (in the sense that the fork will eventually be resolved and we will delete the non-canonical blocks and transactions). Also, this API is really just a nice-to-have for block explorers to display a pretty number on their home page, so I think we can accept some variance.

PR #1630, opened 2024-10-14
Reduce state cloning 

86667:
We were cloning `self.state` lower down in the stack than necessary after `early_proposal_...` functions's behavior has changed slightly. These changes narrow cloning down to only:
- `early_proposal_apply_transcations()` 
- `early_proposal_finish_at()`

The previous `self.state` root is then restored before those fns return.  This results in far cleaner code which is much less error-prone.


86667:
Moved to draft to wait for merge queue so I can rebase..

JamesHinshelwood (approved):
Nice, thank you :)

PR #1634, opened 2024-10-15
Implement GetNumTxnsTxEpoch and GetNumTxnsDSEpoch

### LINKED ISSUE #1542: GetNumTxnsDSEpoch
### TODO list

- [x] Write tests
- [x] Draft implementation
- [ ] Finalize and merge

### LINKED ISSUE #1543: GetNumTxnsTxEpoch
### TODO list

- [x] Write tests
- [x] Draft implementation
- [ ] Finalize and merge

maxconway:
- Closes #1542
- Closes #1543


zilliqa/src/api/zil.rs
```diff
@@ -1040,3 +1042,27 @@ fn get_num_transactions(_params: Params, node: &Arc<Mutex<Node>>) -> Result<Stri
...
+
+fn get_num_txns_ds_epoch(_params: Params, node: &Arc<Mutex<Node>>) -> Result<String> {
+    let node = node.lock().unwrap();
+    let ds_epoch_size = TX_BLOCKS_PER_DS_BLOCK * node.config.consensus.blocks_per_epoch;
+    let current_epoch = node.get_chain_tip() / ds_epoch_size;
+    let current_epoch_first = current_epoch * ds_epoch_size;
+    let num_txns_epoch = node
+        .consensus
+        .block_store
+        .get_num_transactions_since_block_height(current_epoch_first)?;
```
JamesHinshelwood, line 1066: I think these APIs are just meant to return the number of transactions in the specified block?

> Returns the number of validated transactions included in this Transaction epoch. This is represented as a String.

Unless the docs are wrong and ZQ1 does something different?

maxconway: So, looking at the file you sent, James: https://github.com/Zilliqa/Zilliqa/blob/v9.3.4/src/libServer/LookupServer.cpp#L1647, it does indeed look like GetNumTxnsTxEpoch does indeed just return the transactions in the most recent TX block.

It looks to me like GetNumTxnsDsEpoch isn't as clear but I think it's just iterating backwards counting transactions.

So should we go with the following setup to match ZQ1:
- GetNumTxnsTxEpoch returns the transactions in the latest block (and maybe we add an additional explicit API for that?)
- Update the docs for GetNumTxnsTxEpoch to explain what it does.
- GetNumTxnsDsEpoch returns the total transactions over the simulated DS epoch (i.e. up to 100 block chunk)

Does that sound right to you, @JamesHinshelwood and @chetan-zilliqa

maxconway: I've updated this to match what I think ZQ1 does.

zilliqa/src/api/zil.rs
```diff
@@ -1039,3 +1041,34 @@ fn get_num_transactions(_params: Params, node: &Arc<Mutex<Node>>) -> Result<Stri
...
+    let current_epoch = node.get_chain_tip() / ds_epoch_size;
+    let current_epoch_first = current_epoch * ds_epoch_size;
+    let mut num_txns_epoch = 0;
+    for i in current_epoch_first..node.get_chain_tip() {
+        let block = node
+            .consensus
+            .block_store
+            .get_block_by_number(i)
+            .unwrap()
+            .unwrap();
```
JamesHinshelwood, line 1070: Might be nicer to return errors here? I guess we only hit it for catastrophic problems but for neatness? :)

JamesHinshelwood (approved):

zilliqa/src/api/zil.rs
JamesHinshelwood: Thanks, sorry for the slow response. That sounds good to me :)

zilliqa/src/api/zil.rs
maxconway: Done, good shout

JamesHinshelwood (approved):

PR #1638, opened 2024-10-16
Deposit contract extensions

DrZoltanFazekas:
Adds `getStakersData()` that returns an array of BLS pubkeys and an array of `Staker` tuples to make all data about stakers retrievable with a single call.

Also adds the `payable` function `increaseDeposit(bytes)` and the function `decreaseDeposit(bytes, uint256)` which topup and withdraw from a staker's deposit. Both are just temporary to make testing liquid staking easier until withdrawals are implemented.


zilliqa/src/contracts/deposit.sol
```diff
@@ -162,6 +162,27 @@ contract Deposit {
         _stakersMap[blsPubKey].peerId = peerId;
     }
 
+    // temporary function to test liquid staking
+    function increaseDeposit(bytes calldata blsPubKey) public payable {
```
JamesHinshelwood, line 166:
```suggestion
    function tempIncreaseDeposit(bytes calldata blsPubKey) public payable {
```

zilliqa/src/contracts/deposit.sol
```diff
@@ -162,6 +162,27 @@ contract Deposit {
...
+        Staker storage staker = _stakersMap[blsPubKey];
+        require(staker.keyIndex != 0, "unknown staker");
+        require(staker.rewardAddress == msg.sender, "invalid sender");
+        staker.balance += msg.value;
+        totalStake += msg.value;
+    }
+
+    // temporary function to test liquid staking
+    function decreaseDeposit(bytes calldata blsPubKey, uint256 amount) public {
```
JamesHinshelwood, line 175:
```suggestion
    function tempDecreaseDeposit(bytes calldata blsPubKey, uint256 amount) public {
```

JamesHinshelwood (approved):
LGTM, feel free to ignore my suggestions :)

JamesHinshelwood (dismissed):

PR #1639, opened 2024-10-16
Add further verification to CreateTransaction

### LINKED ISSUE #1508: CreateTransaction (and probably other ZQ-native APIs) don't perform the same validation checks  as ZQ1
In particular, CreateTransaction in Zilliqa 1 checks that the toAddress is correctly checksummed either by Zilliqa or EVM rules. ZQ2 doesn't.

There are probably other instances of this through the API; this has the effect that anyone who is relying on eg. bad addresses being rejected by the API will be disappointed and transactions may be submitted successfully that they weren't expecting, leading to tokens getting lost.

rrw-zilliqa:
In fact, since we can't predict the state of the chain when the transaction will be applied, there's only one additional test here - the check that the incoming address (with 0x or without) must have a valid EIP-55 or Zilliqa checksum.

Closes #1508


maxconway (approved):
Looks sane and works for me.

PR #1642, opened 2024-10-17
Remove hex encoding in checkpoint files

86667:
- Write checkpoint file as single byte array
- Add Scilla state to checkpoint test


zilliqa/src/db.rs
```diff
@@ -819,42 +848,48 @@ pub fn checkpoint_block_with_state<P: AsRef<Path> + Debug>(
     writer.write_all(b"\n")?;
 
     // write the block...
-    writer.write_all(hex::encode(bincode::serialize(&block)?).as_bytes())?; // TODO: better serialization (#1007)
-    writer.write_all(b"\n")?;
+    let block_ser = &bincode::serialize(&block)?;
+    // using u32 to store the block size. This gives us length values representing data up to ~500 MB
```
86667, line 852: This is fairly arbitrarily chosen. 4 bytes to store a length give us lengths up to around 500 MB. This is surely way larger than we expect blocks to become.

The same is used for all other data too, we use `u32` to store the lengths of:
- blocks
- transactions
- serialised accounts (Code could be large)
- storage values

For storage keys i used `u16`.

The cost of using  larger-than-necessary representation of length is minimal for most data, apart from storage as it may be may up of many values. Do we have a way to find the limit of an account's storage map?

JamesHinshelwood: The limits of an account's storage will depend on Scilla implementation details and I'm not sure there's any direct limit.

Can we use `u32::try_from` to ensure we will explode if data is too large, rather than truncating the larger bits?

JamesHinshelwood (approved):
Thanks, LGTM :)

zilliqa/src/db.rs
```diff
@@ -883,13 +883,13 @@ pub fn checkpoint_block_with_state<P: AsRef<Path> + Debug>(
             .at_root(bincode::deserialize::<Account>(&serialised_account)?.storage_root);
         let mut account_storage_buf = vec![];
         for (storage_key, storage_val) in account_storage.iter() {
-            account_storage_buf.extend_from_slice(&(storage_key.len() as u16).to_be_bytes());
+            account_storage_buf.extend_from_slice(&u16::try_from(storage_key.len())?.to_be_bytes());
```
JamesHinshelwood, line 886: Aren't storage keys fixed length, just like account keys and thus the length here is unnecessary?

JamesHinshelwood (dismissed):

zilliqa/src/db.rs
rrw-zilliqa: I'd use `u64` for everything and then compress the result, or one day an object will get too large and then you will pay the price of having to test two different versions. Trying to design compact binary representations is the modern version of coding in assembler - these days, the compiler can do it better than you can.

that said, I second using `try_from()` - bugs in this code will be experienced by those running staking validators, who probably don't have time to debug for us, and the files they will be using are probably inconveniently large so will be hard for us to get for debugging (not to mention that when we've fixed the problem, generating the new snapshots for the new checkpoint version will be time-consuming and expensive to distribute) - I think it makes sense to be conservative.

zilliqa/src/db.rs
86667: Its storage keys which do have the variable length for Scilla, as you suspected the account keys might in the original bug ticket

zilliqa/src/db.rs
86667: Good point - I'll use `u64` for all to be safe and compress away the leading 0s. Cheers!

Using `u64` does mean that a value exceeding `(2^32)-1`  will error when we `try_into::<usize>()`  on a 32-bit system. Im not sure if 32-bit systems are worth worrying about in our expected execution environments, what do you guys think?

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):

PR #1643, opened 2024-10-17
Don't execute txns if they are suppose to fail due to lack of funds

bzawisto:
This PR fixes the problem when transactions were dropped if they failed due to lack of funds. In the new implementation used funds are tracked per user and if max_transaction_cost exceeds possessed funds at timeX then such transaction is not even attempted to be executed (which is what geth/reth does). In the preview mempool call, transaction which might fail to execute is returned in queued list (as reth does). In addition, ready transactions are stored now in a tree instead of heap which in some cases helped with excessive cloning.

### DIFF
--- a/zilliqa/src/pool.rs
+++ b/zilliqa/src/pool.rs
@@ -515,7 +515,7 @@ mod tests {
             request_id: RequestId::default(),
         };
 
-        let db = Db::new::<PathBuf>(None, 0)?;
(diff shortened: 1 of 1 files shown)


JamesHinshelwood (approved):
LGTM - Sorry for the slow review.

JamesHinshelwood (approved):

PR #1644, opened 2024-10-17
(feat) Run evm_scilla_js_tests with z2

rrw-zilliqa:
Some boilerplate to allow evm_scilla_js_tests to run in z2 networks; my next move is to try to get them to pass (including the Scilla library tests) - after that if you want it should be easy enough to add them to CICD.

Do say if you want an issue filed for reference here ..


saeed-zil (approved):
LGTM 💯

PR #1663, opened 2024-10-21
Implement GetTotalCoinSupply and GetTotalCoinSupplyAsInt

### LINKED ISSUE #1549: GetTotalCoinSupply
### TODO list

- [x] Write tests
- [ ] Draft implementation
- [ ] Finalize and merge

### LINKED ISSUE #1550: GetTotalCoinSupplyAsInt
### TODO list

- [x] Write tests
- [ ] Draft implementation
- [ ] Finalize and merge

maxconway:
- Plus tests
- serde_json updated to latest version (v1.0.132) in order to gain access to support u128
- Closes #1549
- Closes #1550

### DIFF
--- a/zilliqa/Cargo.toml
+++ b/zilliqa/Cargo.toml
@@ -61,7 +61,7 @@ revm-inspectors = { version = "0.8.1", features = ["js-tracer"] }
 rusqlite = { version = "0.32.1", features = ["bundled", "trace"] }
 serde = { version = "1.0.210", features = ["derive", "rc"] }
 serde_bytes = "0.11.14"
-serde_json = { version = "1.0.128", features = ["raw_value"] }
+serde_json = { version = "1.0.132", features = ["raw_value","arbitrary_precision"] }
 sha2 = "0.10.8"
 sha3 = "0.10.8"
 sled = "0.34.7"
--- a/zilliqa/tests/it/zil.rs
+++ b/zilliqa/tests/it/zil.rs
@@ -1647,62 +1647,7 @@ async fn get_num_txns_tx_epoch_1(mut network: Network) {
 }
 
 #[zilliqa_macros::test]
-async fn get_total_coin_supply(mut network: Network) {
-    let wallet = network.genesis_wallet().await;
-
-    let response: Value = wallet
-        .provider()
-        .request("GetTotalCoinSupply", [""])
-        .await
-        .expect("Failed to call GetTotalCoinSupply API");
-
-    assert!(
-        response.is_string(),
-        "Expected response to be a string, got: {:?}",
-        response
-    );
-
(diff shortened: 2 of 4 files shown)


zilliqa/Cargo.toml
```diff
@@ -61,7 +61,7 @@ revm-inspectors = { version = "0.8.1", features = ["js-tracer"] }
 rusqlite = { version = "0.32.1", features = ["bundled", "trace"] }
 serde = { version = "1.0.210", features = ["derive", "rc"] }
 serde_bytes = "0.11.14"
-serde_json = { version = "1.0.128", features = ["raw_value"] }
+serde_json = { version = "1.0.132", features = ["raw_value","arbitrary_precision"] }
```
JamesHinshelwood, line 64: Is this feature actually required to get `serde_json` to serialize `u128`s? I had one experience in the past (3+ years ago) where this feature interacted badly with some other library that used serde in a weird and hard-to-spot way that makes me wary :)

zilliqa/tests/it/zil.rs
```diff
@@ -1645,3 +1645,92 @@ async fn get_num_txns_tx_epoch_1(mut network: Network) {
...
+        .as_u128()
+        .expect("Expected u128 conversion");
+    assert!(
+        total_coin_supply_as_int > 0,
+        "Total coin supply should be greater than 0"
+    );
+}
+
+#[zilliqa_macros::test]
+async fn compare_total_coin_supply(mut network: Network) {
```
JamesHinshelwood, line 1705: Maybe nitpicking - But can we just put all 3 assertions in this single test?

JamesHinshelwood (dismissed):

zilliqa/Cargo.toml
maxconway: Yep, without it it appears to convert to a float (Number(2.1e28)).
It's a bit tricky to tell what's going on because without `arbitrary_precision` we also can't deserialise values, since `as_u128() ` just returns None, so it's not really possible to test it.

zilliqa/tests/it/zil.rs
maxconway: Done

zilliqa/Cargo.toml
maxconway: Right, it looks like the specific problem is that without `arbitrary_precision`, the values above u64::MAX get serialised to a float, and `as_u128()` quite reasonably don't deserialise a float

JamesHinshelwood (approved):

zilliqa/Cargo.toml
JamesHinshelwood: Huh fair enough, thanks for double-checking :)

Issue #1341, opened 2024-08-23
Additional integration tests for staking


theo-zil:
There exist integration tests for

- Registering a new validator (with stake)
- Unstaking a validator (i.e. removing its entire stake)

Additional tests would be useful, for:

- [ ] Adding stake to an existing validator
    - [ ] through `deposit()`,
    - [ ] through `depositTopup()
 - [ ] Unstaking partial stake from a validator while keeping it `Active`
 - [ ] Withdrawing unstaked funds
 - [ ] A validator suspending itself
 - [ ] Slashing a validator

Issue #1339, opened 2024-08-23
Review block-related constants passed to the EVM; potentially fix epoch ticking


theo-zil:
Currently, when we invoke the EVM to process transactions, we pass the parent block's header as the block context. This means that all block-related constants - most visibly `block.number`; but technically also ones such as `block.prevrandao` (currently little used, but will become relevant at epoch boundaries), `block.gaslimit` (we currently have a static one, but what if we didn't?), `block.coinbase` (currently Address::ZERO, but what about when we set it to reward addresses - #1340 ?) etc. all reference the parent.

This has a particularly curious implication for epoch changes: because `deposit.sol` checks the block number for incrementing the epoch, this means that technically, the epoch state changes (i.e. all validator set changes) happen as the first state change of the _second_ block of the epoch, or the first block _after_ `block.number() & blocks_per_epoch == 0` - since the parent's block number is what's passed.

If this is intended behaviour, then `deposit.sol` should probably be updated to check the pending (i.e. currently-executing) block number by checking for `block.number + 1` in solidity.
And if this is not intended behaviour, then the block context passed to transactions should be fixed - which is slightly tricky as currently we pass a block header, but no complete block header exists when we are proposing a new block; however, most values should still be present, so it should be possible to fill out the context.

Issue #1338, opened 2024-08-23
Staking: fix total_stake change on slashing


theo-zil:
The deposit contract operates on the principle that all changes are queued, and get enacted after at least one epoch (and at most two) have passed. This greatly simplifies validator set tracking for cross-chain applications. Thus, there is an invariant: the total stake only changes at epoch boundaries, and (almost) every change is visible at least one epoch in advance.

Slashing is handled by simply marking slashed validators, and excluding them from leader selection. Otherwise, their stake is not removed from the `totalStake` until an epoch boundary.

However, currently, slashed validators are removed from their committee at the _next_ epoch boundary. This means that total stake can change with very little notice, if a slashing happens. (Additionally, suspensions/jailings are handled in the same way.)
Instead, this should follow the same logic as all other state changes and have a minimum time of one epoch before being enacted.

In practice, this means changing `kickQueue` in deposit.sol to function analogously to the pending and unstaking queues.

Issue #1337, opened 2024-08-23
Staking: implement suspension features


theo-zil:
Currently, validators can suspend themselves (to temporarily cease operating, or to notify the network that the validator node is down for example). Two extra features need to be added to the `deposit.sol` contract:

- [ ] An unsuspend method. It should be callable from the validator's `controlAddress`; check that the validator is indeed suspended; change its status to `Pending` and insert an entry into the current `pendingQueue` (with a value of `0`, as no extra stake is being deposited, of course).
- [ ] A consensus-based suspend method: to be invoked once a consensus mechanism is added for dealing with validators that violate liveness. Currently, suspensions are only voluntary; something triggerable by consensus, similar to slashing, should be added.

PR #1679, opened 2024-10-22
Apply consensus committee changes at epoch boundaries

JamesHinshelwood:
The most important changes in this commit are in `deposit.sol`. Everything else is pretty much just supporting the new interfaces in the deposit contract. The key feature is that we now keep track of the current 'epoch' and only apply changes to the consensus committee two epochs after they are made.

There are a few other changes to the deposit contract in this commit too:
* The initial stakers are now passed to the contract constructor, rather than relying on the `setStake` hack.
* `leader()` was removed since it isn't used anywhere and it doesn't make sense to have both `leader` and `leaderAtView`.
* Delayed withdrawals are now supported - We keep a circular buffer of pending withdrawals for each validator. 2 weeks after unstaking, the validator can make another call to withdraw the funds from the deposit contract.

Reworked version of #1298.

Further work out of scope of this PR:

* Fill out missing integration tests - [Additional integration tests for staking #1341](https://github.com/Zilliqa/zq2/issues/1341)
* Implement unsuspend, and consensus-based suspend - [Staking: implement suspension features #1337](https://github.com/Zilliqa/zq2/issues/1337)
* Epoch length tracking in EVM - part of [Move consensus config on-chain (into shard contract) #1135](https://github.com/Zilliqa/zq2/issues/1135). (Currently, the epoch length is hardcoded in `deposit.sol`.)
* Improving staker address and pubkey validation on registration - [Improve validator address verification in deposit contract #1297](https://github.com/Zilliqa/zq2/issues/1297)
* Allowing validators to change their registered addresses - [Control and reward address changes for validators #1301](https://github.com/Zilliqa/zq2/issues/1301)
* Adjust slashing and suspension queues - [Staking: fix total_stake change on slashing #1338](https://github.com/Zilliqa/zq2/issues/1338)
* Consider implementing churn limits on the validator set - [Queue/churn limits in deposit.sol #1300](https://github.com/Zilliqa/zq2/issues/1300)
* Improving epoch change transaction tracking - [Gas (and account) tracking for epoch changes #1299](https://github.com/Zilliqa/zq2/issues/1299)
* Slashing mechanisms - currently, this reuses the previous mechanism where any validator with >=10% of total stake can slash any other validator. Additionally, slashed funds are not kept track of (effectively, burned).
* Allow reusing unstaked-but-not-yet-withdrawn funds - [Deposit: allow reusing unstaked balance #1329](https://github.com/Zilliqa/zq2/issues/1329)
* Not directly necessary, but somewhat related: block number tracking - [Review block-related constants passed to the EVM; potentially fix epoch ticking #1339](https://github.com/Zilliqa/zq2/issues/1339)


zilliqa/src/contracts/deposit.sol
```diff
@@ -138,123 +366,173 @@ contract Deposit {
...
+            bytes storage lastStakerKey = futureCommittee.stakers[
+                futureCommittee.stakers.length - 1
+            ];
+            uint256 lastStakerBalance = futureCommittee.balances[
+                futureCommittee.balances.length - 1
+            ];
+            Staker storage lastStaker = _stakersMap[lastStakerKey];
+            futureCommittee.stakers[uint(staker.keyIndex)] = lastStakerKey;
+            futureCommittee.balances[uint(staker.keyIndex)] = lastStakerBalance;
+            lastStaker.keyIndex = staker.keyIndex;
```
86667, line 456: As you mention in the FIXME below were setting the `keyIndex` on `lastStaker` too early. Whenever we do so we also need to set `staker.keyIndex` to something else, perhaps -1. Not sure if its omitted here on purpose or forgotten

zilliqa/src/contracts/deposit.sol
```diff
@@ -138,123 +366,173 @@ contract Deposit {
...
         require(pop, "rogue key check");
 
-        uint256 keyIndex = _stakersMap[blsPubKey].keyIndex;
-        if (keyIndex == 0) {
-            // The staker will be at index `_stakerKeys.length`. We also need to add 1 to avoid the 0 sentinel value.
-            _stakersMap[blsPubKey].keyIndex = _stakerKeys.length + 1;
-            _stakerKeys.push(blsPubKey);
+        Staker storage staker = _stakersMap[blsPubKey];
+        // This must be a new staker, meaning the control address must be zero.
+        require(staker.controlAddress == address(0), "staker already exists");
```
86667, line 375: The name of this fn is misleading. Since a staker can call this successfully only once is something like `initialiseStaker()` more clear

86667 (commented):
Nice! Couple minor comments/questions

z2/resources/config.tera.toml
```diff
@@ -9,7 +9,7 @@ eth_chain_id = {{ eth_chain_id }}
 allowed_timestamp_skew = { secs = 60, nanos = 0 }
 data_dir = "/data"
 consensus.genesis_accounts = [ ["{{ genesis_address }}", "21_000_000_000_000_000_000_000_000_000"] ]
-consensus.genesis_deposits = [ ["{{ bootstrap_bls_public_key }}", "{{ bootstrap_peer_id }}", "100_000_000_000_000_000_000_000_000", "{{ genesis_address }}"] ]
+consensus.genesis_deposits = [ ["{{ bootstrap_bls_public_key }}", "{{ bootstrap_peer_id }}", "100_000_000_000_000_000_000_000_000", "{{ genesis_address }}", "{{ genesis_address }}"] ]
```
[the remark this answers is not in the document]
JamesHinshelwood: I figured it only makes sense to update these once we deploy the change to the actual networks - And at that point we'll be committing something anyway in `zq2-devnet.yaml`.

@frankmeds any opinions?

zilliqa/src/contracts/deposit.sol
```diff
@@ -1,40 +1,248 @@
...
+        uint256 position = randomness % committee().totalStake;
         uint256 cummulative_stake = 0;
 
-        for (uint256 i = 0; i < _stakerKeys.length; i++) {
-            bytes storage stakerKey = _stakerKeys[i];
-            Staker storage staker = _stakersMap[stakerKey];
+        // TODO: Consider binary search for performance. Or consider an alias method for O(1) performance.
+        for (uint256 i = 0; i < committee().stakers.length; i++) {
+            bytes memory stakerKey = committee().stakers[i];
+            uint256 stakedBalance = committee().balances[i];
```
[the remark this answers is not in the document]
JamesHinshelwood: Thanks for bringing that up - I assumed the Solidity compiled was smart enough to see that `committee()` was pure, but from debugging I found that it isn't :disappointed: 

I've removed the repeated calls.

zilliqa/src/contracts/deposit.sol
JamesHinshelwood: @DrZoltanFazekas Any opinions on naming preference?

z2/resources/config.tera.toml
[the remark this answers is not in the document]
86667: That makes sense.

zilliqa/src/contracts/deposit.sol
```diff
@@ -1,40 +1,256 @@
...
+                "control address cannot be zero"
+            );
+
+            Committee storage currentCommittee = committee();
+            require(
+                currentCommittee.stakerKeys.length < maximumStakers,
+                "too many stakers"
+            );
+
+            Staker storage staker = _stakersMap[blsPubKey];
```
86667, line 198: is this `_stakersMap` variable ever being set?

86667 (approved):

PR #1680, opened 2024-10-22
More flexible block fetching, and a few bug fixes.

### LINKED ISSUE #1364: Checkpoints break the syncing algorithm in `BlockStore`

rrw-zilliqa:
.. see the (rather long) commit list for a list of things this fixes.

We also add the ability for z2 to join nodes to a configured network, for debugging.

Closes #1364

### DIFF
--- a/z2/resources/chain-specs/zq2-devnet.toml
+++ b/z2/resources/chain-specs/zq2-devnet.toml
@@ -5,7 +5,7 @@ bootstrap_address = [ "12D3KooWJNUNgs1UM9mw1Pfskxk4o3VWz37cpneSHZsUecGGZgkb", "/
 eth_chain_id = 33469
 allowed_timestamp_skew = { secs = 60, nanos = 0 }
 data_dir = "/data"
-consensus.genesis_accounts = [ ["0x977a797a0546cdb1082c2368921af455dfdc7a5a", "21_000_000_000_000_000_000_000_000_000"] ]
+consensus.genesis_accounts = [ ["0x977a797a0546cdb1082c2368921af455dfdc7a5a", "210_000_000_000_000_000_000_000_000"] ]
 consensus.genesis_deposits = [ ["8a968db895dc1f17d1e13bbf1582078f82657247820411ad3fe4f5816a75566cd9a4ba9f8e73f6306d7a7b7086e4e8a6", "12D3KooWJNUNgs1UM9mw1Pfskxk4o3VWz37cpneSHZsUecGGZgkb", "100_000_000_000_000_000_000_000_000", "0x977a797a0546cdb1082c2368921af455dfdc7a5a"] ]
 
 # Reward parameters
--- a/z2/src/bin/z2.rs
+++ b/z2/src/bin/z2.rs
@@ -1,10 +1,11 @@
 use std::{
     collections::{HashMap, HashSet},
     env, fmt,
+    str::FromStr,
 };
 
 use alloy::primitives::B256;
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use clap::{builder::ArgAction, Args, Parser, Subcommand};
 use clap_verbosity_flag::{InfoLevel, Verbosity};
 use z2lib::{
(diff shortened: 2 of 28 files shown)


z2/resources/chain-specs/zq2-devnet.toml
```diff
@@ -5,7 +5,7 @@ bootstrap_address = [ "12D3KooWJNUNgs1UM9mw1Pfskxk4o3VWz37cpneSHZsUecGGZgkb", "/
 eth_chain_id = 33469
 allowed_timestamp_skew = { secs = 60, nanos = 0 }
 data_dir = "/data"
-consensus.genesis_accounts = [ ["0x977a797a0546cdb1082c2368921af455dfdc7a5a", "21_000_000_000_000_000_000_000_000_000"] ]
+consensus.genesis_accounts = [ ["0x977a797a0546cdb1082c2368921af455dfdc7a5a", "210_000_000_000_000_000_000_000_000"] ]
```
frankmeds, line 8: changing the value in the config.tera.toml may be worth changing the value also in all the chain specs except for prototestnet. The reason is because the devnets are deployed quite often. Does this make sense?

z2/src/bin/z2.rs
```diff
@@ -386,6 +389,9 @@ struct JoinStruct {
     /// Specify the ZQ2 chain you want join
     #[clap(long = "chain")]
     chain_name: chain::Chain,
+    /// Specify the container label to run
+    #[clap(long = "container")]
+    container: Option<String>,
```
frankmeds, line 394: not mandatory change: can be more intuitive to replace "container" with "image" since recall immediately the value of the string?

frankmeds (approved):

z2/src/deployer.rs
```diff
@@ -522,7 +522,9 @@ pub async fn run_restore(config_file: &str, filename: &str, max_parallel: usize)
 }
 
 pub async fn run_reset(config_file: &str, node_selection: bool) -> Result<()> {
-    let config = NetworkConfig::from_file(config_file).await?;
+    let config = NetworkConfig::from_file(config_file)
+        .await
+        .context(format!("Cannot read {config_file}"))?;
```
frankmeds, line 527: having the function `from_file` used in multiple places, may be worth adding `context` within it rather than having different behaviours?

rrw-zilliqa: Will do this in a separate PR else this one will never merge..

z2/src/bin/z2.rs
rrw-zilliqa: Separate PR else I'll never merge, but good point - thanks .

z2/resources/chain-specs/zq2-devnet.toml
rrw-zilliqa: OK - again, separate PR else will never merge.

rrw-zilliqa: Fixed in https://github.com/Zilliqa/zq2/pull/1683 (I hope!)

z2/src/bin/z2.rs
rrw-zilliqa: Want to drop me a line on slack? This is called `container` because the parameter it feeds is called `container` and that feeds a Tera variable called `container`, so probably a good plan to work out how far down that stack we'd like the rename to go ..

PR #1681, opened 2024-10-23
Feature: fast-foward self-proposals

shawn-zil:
In the case where the Leader sends out a Proposal, that leader is allowed to *fast-forward* the validation of that block since it had already done all the work to produce the Proposal. This effectively reduces some of the double work done by Leaders i.e. executing transactions once when producing the Proposal and subsequently when validating it.

In effect, what happens is:
1. State trie is stored in the DB during block assembly, and the state-root-hash pointer is advanced during validation.
2. Receipts/Addresses are cached in memory during block assembly, and committed to DB during validation.
3. Actual execution of each transaction is skipped during validation.

The result is an orders of magnitude improvement in validation time for the node that was the previous Leader. Other nodes continue doing all the necessary work to validate and mutate state as normal.

In the case where the previous Leader is also the next Leader, this allows the Vote to be quickly cast and the early assembly for the next block to be started.

Also includes some minor fixes/features:
- Routing incoming Proposals to the requests queue to skip the broadcasts queue.
- Fixed an issue where it takes longer for the Leader to recover verified transactions from the Proposal breaking a previous enhancement.
- Fixed a subtle issue where new transactions get inserted under stale state, which may result in blocked transactions.
- miscellaneous cleanup.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -464,9 +464,12 @@ impl Consensus {
                     )));
                 };
             } else {
-                self.reset_timeout.send(Duration::from_millis(
-                    empty_block_timeout_ms - time_since_last_view_change + 1,
-                ))?;
+                self.reset_timeout.send(
+                    self.config
+                        .consensus
+                        .empty_block_timeout
+                        .saturating_sub(Duration::from_millis(time_since_last_view_change)),
+                )?;
                 return Ok(None);
             }
         }
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -2737,10 +2737,11 @@ impl Consensus {
             info!("Fast-forward self-proposal");
             // recover previous receipts from cache
             for (tx_index, txn_hash) in block.transactions.iter().enumerate() {
+                // Retrieve set of receipts
                 block_receipts.push((
                     self.receipts_cache
                         .remove(txn_hash)
-                        .expect("receipt inserted during proposal assembly"),
+                        .expect("receipt cached during proposal assembly"),
                     tx_index,
                 ));
                 // TODO: Apply 'touched-address' from cache
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -1175,7 +1175,6 @@ impl Consensus {
         {
             return Ok(());
         }
-        info!("assemble early proposal for view {}", self.view.get_view());
 
         let (qc, parent) = match agg {
             // Create dummy QC for now if aggQC not provided
@@ -1202,7 +1201,20 @@ impl Consensus {
             }
         };
 
-        info!("parent block number: {}", parent.header.number);
+        // This is a partial header of a block that will be proposed with some transactions executed below.
+        // It is needed so that each transaction is executed within proper block context (the block it belongs to)
+        let executed_block_header = BlockHeader {
+            view: self.view(),
+            number: parent.header.number + 1,
+            timestamp: parent.header.timestamp, // will be overridden by `finish_early_proposal_at`
+            gas_limit: self.config.consensus.eth_block_gas_limit,
+            ..BlockHeader::default()
+        };
+
+        info!(
+            "assemble early proposal {} in view {}",
+            executed_block_header.number, executed_block_header.view
+        );
 
         // Ensure sane state
         if self.state.root_hash()? != parent.state_root_hash() {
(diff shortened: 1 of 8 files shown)


zilliqa/src/consensus.rs
```diff
@@ -434,9 +442,12 @@ impl Consensus {
                     )));
                 };
             } else {
-                self.reset_timeout.send(Duration::from_millis(
-                    empty_block_timeout_ms - time_since_last_view_change + 1,
-                ))?;
+                self.reset_timeout.send(
+                    self.config
+                        .consensus
+                        .empty_block_timeout
```
86667, line 448: is this missing a `.as_millis()` as on line 425

zilliqa/src/consensus.rs
```diff
@@ -1119,7 +1131,7 @@ impl Consensus {
             proposal.header.transactions_root_hash,
             proposal.header.receipts_root_hash,
             proposal.transactions,
-            proposal.header.timestamp,
+            SystemTime::max(SystemTime::now(), parent_block.timestamp()),
```
86667, line 1134: I tried doing this also, it seems transactions use the block timestamp when executing so we cannot change it here after having executed transactions for this block. See: https://github.com/Zilliqa/zq2/pull/1606

zilliqa/src/consensus.rs
shawn-zil: It sends and receives a `Duration` so I don't think that's necessary.

zilliqa/src/consensus.rs
shawn-zil: Noted. Will revert.

zilliqa/src/block_store.rs
```diff
@@ -132,8 +132,10 @@ impl BlockCache {
     /// (note that this will be BLOCK_CACHE_HEAD_BUFFER_ENTRIES >> shift cached views, since the
     /// head cache is set associative)
     pub fn min_head_cache_key(&self, highest_known_view: u64) -> u128 {
+        let delta = u128::try_from(constants::BLOCK_CACHE_HEAD_BUFFER_ENTRIES).unwrap();
         let highest_key = u128::from(highest_known_view + 1) << self.shift;
-        highest_key - u128::try_from(constants::BLOCK_CACHE_HEAD_BUFFER_ENTRIES).unwrap()
```
shawn-zil, line 136: This fixes a bunch of failing tests in `cargo test` for me.

rrw-zilliqa: "oops!" - sorry! Well caught ... I wonder why that didn't panic in dev mode for me?

rrw-zilliqa (approved):
Looks good to me, FWIW! Sorry!

zilliqa/src/consensus.rs
```diff
@@ -2718,150 +2738,178 @@ impl Consensus {
...
-            };
-            verified_txns.push(txn);
-        }
+        let mut block_receipts: Vec<(TransactionReceipt, usize)> = Vec::new();
 
-        let mut block_receipts = Vec::new();
-        let mut cumulative_gas_used = EvmGas(0);
-        let mut receipts_trie = eth_trie::EthTrie::new(Arc::new(MemoryDB::new(true)));
-        let mut transactions_trie = eth_trie::EthTrie::new(Arc::new(MemoryDB::new(true)));
+        if from.is_some_and(|peer_id| peer_id == self.peer_id()) {
```
86667, line 2743: Is it worth checking here that we have the correct number of transaction receipts in memory here in case the node has restarted in-between broadcasting the proposal and calling this code? 

Similarly, we may begin building this data without using it, say if we think were leader but there is a fork and the other branch wins. In this case should we nullify our in-memory receipts and add the transactions back to the mempool?

86667 (approved):
Nice :+1: 
If you agree with my comments feel free to address in follow up PR or issue future improvement ticket

zilliqa/src/consensus.rs
shawn-zil: Good idea. I'll implement a flush to the receipts cache.

zilliqa/src/block_store.rs
rrw-zilliqa: https://github.com/Zilliqa/zq2/pull/1703

zilliqa/src/consensus.rs
```diff
@@ -1754,11 +1793,21 @@ impl Consensus {
             return Ok(TxAddResult::Duplicate(txn.hash));
         }
 
-        let account = self.state.get_account(txn.signer)?;
+        // Perform insertion under early state, if available
+        let early_account = match self.early_proposal.as_ref() {
```
shawn-zil, line 1797: the code had a subtle error here, where new txns were being inserted under stale state/account info. this fix uses the state in the early proposal, if present.

zilliqa/src/consensus.rs
```diff
@@ -1754,11 +1793,21 @@ impl Consensus {
...
+            Some((block, _, _, _)) => {
+                // let mut state = self.state.clone();
```
86667, line 1799: comments should be removed?

86667 (approved):

86667 (approved):

shawn-zil:
Also impacts #1322

PR #1684, opened 2024-10-23
(feat) Trigger checkpoints via an admin_ RPC call

### LINKED ISSUE #1635: Consider controlling checkpoint generation externally
Instead of configuring the node to produce checkpoints every X blocks, one could imagine a simpler design where the node just generates a checkpoint when it is told to, via an API.

rrw-zilliqa:
Closes #1635

### DIFF
--- a/zilliqa/src/api/admin.rs
+++ b/zilliqa/src/api/admin.rs
@@ -1,0 +1,40 @@
+//! An administrative API
+
+use std::sync::{Arc, Mutex};
+
+use alloy::eips::BlockNumberOrTag;
+use anyhow::{anyhow, Result};
+use jsonrpsee::{types::Params, RpcModule};
+use serde::{Deserialize, Serialize};
+
+use crate::{api::to_hex::ToHex, node::Node};
+
+pub fn rpc_module(node: Arc<Mutex<Node>>) -> RpcModule<Arc<Mutex<Node>>> {
+    super::declare_module!(node, [("admin_generateCheckpoint", checkpoint)])
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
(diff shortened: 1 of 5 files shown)


zilliqa/src/api/admin.rs
```diff
@@ -0,0 +1,40 @@
...
+
+fn checkpoint(params: Params, node: &Arc<Mutex<Node>>) -> Result<CheckpointResponse> {
+    let mut params = params.sequence();
+    let block_number: BlockNumberOrTag = params.next()?;
+    let mut node = node.lock().unwrap();
+    let block = node
+        .get_block(block_number)?
+        .ok_or(anyhow!("Block {block_number} does not exist"))?;
+
+    let (file_name, hash) = node.consensus.checkpoint_at(block.number())?;
```
JamesHinshelwood, line 34: How long does this take? Presumably it can be a long time for large checkpoints - Should this API be asynchronous?

JamesHinshelwood (approved):

zilliqa/src/api/admin.rs
rrw-zilliqa: Should be, and is - all this call actually does is to retrieve the relevant block (which should be quite fast) and send an "ExportBlockCheckpoint" message to the node, which causes the checkpoint to actually happen at some later point. The usual write-to-temp-and-rename mechanism then ensures that once the file exists, it must be complete.

JamesHinshelwood: Ah cool - It basically already is async then! Thanks and sorry for not checking myself :)

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #1692, opened 2024-10-24
Add contract_address in response if code is present

### LINKED ISSUE #1691: Add contract address in response for contract creation call of scilla

chetan-zilliqa:
This fixes #1691 and https://github.com/Zilliqa/gozilliqa-sdk/issues/122

### DIFF
--- a/zilliqa/src/api/zil.rs
+++ b/zilliqa/src/api/zil.rs
@@ -19,6 +19,15 @@ use jsonrpsee::{
 use k256::elliptic_curve::sec1::ToEncodedPoint;
 use serde::{Deserialize, Deserializer};
 use serde_json::{json, Value};
+use sha2::Sha256;
+use sha3::{
+    digest::generic_array::{
+        sequence::Split,
+        typenum::{U12, U20},
+        GenericArray,
+    },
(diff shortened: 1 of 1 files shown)


zilliqa/src/api/zil.rs
```diff
@@ -162,6 +171,12 @@ where
     let s = String::deserialize(deserializer)?;
     s.parse().map_err(serde::de::Error::custom)
 }
+/// Helper function to extract signer address from a public key
+fn extract_signer_address(key: &schnorr::PublicKey) -> Address {
```
JamesHinshelwood, line 175: We already do this work when we call `txn.verify()` in `Consensus::handle_new_transaction`. Can we instead pass the signer up from there?

chetan-zilliqa: Yes, I considered that, but this code handles both Scilla and EVM transactions. I have two questions:

How can we differentiate between an EVM payment transaction and a smart contract transaction?
How can we determine the contract address for an EVM transaction?

JamesHinshelwood (dismissed):

frankmeds (approved):

PR #1708, opened 2024-10-25
Adds placeholders for in-progress APIs, plus comments

maxconway:
This is a bit of a housekeeping PR - it shouldn't change functionality substantially.

I've added placeholder API functions and tests for 9 API endpoints. This is intended to make merging easier by avoiding the situation where I have a PR for one API endpoint in flight, but get merge conflicts if I try to start work on another because they're both adding to the end of the file. This approach of adding placeholders avoids the alternative of creating loads of separate files.

In addition, I've done the same thing for tests, added the return type for getminerinfo, and added comments to API endpoint functions indicating what API endpoint they service.

@JamesHinshelwood Hopefully this avoids your concerns about making things more complicated with lots of files, while still making merges easier.

### DIFF
--- a/zilliqa/src/api/zil.rs
+++ b/zilliqa/src/api/zil.rs
@@ -24,9 +24,9 @@ use super::{
     to_hex::ToHex,
     types::zil::{
         self, BlockchainInfo, DSBlock, DSBlockHeaderVerbose, DSBlockListing, DSBlockListingResult,
-        DSBlockRateResult, DSBlockVerbose, GetCurrentDSCommResult, RecentTransactionsResponse,
-        SWInfo, ShardingStructure, SmartContract, TXBlockRateResult, TransactionBody,
-        TxBlockListing, TxBlockListingResult, TxnBodiesForTxBlockExResponse,
+        DSBlockRateResult, DSBlockVerbose, GetCurrentDSCommResult, MinerInfo,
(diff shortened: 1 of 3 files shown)


zilliqa/src/api/zil.rs
```diff
@@ -1074,12 +1121,59 @@ fn get_num_txns_ds_epoch(_params: Params, node: &Arc<Mutex<Node>>) -> Result<Str
...
+    todo!("API getshardingstructure is not implemented yet");
+}
+
+// GetSmartContractSubState
+fn getsmartcontractsubstate(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
+    todo!("API getsmartcontractsubstate is not implemented yet");
+}
+
+// GetSoftConfirmedTransaction
+fn getsoftconfirmedtransaction(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<GetTxResponse> {
```
JamesHinshelwood, line 1167: Would prefer `get_soft_confirmed_transaction` and similar for the other APIs, but happy for you to change that when you actually implement them :)

JamesHinshelwood (approved):

zilliqa/src/api/zil.rs
maxconway: Sure, will do when I do

Issue #1609, opened 2024-10-10
The node's current view should be persisted in the database


JamesHinshelwood:
It is incorrect for a node to send a `NewView` and a `Proposal` in the same view. Currently `self.view` is held in-memory in `Consensus`. This makes it possible for a node to send a `NewView` and `Proposal` for the same view, if it is restarted in between.

Instead, we should store the current view in the database and ensure we never equivocate by breaking the rule above.

This feature is also desirable for keeping the view of nodes in sync when many proposals are missed. There is an open question about whether we should send missed `NewView` messages when we start - Please discuss before implementing.

Issue #1709, opened 2024-10-25
Use timestamp to estimate amount of time a node has been down 


86667:
Expanding on [this work](https://github.com/Zilliqa/zq2/pull/1710) to reduce restart times when a network has been unable to find consensus and produce proposals.

Store timestamp with view and use to jump ahead at startup the minimum number of views which have passed since being down

PR #1710, opened 2024-10-25
BREAKING CHANGE: Persist node's current view in the database

86667:
Closes https://github.com/Zilliqa/zq2/issues/1609 and https://github.com/Zilliqa/zq2/issues/1709.

For the purpose of reducing re-sync time after consensus has not been reached for a while. 

For example, before these changes a node on `view: 30` with `finalised_view: 10` would start from `view: 10` if restarted. Now, they'll start back on `view :30`. If enough time has passed since they went down then they will jump ahead the minimum number of views.

These changes make breaking changes to the db

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -325,7 +325,7 @@ impl Consensus {
 
                     // If current_view was written to disk then always start from there. Otherwise start from (highest out of high block and finalised block) + 1
                     let start_view = db
-                        .get_current_view()?
+                        .get_latest_view()?
                         .or_else(|| {
                             Some(std::cmp::max(high_block.view(), finalized_block.view()) + 1)
                         })
@@ -334,9 +334,9 @@ impl Consensus {
                     let mut view = View::new(start_view);
 
                     // If timestamp of when current view was written exists then use it to estimate the minimum number of blocks the network has moved on since shut down
-                    if let Some(current_view_timestamp) = db.get_current_view_timestamp()? {
+                    if let Some(latest_view_timestamp) = db.get_latest_view_timestamp()? {
                         view.update_view_after_being_inactive(
-                            current_view_timestamp.elapsed()?,
+                            latest_view_timestamp.elapsed()?,
                             config.consensus.consensus_timeout.as_millis() as u64,
                         );
                     }
@@ -2592,7 +2592,7 @@ impl Consensus {
     /// Set view in memory and update tip_info.current_view in storage
     pub fn set_view(&mut self, view: u64) -> Result<()> {
         if self.view.set_view(view) {
-            self.db.set_current_view(view)?;
+            self.db.set_latest_view(view)?;
         }
         Ok(())
     }
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -213,11 +213,38 @@ impl View {
                 trace!("Tried to set view to same view - this is incorrect");
             }
         };
-        return false;
+        false
     }
 
     pub fn last_timeout(&self) -> SystemTime {
         self.last_timeout
+    }
+
+    /// Calculate how long we should wait before timing out for this view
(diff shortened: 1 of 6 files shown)


zilliqa/src/consensus.rs
```diff
@@ -2438,6 +2492,14 @@ impl Consensus {
         self.view.get_view()
     }
 
+    /// Set view in memory and update tip_info.latest_view in storage
```
JamesHinshelwood, line 2495: Do we need to keep view in memory at all? I'd prefer a single source-of-truth unless there's a good reason not to?

zilliqa/src/db.rs
```diff
@@ -253,6 +253,8 @@ impl Db {
                 PRIMARY KEY (address, tx_hash));
             CREATE TABLE IF NOT EXISTS tip_info (
                 latest_finalized_view INTEGER,
+                latest_view INTEGER,
+                latest_view_timestamp BLOB,
```
JamesHinshelwood, line 257: nit: I quite like the convention of suffixing timestamp DB columns with `_at` - i.e. `latest_view_updated_at`?

JamesHinshelwood: SQLite has a `TIMESTAMP` data type - Should we use it?

zilliqa/src/db.rs
```diff
@@ -253,6 +253,8 @@ impl Db {
                 PRIMARY KEY (address, tx_hash));
             CREATE TABLE IF NOT EXISTS tip_info (
                 latest_finalized_view INTEGER,
+                latest_view INTEGER,
```
JamesHinshelwood, line 256: I wonder if `latest_view` might be a slightly misleading name. It looks very similar to `latest_finalized_view` which is pointing to our latest finalized block, however `latest_view` won't ever point to an actual block (by design).

I can't think of anything good though, other than just `view`?

zilliqa/src/consensus.rs
```diff
@@ -1893,10 +1946,11 @@ impl Consensus {
                     new_high_qc_block_view + 1,
                     current_high_qc_view,
                 );
+                //TODO write high_qc and current_view in one db call
```
JamesHinshelwood, line 1949: Is it the case that we always call `set_view` before sending a `Proposal`/`View`/`NewView` for a given view?

zilliqa/src/db.rs
86667: Updated column name to `view_updated_at`. I'm not sure why weve used BLOB rather than timestamp but it does allow for simpler conversion to/from. We could move over to TIMESTAMP but i see no advantage to doing so?

zilliqa/src/db.rs
86667: Agreed, updated to `view`

zilliqa/src/consensus.rs
```diff
@@ -279,28 +235,45 @@ impl Consensus {
...
+                        .unwrap();
+
                     trace!(
                         "recovery: high_block view {0}, finalized_number {1} , start_view {2}",
                         high_block.view(),
                         finalized_number,
                         start_view
                     );
 
+                    // If timestamp of when current view was written exists then use it to estimate the minimum number of blocks the network has moved on since shut down
```
JamesHinshelwood, line 267: I think when doing this catch up we should also send the `NewView`s for the missed views that we didn't send. Otherwise we still need to wait for the next view to progress before the nodes will re-sync.

We could discuss and make that change in a separate PR if you prefer.

JamesHinshelwood (approved):
LGTM :)

zilliqa/src/consensus.rs
86667: Will create a separate issue.

JamesHinshelwood (approved):

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #1720, opened 2024-10-28
fix: Do not stop proposal of early_proposal when buffered votes are counted

86667:
After dropping and re-syncing it is possible that the current leader receives votes and adds them to their `buffer_votes` because they are not yet caught up to the view which the votes apply to. 

In this circumstance there is a bug which causes `create_next_block_on_timeout` to be set to `false` after having just been set to `true`, and thus the proposal is never sent.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -727,6 +727,14 @@ impl Consensus {
                             ExternalMessage::Proposal(Proposal::from_parts(block, transactions)),
                         )));
                     }
+                    // A bit hacky: processing of our buffered votes may have resulted in an early_proposal be created and awaiting empty block timeout for broadcast. In this case we must return now
+                    if self.create_next_block_on_timeout
+                        && self.early_proposal.is_some()
+                        && self.early_proposal.as_ref().unwrap().0.view() == proposal_view + 1
+                    {
+                        trace!("supermajority reached, early proposal awaiting broadcast");
+                        return Ok(None);
+                    }
                 }
 
                 // If we reach this point, we had some buffered votes but they were not sufficient to reach a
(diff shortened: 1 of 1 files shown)


JamesHinshelwood (approved):

zilliqa/src/consensus.rs
```diff
@@ -720,14 +720,18 @@ impl Consensus {
...
 
                 let Some(next_leader) = next_leader else {
                     warn!("Next leader is currently not reachable, has it joined committee yet?");
                     return Ok(None);
                 };
 
+                if next_leader.public_key != self.public_key() {
+                    // Leader for this view is not us so clear any early work
+                    self.create_next_block_on_timeout = false;
+                    self.early_proposal = None;
```
JamesHinshelwood, line 732: Is it possible that `early_proposal` contains some state which we do need to wipe even if we are the next proposer as well?

86667: It is possible, and is handled in `assemble_early_proposal_at()` by overwriting `early_proposal` unless:

```
        if self.early_proposal.is_some()
            && self.early_proposal.as_ref().unwrap().0.view() == self.view()
        {
            return Ok(());
        }
```

86667:
Alright, what is pushed is the solution to the problem described but is a bit hacky. 

Ultimately, the problem we have is that proposals can be broadcast via `timeout()` or immediately depending on whether we have passed the block timeout. From `vote()` we return `Ok(None)` in the former case and `Ok(Propsal)` in the latter. 

Im going to spend a bit of time attempting to remove this and have all proposal broadcasting happen in `timeout()` but if the change proves to be large i'll push for merging this as is.

JamesHinshelwood (approved):

PR #1722, opened 2024-10-29
Fix/api response consistency zq1 and  zq2

chetan-zilliqa:


zilliqa/src/api/zil.rs
```diff
@@ -448,6 +449,12 @@ fn get_smart_contract_state(params: Params, node: &Arc<Mutex<Node>>) -> Result<V
 
     let state = node.get_state(&block)?;
     let account = state.get_account(address)?;
+    if account == Account::default() {
```
JamesHinshelwood, line 452: Can we do this check with `state.has_account` instead? Otherwise if the account exists, but has zero balance we will return an error, whereas I guess ZQ1 would return the account?

JamesHinshelwood (approved):

PR #1723, opened 2024-10-29
Interop gas charging workaround 

bzawisto:
This PR introduces a workaround and performs different gas tracking for evm->scilla calls. In order to figure out if such logic should be applied, each account has additional creation_at_block field saying if it's ZQ1 or ZQ2(block_number). 
Since we need to know how much gas is left inside the precompile - each evm run has thin ZQ2Inspector which tracks gas left and stores the value in PendingSubstate. Later, PendingSubstate is accessible from precompile and we can provide suitable value to scilla runs from precompile.


86667 (approved):
:+1:

zilliqa/src/precompiles/scilla.rs
```diff
@@ -333,24 +333,61 @@ pub fn scilla_call_handle_register<I: Inspector<PendingState> + ScillaInspector>
...
+                inputs.gas_limit
+            }
+        };
+
+        let gas = Gas::new(gas_limit);
         let outcome =
             scilla_call_precompile(&inputs, gas.limit(), &mut ctx.evm.inner, &mut ctx.external);
 
+        let result_gas = {
+            if zq1_gas_rule {
```
JamesHinshelwood, line 374: Are you sure the ZQ1 implementation charges 0 gas? I thought it just charged less gas?

zilliqa/src/inspector.rs
```diff
@@ -350,3 +356,91 @@ impl ScillaInspector for OtterscanOperationInspector {
...
+
+    fn log(&mut self, interp: &mut Interpreter, context: &mut EvmContext<PendingState>, log: &Log) {
+        self.inner.log(interp, context, log);
+    }
+    fn call(
+        &mut self,
+        context: &mut EvmContext<PendingState>,
+        inputs: &mut CallInputs,
+    ) -> Option<CallOutcome> {
+        Inspector::call(&mut self.inner, context, inputs)
```
JamesHinshelwood, line 398: If these methods are not changing the default implementation from the `Inspector` trait, we should not list them here.

zilliqa/src/consensus.rs
```diff
@@ -2592,6 +2589,10 @@ impl Consensus {
         &self.state
     }
 
+    pub fn state_mut(&mut self) -> &mut State {
```
JamesHinshelwood, line 2592: Why do we expose this? Its deliberate that `state` isn't `pub` to ensure `Consensus` is the only one who can mutate it.

zilliqa/src/exec.rs
```diff
@@ -1036,6 +1068,7 @@ impl PendingState {
         PendingState {
             pre_state: state,
             new_state: HashMap::new(),
+            gas_left: Gas::new(0),
```
JamesHinshelwood, line 1071: Where is this used? I tried to find it in the precompile implementation but couldn't. I'm slightly confused as to what this is actually for?

JamesHinshelwood:
We ended up using #1933

PR #1724, opened 2024-10-29
Implement GetTransactionStatus API

### LINKED ISSUE #1551: GetTransactionStatus
### TODO list

- [x] Write tests
- [x] Draft implementation
- [ ] Finalize and merge

maxconway:
- Also changed some API placeholder names to snakecase

Transaction status code is returned as:
- 3 if the transaction receipt shows successfully accepted,
- 1 if it is not yet accepted
- my guess at the most appropriate error from the docs if there is one error
- 255 (internal error) for multiple errors

As far as I can see, ZQ2 transactions and transaction receipts don't have the same information in them that the ZQ1 GetTransactionStatus API docs expect, so my approach here was to try and get as close as possible while still matching the docs, and presumably we can then later add a new API which matches ZQ2.

Closes #1551


zilliqa/src/api/types/zil.rs
```diff
@@ -618,6 +618,136 @@ pub struct TransactionReceiptResponse {
...
+    pub epoch_updated: String,
+    #[serde(rename = "gasLimit")]
+    pub gas_limit: String,
+    #[serde(rename = "gasPrice")]
+    pub gas_price: String,
+    #[serde(rename = "lastModified")]
+    pub last_modified: String,
+    #[serde(rename = "modificationState")]
+    pub modification_state: u64,
+    pub status: u64,
```
JamesHinshelwood, line 641: It might be useful to link to https://github.com/Zilliqa/Zilliqa/blob/master/src/common/TxnStatus.h#L23 somewhere so we know what `1`, `3`, etc. mean.

zilliqa/src/api/types/zil.rs
```diff
@@ -618,6 +618,136 @@ pub struct TransactionReceiptResponse {
...
+        } else if receipt.accepted.is_none() {
+            1
+        } else {
+            let errors: Vec<ScillaError> =
+                receipt.errors.into_iter().flat_map(|(_k, v)| v).collect();
+            if errors.len() == 1 {
+                match errors[0] {
+                    ScillaError::CallFailed => 11,
+                    ScillaError::CreateFailed => 255,
+                    ScillaError::OutOfGas => 0,
```
JamesHinshelwood, line 717: It seems like this should be `22`? Maybe we should use an enum or at least some `const`s for these magic numbers and refer back to the ZQ1 source?

Issue #1509, opened 2024-09-27
CreateTransaction validates gasLimit


rrw-zilliqa:
in ZQ1 it was possible to successfully submit transactions with very high gas limits (eg. 100_000) - and in some cases this was necessary because ZQ1 gas estimation wasn't reliable.

ZQ2 will validate gasLimit and reject the transaction if balance < gasLimit * gasPrice + amount . This means that transactions which advertise enormous gas limits will generally fail.

The ZQ2 way is technically correct, but may interfere with existing applications; we might consider accepting the transaction failing on execution if the user runs out of credit to pay the actual gas.

rrw-zilliqa:
We've decided (17/10) to act on this by exempting Scilla transactions from the balance check validation.

DrZoltanFazekas:
Is the exemption implemented #1726 ? If so, we should close this issue.

JamesHinshelwood:
Assume this was implemented by #1726

PR #1726, opened 2024-10-29
Improve Scilla gas charging

rrw-zilliqa:
Fixes (I hope!) #1509


zilliqa/src/exec.rs
```diff
@@ -506,6 +502,20 @@ impl State {
         Ok((e, state.finalize(), cfg.env))
     }
 
+    // The rules here are somewhat odd, and inherited from ZQ1
```
JamesHinshelwood, line 505: @bzawisto Could you double-check these rules are consistent with how we figured Ethereum transactions charge gas when the gas limit is reached?

PR #1733, opened 2024-10-31
Speed up tests in CI

JamesHinshelwood:
A few changes to speed up our CI times:
* Remove the sharding tests. This isn't a launch feature for us, so I don't want to pay the cost of running these tests before it is useful. Also, the tests were written in quite a slow way and I'd like to rewrite them before re-adding.
* Only run 20 iterations of each test instead of 100.
* Build and run in debug mode rather than release mode. Since we've decreased the iteration count, debug mode is now faster in total.


86667 (dismissed):
:ok_hand:

DrZoltanFazekas (approved):

PR #1735, opened 2024-11-01
Fix overflow in `create_transaction_errors` test

JamesHinshelwood:
This bug causes 1/256th of test runs to fail with overflow checks enabled.


86667 (approved):

86667:
This one got me earlier. Cheers

Issue #1746, opened 2024-11-04
Implement `getControlAddress()`


DrZoltanFazekas:
Much like there is a `getRewardAddress()` function to retrieve the reward address of a validator identified by its BLS pubkey, there should be a `getControlAddress()` function to retrieve the signer address (aka control address).

Issue #1745, opened 2024-11-04
Implement `totalStake()`


DrZoltanFazekas:
Before the recent changes to the `Deposit` contract there was a  public variable `totalStake` that was moved inside the `Committee` struct and is not accessible anymore. The same getter function shall be implemented again and return the total stake of the current committee.

Issue #1301, opened 2024-08-14
Control and reward address changes for validators


theo-zil:
Add methods in `deposit.sol` to allow validators to change their `rewardAddress` and `controlAddress`.

DrZoltanFazekas:
The `setControlAddress()` function is also used by the delegated staking contracts both liquid and non-liquid variant) to change their `controlAdress` to their `owner()` instead of their own address that was set in the `deposit()` function

DrZoltanFazekas:
Currently, I implemented two separate functions `setControlAddress()` and `setRewardsAddress()` that expect the BLS pubkey of the staker as an argument. In a later PR we can change this to functions without arguments that detect the staker based on the `msg.sender` i.e. the current control address:
```
        bytes storage stakerKey = _stakerKeys[msg.sender];
        require(stakerKey.length != 0, "staker does not exist");
        Staker storage staker = _stakersMap[stakerKey];

```

We can also add a function which allows to change both the reward address and the control address in a single transaction or make the change of the control address a two-step process where the new control address has to confirm the change for it to become effective.

PR #1747, opened 2024-11-04
Add getter and setter functions to `Deposit` contract

DrZoltanFazekas:
Implement `getTotalStake()` as per #1745, `getControlAddress()` as per #1746, as well as `setControlAddress()` and `setRewardAddress()` as defined in #1301

### DIFF
--- a/zilliqa/src/contracts/deposit.sol
+++ b/zilliqa/src/contracts/deposit.sol
@@ -163,6 +163,12 @@ contract Deposit {
 
     uint64 public blocksPerEpoch;
 
+    modifier onlyControlAddress(bytes calldata blsPubKey) {
+        require(blsPubKey.length == 48);
+        require(_stakersMap[blsPubKey].controlAddress == msg.sender, "sender is not the control address");
+        _;
+    }
+
     constructor(
         uint256 _minimumStake,
         uint256 _maximumStakers,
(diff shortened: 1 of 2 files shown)


zilliqa/src/contracts/deposit.sol
```diff
@@ -273,6 +282,10 @@ contract Deposit {
         return committee().stakerKeys;
     }
 
+    function getTotalStake() public view returns (uint256) {
```
JamesHinshelwood, line 285: Related but not part of this PR - I would suggest removing `totalStake` from `Committee`. The maximum number of stakes you need to add is only `256` and caching that result here just adds a potential inconsistency bug. In fact `consensus.rs` ignores the `totalStake` and calculates it for itself anyway.

zilliqa/src/contracts/deposit.sol
```diff
@@ -311,6 +324,30 @@ contract Deposit {
...
+        address rewardAddress
+    ) public onlyControlAddress(blsPubKey) {
+        _stakersMap[blsPubKey].rewardAddress = rewardAddress;
+    }
+
+    function setControlAddress(
+        bytes calldata blsPubKey,
+        address controlAddress
+    ) public onlyControlAddress(blsPubKey) {
+        _stakersMap[blsPubKey].controlAddress = controlAddress;
```
JamesHinshelwood, line 348: In future we could consider 2-step control transfer to avoid transferring control to an address that isn't callable (e.g. because no one has the secret key).

PR #1751, opened 2024-11-05
Implement GetSmartContractSubState

### LINKED ISSUE #1545: GetSmartContractSubState
### TODO list

- [x] Write tests
- [x] Draft implementation
- [ ] Finalize and merge

### LINKED ISSUE #1671: Z2-Converter API Mainnet : GetSmartContractSubState
### Issue Summary:
Calling `GetSmartContractSubState` on the ZQ1 and ZQ2 has the following diff.
Here are the URLs for ZQ1 and ZQ2.
ZQ2 : http://34.124.218.159:4201/
ZQ1 : http://34.124.185.137:4201/
### Result:
**Missing keys:**
```
```
**Different Keys:**
```
 ZQ2 Output (GetSmartContractSubState_output_ZQ2.json):
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32601,
    "message": "Method not found"
  }
}

 ZQ1 Output (GetSmartContractSubState_output_zilliqa.json):
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "bnum_req": "30800"
  }
}


```
### Observations:
GetSmartContractSubState API is not implemented in ZQ2.

###  Conclusion:
If the above differences are expected and align with the migration to ZQ2, no further action is required. However, if these differences are unexpected, they may indicate issues that must be addressed.

maxconway:
Based on GetSmartContractState, with added filters
Plus a test
Closes #1671 
Closes #1545

### DIFF
--- a/zilliqa/src/api/zil.rs
+++ b/zilliqa/src/api/zil.rs
@@ -33,8 +33,8 @@ use super::{
         self, BlockchainInfo, DSBlock, DSBlockHeaderVerbose, DSBlockListing, DSBlockListingResult,
         DSBlockRateResult, DSBlockVerbose, GetCurrentDSCommResult, MinerInfo,
         RecentTransactionsResponse, SWInfo, ShardingStructure, SmartContract, TXBlockRateResult,
-        TransactionBody, TxBlockListing, TxBlockListingResult, TxnBodiesForTxBlockExResponse,
-        TxnsForTxBlockExResponse,
+        TransactionBody, TransactionStatusResponse, TxBlockListing, TxBlockListingResult,
+        TxnBodiesForTxBlockExResponse, TxnsForTxBlockExResponse,
     },
 };
 use crate::{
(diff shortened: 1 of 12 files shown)


zilliqa/src/api/zil.rs
```diff
@@ -1242,9 +1191,77 @@ fn get_sharding_structure(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<(
...
+
+            let (var_name, indices) = split_storage_key(&k)?;
+            if let Some(ref x) = requested_varname {
+                if *x != var_name {
+                    continue;
+                }
+            }
+            let mut var = result.entry(var_name.clone());
+
+            for index in indices {
```
JamesHinshelwood, line 1237: The indices are ordered and easily seekable. We could save a lot of work by only querying the indices we need from the state.

As it stands, `GetSmartContractSubState` requires almost the same amount of work for us as `GetSmartContractState`.

zilliqa/tests/it/zil.rs
```diff
@@ -1926,9 +1926,40 @@ async fn get_sharding_structure(mut _network: Network) {
...
+    assert!(
+        response["_balance"]
+            .as_str()
+            .unwrap()
+            .parse::<u64>()
+            .is_ok(),
+        "Invalid balance format"
+    );
+
+    // Verify the admins field if it exists
```
JamesHinshelwood, line 1955: Small nit - It might be sane to verify that no other fields are included in the response. Otherwise, I think the `GetSmartContractState` (not `SubState`) would also pass this test.

Issue #1676, opened 2024-10-22
Z2-Converter API Mainnet : GetTxBlock


chetan-zilliqa:
### Issue Summary:
Calling `GetTxBlock` on the ZQ1 and ZQ2 has the following diff.
Here are the URLs for ZQ1 and ZQ2.
ZQ2 : http://34.124.218.159:4201/
ZQ1 : http://34.124.185.137:4201/
### Result:
**Missing keys:**
```
Keys present in ZQ1 but missing in ZQ2:
  root['result']['header']['DSBlockNum']

Keys present in ZQ2 but missing in ZQ1:
  root['result']['header']['DsBlockNum']

```
**Different Keys:**
```
Values that differ between ZQ2 and ZQ1:
  root['result']['body']['BlockHash']:
    ZQ2: 7c7c754408e16f11b93921ce9ee6aa704cd3189e06d9b2e141efbd3ae0f93b35
    ZQ1: 424da7e159b0a4504ce307952b285b7eb6ed2693df41cc81a8e96862f58c4b9b
  root['result']['body']['HeaderSign']:
    ZQ2: 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
    ZQ1: 0F2902579BD02FACE1778FDAAC4872000DD3C463A82405354CC342DABC64CD001D8DA143F0688A520E21E7382CEC7C366FA1D9F98C0B6B0F9A946CC4F67AA0B4
  root['result']['header']['MbInfoHash']:
    ZQ2: 0000000000000000000000000000000000000000000000000000000000000000
    ZQ1: 2f9dd0a266f5ef9a4294ee1f961b096e4af7f9da3cb878d88ee1e89a54ce8759
  root['result']['header']['MinerPubKey']:
    ZQ2: 0x000000000000000000000000000000000000000000000000000000000000000000
    ZQ1: 0x020177703339FDD3C56E3A5E4BAEDB253299C53BFC7A213679F9AD81ABA5BAB024
  root['result']['header']['NumMicroBlocks']:
    ZQ2: 0
    ZQ1: 4
  root['result']['header']['PrevBlockHash']:
    ZQ2: e29d9c1aefde7728284a240584045b33e67f54cd061611401d0313f910b4f038
    ZQ1: 4ab87748ca5aaf0dd451c01f877ff92b4dd918a8c31fdf95872ff90368d76a62
  root['result']['header']['StateDeltaHash']:
[...]

PR #1752, opened 2024-11-05
Some fixes for get_tx_block

maxconway:
- Added code to get miner address
- Added comments for variables that appear obsolete in ZQ2
- Set version number to 2 (it was 1 in ZQ1 so I'm assuming it should now be 2)
- Retrieved correct values for gas_limit and gas_used
- Refactored
- Added test

This seemed to be enough changes to be worth a PR, but there are still issues with incorrect hashes which I haven't addressed yet.

Addresses #1676

### DIFF
--- a/zilliqa/src/api/types/zil.rs
+++ b/zilliqa/src/api/types/zil.rs
@@ -21,7 +21,7 @@ use crate::{
     serde_util::num_as_str,
     time::SystemTime,
     transaction::{
-        self, ScillaGas, SignedTransaction, TransactionReceipt, VerifiedTransaction, ZilAmount,
+        ScillaGas, SignedTransaction, TransactionReceipt, VerifiedTransaction, ZilAmount,
     },
 };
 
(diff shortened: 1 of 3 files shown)


zilliqa/src/api/types/zil.rs
```diff
@@ -31,14 +31,16 @@ pub struct TxBlock {
...
+    pub fn new(block: &Block, proposer: Address) -> Self {
         // TODO(#79): Lots of these fields are empty/zero and shouldn't be.
+        let mut scalar = [0; 32];
+        scalar[31] = 1;
         TxBlock {
             header: TxBlockHeader {
-                version: 0,
-                gas_limit: transaction::ScillaGas(0),
-                gas_used: 0,
+                version: 2,                                    // ZQ2
```
JamesHinshelwood, line 41: If we're doing this - Should we take care to keep the `version` for converted ZQ1 blocks as `0`? Or alternatively just don't expose the difference between ZQ1 and ZQ2 blocks in this field (it is implied by block height anyway).

maxconway: This might be another case of something we don't want to bring over from ZQ1 - my reasoning was that there was a version field there, so if we also return a version number in ZQ2, it's less likely to break anyone's code that depends on it.

Overall it's a question of how much backwards compatibility we're targetting

PR #1762, opened 2024-11-07
Some GetTransaction bugfixes

maxconway:
- event logs field now always returned
- accepted field now always returned
- transitions field now always returned
- added test

Not addressed:
- nonces sometimes different (looks like its outside the API code)
- invalid signature errors (looks like its outside the API code)
- version different (presumed acceptable)
- capitalisation different (presumed acceptable)
- cumulative_gas sometimes zero (looks like its outside the API code)

Relates to #1666

### DIFF
--- a/zilliqa/src/api/types/zil.rs
+++ b/zilliqa/src/api/types/zil.rs
@@ -202,68 +202,48 @@ impl GetTxResponse {
             SignedTransaction::Zilliqa { tx, sig, key } => (
                 ((tx.chain_id as u32) << 16) | 1,
                 tx.to_addr,
-                key.to_encoded_point(true)
-                    .as_bytes()
-                    .to_hex()
-                    .to_uppercase(),
-                <[u8; 64]>::from(sig.to_bytes()).to_hex().to_uppercase(),
+                key.to_encoded_point(true).as_bytes().to_hex(),
+                <[u8; 64]>::from(sig.to_bytes()).to_hex(),
                 (!tx.code.is_empty()).then_some(tx.code),
-                (!tx.data.is_empty())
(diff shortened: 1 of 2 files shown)


zilliqa/src/api/types/zil.rs
```diff
@@ -291,7 +289,7 @@ impl GetTxResponse {
                     })
                     .collect(),
                 success: receipt.success,
-                accepted: receipt.accepted,
+                accepted: receipt.accepted.unwrap_or(false),
```
JamesHinshelwood, line 292: This suggests that ZQ1 returns `accepted: false` for:
* Eth transactions
* Scilla contract creations

Have we double-checked that is definitely the case?

maxconway: No, my assumption was that a missing accepted field was an error state.
We could do `accepted: receipt.accepted.unwrap_or(receipt.success)` instead?
The reason for doing this is was that we're always returning an accepted value to match ZQ1, so we could also just not.

JamesHinshelwood: I guess my question is what does ZQ1 return as `accepted` for transactions which don't have a natural value? (for context - `accepted` in Scilla-land means "the Scilla contract you sent ZIL to accepted the funds")

maxconway: #1666 Suggests that ZQ1 returns false where the previous ZQ2 implementation would have returned nothing since receipt.accepted == none. But I don't necessarily know the details of transaction 4964bf50c8f59... so I might be misunderstanding. Does that match what you thing that transaction does, @chetan-zilliqa

PR #1765, opened 2024-11-07
Don't verify signed txn for zil api

bzawisto:


chetan-zilliqa (approved):

chetan-zilliqa (dismissed):

chetan-zilliqa (approved):

JamesHinshelwood (commented):
Why would we be storing transactions in our database that don't verify? IMO we shouldn't be and if we are there is a bug somewhere else..?

JamesHinshelwood:
Thanks @bzawisto!

Is it easy to tell how many bad transactions there are? If the number is low maybe we can list the special cases explicitly...

Issue #1756, opened 2024-11-06
Z2-Converter : eth_getTransactionReceipt has output difference for ZQ1 and ZQ2


chetan-zilliqa:
ZQ2:
```
curl --request POST \
  --url http://34.124.218.159:4201/ \
  --header 'Content-Type: application/json' \
  --data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "eth_getTransactionReceipt",
    "params": ["0x6d31bddd1618a761dce907ed501652bd57131e57eaccdad49372cc58bd9dda65"]
}'
```

ZQ1:

```
curl --request POST \
  --url http://34.124.185.137:4201/ \
  --header 'Content-Type: application/json' \
  --data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "eth_getTransactionReceipt",
    "params": ["0x6d31bddd1618a761dce907ed501652bd57131e57eaccdad49372cc58bd9dda65"]
}'
```

Here is the difference:
```
--- Viewing difference for eth_getTransactionReceipt (0x6d31bddd1618a761dce907ed501652bd57131e57eaccdad49372cc58bd9dda65) ---

Keys present in ZQ1 but missing in ZQ2:
  root['result']['r']
  root['result']['root']
  root['result']['s']
  root['result']['v']

Values that differ between ZQ2 and ZQ1:
  root['result']['blockHash']:
    ZQ2: 0xa9684bf32e8c189084f7a7da8a7ee4ff196050cb98a5913f7185bcdd8a068b02
    ZQ1: 0x15af3fae9524d270548fe3f5cb4568b49395bb7bfb3e3cd8024fa10948505daa
  root['result']['cumulativeGasUsed']:
    ZQ2: 0xfb
    ZQ1: 0x19bcc
  root['result']['gasUsed']:
    ZQ2: 0xfb
    ZQ1: 0x19bcc

```

chetan-zilliqa:
The following keys are missing and different.

```
Keys present in ZQ1 but missing in ZQ2:
  root['result']['root']

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>



Values that differ between ZQ2 and ZQ1:
  root['result']['cumulativeGasUsed']:
    ZQ2 -> 0xfb
    ZQ1 -> 0x19bcc
  root['result']['gasUsed']:
    ZQ2 -> 0xfb
    ZQ1 -> 0x19bcc
  root['result']['r']:
[...]

PR #1766, opened 2024-11-07
Add fields to get transaction receipt

maxconway:
Adds r, s, v and root fields to getTransactionReceipt, along with a test

References #1756 but it doesn't fix the differences in block hashes or gas used fields


zilliqa/src/api/types/eth.rs
```diff
@@ -333,6 +333,14 @@ pub struct TransactionReceipt {
...
     #[serde(serialize_with = "bool_as_int")]
     pub status: bool,
+    #[serde(serialize_with = "hex")]
+    pub v: u64,
+    #[serde(serialize_with = "hex")]
+    pub r: U256,
+    #[serde(serialize_with = "hex")]
+    pub s: U256,
+    #[serde(serialize_with = "hex")]
+    pub root: B256,
```
JamesHinshelwood, line 343: Is this field documented anywhere? I can't find references to it in other docs.

maxconway: Its returned by ZQ1, apparently, according to #1756. I assumed it meant the transaction root?

JamesHinshelwood: I found 

> The post-transaction state root. Only specified for transactions included before the Byzantium upgrade.

on https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_gettransactionreceipt/

This doesn't make sense to return since we aren't following the same forks as Ethereum mainnet and we can assume all Eth transactions are 'post-Byzantium'.

More generally we should be careful to assume that just because ZQ1 returned something, we should do the same. Lots of what ZQ1 does is not sensible and we should avoid copying across that insensibility.

JamesHinshelwood: Yes - I think these are part of the other Eth specs.

maxconway: Right, I've removed root in the latest push

JamesHinshelwood (approved):

JamesHinshelwood (approved):

JamesHinshelwood (approved):

JamesHinshelwood (approved):

JamesHinshelwood (approved):

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #1777, opened 2024-11-08
Group related dependabot updates together

JamesHinshelwood:
Often these dependencies need to updated in lock-step to compile (since they have public dependencies on each other).


rrw-zilliqa (approved):
Looks good to me :-)

Issue #1525, opened 2024-09-30
(Empty) block time increases to 2s in a 60 validator network


DrZoltanFazekas:
The block interval of creating empty blocks increased from 1s average to 2s average in the `perftest`, a network with

- 30 validators in `asia-southeast1`
- 15 validators in `europe-west1`
- 15 validators in `europe-west2`

The network was stopped to save costs, but @frankmeds can restart it any time someone is ready to investigate this issue.

shawn-zil:
Some quick measurements.

iperf3 between `europe-west2` and `europe-west1`:
```
[  5]   0.00-10.04  sec  1.11 GBytes   948 Mbits/sec                  receiver

                                        "max_rtt":      10002,
                                        "min_rtt":      9609,
                                        "mean_rtt":     9786,
```

iperf3 between `asia-southeast1` and `europe-west1` :
```
[  5]   0.00-10.17  sec   183 MBytes   151 Mbits/sec                  receiver

                                        "max_rtt":      163871,
                                        "min_rtt":      162603,
                                        "mean_rtt":     163059,
```

shawn-zil:
The way that gossipsub works is that it transmits the message to at least `mesh_n` nodes; and the default configuration for `mesh_n` is 6 nodes.

Given that the validators are distributed with `1:1` ratio between `europe*` and `asia*`; and selected randomly without bias; we can assume that gossipsub would normally select 3-nodes in `asia*` and 3-nodes in `europe*` for broadcast.

To broadcast a Proposal, the network transmission time could take up to `3 * ~168ms + 3 * ~10ms` or `~534ms`; assuming messages are sent sequentially.

Given the size of the network is 60 nodes; to ensure that a Proposal reaches enough nodes to form a quorum requires at least 2 perfect (that do not have overlapping nodes) hops i.e. `6 + 6*6 = 42`. It is more likely that it would take 3-hops (with some overlap).

If we take the most ideal case, i.e. the messages are transmitted concurrently, and it only requires 2 perfect hops, it would still require `504ms` for the original Proposal and it's Votes to reach the next Leader. In the worst case, i.e. the messages are transmitted sequentially and it requires 3 hops, it would require > `1.5s` for the same.

Therefore, it is unsurprising that block times can take 2s for empty blocks.

shawn-zil:
[...]

PR #1792, opened 2024-11-12
Faster vote handling

shawn-zil:
Analysis of the vote handling process on local, finds that in the case of a 4-validator network, it takes about 25ms to handle a vote. Of this, ~13ms is spent in `total_weight()` and ~7ms is spent in `get_stakers()` calls to the EVM. This represents about 80% of the overall vote handling time. Increasing the number of validators also increases the vote handling time (as observed earlier).

This PR introduces some changes:
1. Rewrite of `total_weight()` from O(n) to O(1) time, which reduced the handling time from 13ms to ~3ms by calling the `getTotalStake()` deposit contract; then adding a caching mechanism reduced this further to a negligible cost after the first call.
2. Adding caching to `get_stakers()`, which reduces the call cost to negligible, after the first call.

This brings into the question of whether it might make sense to build a generalised caching mechanism for all EVM calls. But that might be overkill for now, and is something to think about if the need arises.

I think it's fine to merge these changes first, since they definitely provide some measure of improvement to the vote handling process. Hopefully, we can test and measure the real impact of these changes once the present devops issues have been resolved.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -412,8 +412,8 @@ impl Consensus {
                 ..block.header
             };
             let root_hash = self.state.root_hash()?;
-            let stakers = self.get_committee(root_hash, next_block_header)?; // self.state.get_stakers(next_block_header)?;
-                                                                             // If we're in the genesis committee, vote again.
+            let stakers = self.get_stakers_mut(root_hash, next_block_header)?; // self.state.get_stakers(next_block_header)?;
+                                                                               // If we're in the genesis committee, vote again.
             if stakers.iter().any(|v| *v == self.public_key()) {
                 info!("timeout in view: {:?}, we will vote for block rather than incrementing view, block hash: {}", view, block.hash());
                 let leader = self.leader_at_block(&block, view).unwrap();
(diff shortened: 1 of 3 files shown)


86667 (approved):
Seems reasonable. Very nice speedup

JamesHinshelwood:
@shawn-zil Do you have a benchmark or easily reproducible way of testing the vote processing speed? I'm keen on exploring options to do the caching at a slightly lower level which avoids adding even more state and complexity into `Consensus`.

I'm testing a state cache in devnet this morning - https://github.com/Zilliqa/zq2/commit/61e3ae91ed75a67917cfb06c7e0466000e565939

Can you test the same change with your specific vote-processing issue?

shawn-zil:
> @shawn-zil Do you have a benchmark or easily reproducible way of testing the vote processing speed? I'm keen on exploring options to do the caching at a slightly lower level which avoids adding even more state and complexity into `Consensus`.
> 
> I'm testing a state cache in devnet this morning - https://github.com/Zilliqa/zq2/commit/61e3ae91ed75a67917cfb06c7e0466000e565939
> 
> Can you test the same change with your specific vote-processing issue?

@JamesHinshelwood 
I'm not doing anything fancy, just using timestamp.elapsed and printing it out at different points of the function to take readings.

Sorry.

JamesHinshelwood:
Once I've deployed my change to devnet, would you mind checking the performance there too to see if it helps with your issue?

shawn-zil:
Also impacts #1525

PR #1793, opened 2024-11-12
Rewrite of GetSmartContractSubState

maxconway:
This rewrite uses the storage trie for queries rather than just iterating over elements. 
Should be faster for querying small numbers of variables from large contract states, but this is not benchmarked.
Also added a more in-depth test

In #1751 (which added GetSmartContractSubState originally) the GetSmartContractState function was refactored so that it could share a common for with GetSmartContractSubState, but this PR makes enough changes to GetSmartContractSubState that this no longer makes sense, so the code sharing has been removed and GetSmartContractState now contains all its own code, though it is called for degenerate cases of GetSmartContractSubState.

This is based on discussion [here](https://github.com/Zilliqa/zq2/pull/1751#discussion_r1834648464) and [here](https://github.com/Zilliqa/zq2/pull/1751#discussion_r1834649288)


zilliqa/src/api/zil.rs
```diff
@@ -1209,56 +1270,32 @@ fn get_smart_contract_state_internal(
...
-                var = next.entry(key.clone());
-            }
-
-            var.or_insert(serde_json::from_slice(&v)?);
+        let indicies_encoded = requested_indices
+            .iter()
+            .map(|x| serde_json::to_vec(&x))
+            .collect::<std::result::Result<Vec<_>, _>>()?;
+        let k = storage_key(&var_name, &indicies_encoded);
+        let v = match trie.get(&k)? {
```
JamesHinshelwood, line 1281: I don't recall the intended behaviour of this API but do we need to handle the case where the user passes a requested index as a prefix and expects us to return many keys with that prefix? @chetan-zilliqa might be able to advise on some example requests and responses from ZQ1.

maxconway: Yeah if there's some examples to give more info on what's the desired behaviour here that'd be good because the documentation isn't super detailed for this one so I was filling in the blanks a little. I was also guessing that we needed `_balance` returned from this too.

chetan-zilliqa: Fetching a complete contract state is cumbersome and bandwidth intensive. If a user is only interested in substate(for eg: perticular contract variable or key) then it can be fetched from the contract.
Eg:
```
curl --request POST \
  --url https://api.devnet.zilliqa.com/ \
  --header 'Content-Type: application/json' \
  --data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "GetSmartContractSubState",
	"params": ["zil15tjx2l0gzzxaxucwk50stgw5smthhckl8qcqmf","bnum_req", []]
}'

curl --request POST \
  --url https://api.devnet.zilliqa.com/ \
  --header 'Content-Type: application/json' \
  --data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "GetSmartContractSubState",
	"params": ["zil15lr86jwg937urdeayvtypvhy6pnp6d7p8n5z09","ssnlist", []]
}'

curl --request POST \
  --url https://api.devnet.zilliqa.com/ \
  --header 'Content-Type: application/json' \
  --data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "GetSmartContractSubState",
	"params": ["0xa7c67d49c82c7dc1b73d231640b2e4d0661d37c1","deposit_amt_deleg", ["0xec31349b7f17ea28ff59c7586c7734be23e43443"]]
}'
curl --request POST \
  --url https://api.devnet.zilliqa.com/ \
  --header 'Content-Type: application/json' \
  --data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "GetSmartContractSubState",
	"params": ["0xa7c67d49c82c7dc1b73d231640b2e4d0661d37c1","balances", ["0x406111ecfd389ad6e8a38536d72689dd5010b11a"]]
}'

curl --request POST \
  --url https://api.devnet.zilliqa.com/ \
  --header 'Content-Type: application/json' \
  --data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "GetSmartContractSubState",
	"params": ["a2e4657de8108dd3730eb51f05a1d486d77be2df","stake_ssn_per_cycle", []]
}'


```

chetan-zilliqa: Above is the example is of staking contract  - https://viewblock.io/zilliqa/address/zil15lr86jwg937urdeayvtypvhy6pnp6d7p8n5z09?tab=state

chetan-zilliqa: @maxconway @JamesHinshelwood

maxconway: So I've tried all those, they all behave as I'd expect I think, what's your conclusion, @chetan-zilliqa? I've had another look at the trie code and I'm guessing what you're getting at is using iter_by_prefix rather than get, James? I'm not sure how constructing the prefix will interact with de-jsonifying the indices, but I'd imagine I'll be able to work it out.

maxconway: Right, @JamesHinshelwood the prefix handling capability is added

JamesHinshelwood (approved):
LGTM :)

PR #1794, opened 2024-11-12
Updated GetTransactionStatus API endpoint to use an enum

maxconway:
Updated GetTransactionStatus API endpoint to use an enum to represent possible transaction statuses

Added serde_repr to serialise to numeric representation

### DIFF
--- a/zilliqa/src/api/types/zil.rs
+++ b/zilliqa/src/api/types/zil.rs
@@ -7,6 +7,7 @@ use alloy::{
 use anyhow::Result;
 use k256::elliptic_curve::sec1::ToEncodedPoint;
 use serde::{Deserialize, Serialize};
+use serde_repr::{Deserialize_repr, Serialize_repr};
 
 use super::{hex, hex_no_prefix, option_hex_no_prefix};
 use crate::{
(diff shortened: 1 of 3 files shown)


zilliqa/src/api/types/zil.rs
```diff
@@ -619,6 +620,42 @@ pub struct TransactionReceiptResponse {
     pub success: bool,
 }
 
+// From https://github.com/Zilliqa/Zilliqa/blob/master/src/common/TxnStatus.h#L23
+#[derive(Serialize_repr, Deserialize_repr, Clone)]
+#[repr(u8)] // Because otherwise it's weird that 255 is a special case
+pub enum TxnStatusEnum {
```
JamesHinshelwood, line 626: Nittiest of nit picks - We know its an `enum` already so just call it `TxnStatus` :)

JamesHinshelwood (approved):

zilliqa/src/api/types/zil.rs
maxconway: I thought about that, but I wanted to make it clear that it's not the whole transaction status - the return type of get_transaction_status isn't just that enum, it's the whole of TransactionStatusResponse, with various other fields too. I agree that enum isn't the nicest way of putting it though, so I'm just about to push a commit changing it to TxnStatusCode, if that works for you?

JamesHinshelwood (approved):

PR #1795, opened 2024-11-12
Resend NewView messages after consensus_timeout seconds

86667:
In the case of a long timeout period it is possible for all nodes to be within the same view but the leader to not have all NewView message. Eg if a node timed out recently. 

Also when a node jumps to a view after starting up again we want a `NewView` message to be sent so that consensus can continue immediately if all other nodes are on the same view. 

We send the `NewView` message here after every `consensus_timeout` seconds. Unfortunately sending the message immediately on startup fails with `DialError` which i assume is due to the peer handshakes not yet having completed.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -253,7 +253,7 @@ impl Consensus {
                         .ok_or_else(|| anyhow!("missing finalized block!"))?;
 
                     // If latest view was written to disk then always start from there. Otherwise start from (highest out of high block and finalised block) + 1
-                    let mut start_view = db
+                    let start_view = db
                         .get_view()?
                         .or_else(|| {
                             Some(std::cmp::max(high_block.view(), finalized_block.view()) + 1)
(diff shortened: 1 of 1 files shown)


zilliqa/src/consensus.rs
```diff
@@ -364,6 +364,7 @@ impl Consensus {
             }
             // treat genesis as finalized
             consensus.set_finalized_view(latest_block_view)?;
+            return Ok(consensus);
```
86667, line 367: Return here if we are starting from genesis - the code below does not apply in this case

zilliqa/src/consensus.rs
```diff
@@ -111,6 +111,13 @@ type EarlyProposal = (
     EthTrie<MemoryDB>,
 );
 
+#[derive(Debug)]
+struct NewViewMessageStore {
+    peer_id: PeerId,
+    new_view: NewView,
+    resend_count: u8,
```
JamesHinshelwood, line 118: What is this count used for? Do you care that it will overflow quite quickly (5 seconds * 255 = ~21 minutes)?

zilliqa/src/consensus.rs
```diff
@@ -175,6 +182,8 @@ pub struct Consensus {
     pub new_receipts: broadcast::Sender<(TransactionReceipt, usize)>,
     pub new_transactions: broadcast::Sender<VerifiedTransaction>,
     pub new_transaction_hashes: broadcast::Sender<Hash>,
+    /// Store of most recent NewView message if one exists
+    new_view_message: Option<NewViewMessageStore>,
```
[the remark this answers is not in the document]
86667: Straight after startup the time is ~120ms which is why i added, although just checked building it later on and its more like ~14ms. To do with the evm call for leader i suspect. I'll remove this store then and rebuild each time.

zilliqa/src/consensus.rs
```diff
@@ -379,9 +380,51 @@ impl Consensus {
...
+                consensus.config.consensus.consensus_timeout,
+            );
+            let start_view_jumped_ahead = start_view + view_diff;
+            consensus.db.set_view(start_view_jumped_ahead)?;
+            debug!(
+                "Atleast {} views changed since last view was written. new start_view {}",
+                view_diff, start_view_jumped_ahead
+            );
+
+            // Build a NewView message ready for sending for the view we jump into. If other nodes are still on this view then this will allow for consensus to resume immediately.
```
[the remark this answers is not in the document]
86667: Fair. I've removed the NewView store.

We now re-send `NewView` messages at each `consensus_timeout`. This neatly allows us to avoid implementing functionality for sending a message after a delay since restarted nodes will send their NewView message via this re-send mechanism :)

zilliqa/src/consensus.rs
86667: doh, should have been u16. - removed now anyway.

JamesHinshelwood (approved):
LGTM, thanks!

PR #1813, opened 2024-11-13
Add cache for state trie keys

JamesHinshelwood:
This adds a LRU cache to the state trie, which keeps recently accessed keys in memory. The size of the cache is limited by the total number of bytes stored. The cache size is configurable and defaults to 256 MiB.

This commit also removes the stakers cache in `Consensus` since I claim it achieves 90% of the performance speed-up of that cache, without the added state and complexity in consensus.


zilliqa/src/consensus.rs
```diff
@@ -162,9 +162,6 @@ pub struct Consensus {
     /// Receipts cache
     receipts_cache: HashMap<Hash, (TransactionReceipt, Vec<Address>)>,
     receipts_cache_hash: Hash,
-    /// Cache stake/stakers
-    stake_cache: BTreeMap<u64, u128>,
-    stakers_cache: BTreeMap<(u64, Hash), Vec<NodePublicKey>>,
```
JamesHinshelwood, line 167: Unresolved comment to ensure @shawn-zil has a chance to test my assertion that this change doesn't hurt performance before we merge it.

zilliqa/src/consensus.rs
```diff
@@ -2814,32 +2768,18 @@ impl Consensus {
...
-    /// Retrieves total stake from Deposit contract via EVM.
-    fn total_weight_raw(&self, executed_block: BlockHeader) -> u128 {
-        self.state
-            .get_total_stake(executed_block)
-            .unwrap()
-            .unwrap()
-            .into()
+    fn total_weight(&self, committee: &[NodePublicKey], executed_block: BlockHeader) -> u128 {
+        committee
+            .iter()
```
shawn-zil, line 2773: Just sticking this comment here as a note.
This would be O(n), though the state caching mechanism now makes this cheap.

zilliqa/src/consensus.rs
shawn-zil: The initial numbers are good:
- Original 25 validators: 2.47s
- PR 1792 (Shawn) - 0.51s
- PR 1813 (James) - 0.34s

So, I've assigned this PR to close the issue. Thanks!

shawn-zil (approved):

PR #1814, opened 2024-11-13
API response comparator tool

chetan-zilliqa:
Publishing the tool to GitHub for easy access by others who may wish to use it.


maxconway:
This all looks good. If it was possible for this to be able to bring up and compare the local version of ZQ2 that would be particularly useful, but I'm guessing that's too complicated?

JamesHinshelwood (approved):

PR #1816, opened 2024-11-13
Present proper nonce at ZIL API level

bzawisto:
`nonce()` modifies the raw nonce for ZIL transactions whereas the same nonce should be returned in the API call as sent in create transaction call.

### DIFF
--- a/evm_scilla_js_tests/test/ContractRevert.ts
+++ b/evm_scilla_js_tests/test/ContractRevert.ts
@@ -22,11 +22,11 @@ describe("Revert Contract Call", function () {
   });
 
   it("Should revert transaction with a custom message if the called function reverts with custom message", async function () {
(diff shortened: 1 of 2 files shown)


zilliqa/src/api/types/zil.rs
```diff
@@ -195,13 +195,13 @@ impl GetTxResponse {
         receipt: TransactionReceipt,
         block_number: u64,
     ) -> Result<GetTxResponse> {
-        let nonce = tx.tx.nonce().unwrap_or_default();
```
JamesHinshelwood, line 198: Do we need to change `TransactionStatusResponse` in this file too?

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #1818, opened 2024-11-13
Improvements in `deposit.sol`

DrZoltanFazekas:
The PR shall implement two urgent improvements:
- `getFutureStake()` function to retrieve the staker's latest balance even if not yet effective i.e. future balance (required for the correct price calculation in liquid staking)
- unbondig period set in the constructor to the value of `nodes.consensus.unbonding_period` configured in the toml file (to enable shorter testing cycles of stake withdrawals on the devnet)


JamesHinshelwood (dismissed):
LGTM - Did you also want to set the unbonding period in the constructor?

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):

PR #1822, opened 2024-11-14
Return proper error response for non-existing account queries via ZIL api

bzawisto:


zilliqa/src/api/zil.rs
```diff
@@ -574,7 +575,18 @@ fn get_smart_contract_init(params: Params, node: &Arc<Mutex<Node>>) -> Result<Ve
...
         .get_block(BlockId::latest())?
         .ok_or_else(|| anyhow!("Unable to get the latest block!"))?;
-    let account = node.get_state(&block)?.get_account(address)?;
+
+    let state = node.get_state(&block)?;
+
+    if !state.has_account(address)? {
+        return Err(ErrorObject::owned(
+            RPCErrorCode::RpcInvalidAddressOrKey as i32,
+            "Address does not exist".to_string(),
```
JamesHinshelwood, line 584: We return the missing `address` in the error for `GetSmartContractCode` but not for the other two APIs - Is that intended?

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):

Issue #1590, opened 2024-10-08
Missing APIs in ZQ2 


chetan-zilliqa:
There are some APIs missing in ZQ2 that may or may not be necessary. Listing them below:

GetShardingStructure
GetNumTransactions
GetPrevDifficulty
GetTotalCoinSupply and GetTotalCoinSupplyAsInt – These are used by external parties to retrieve the total coin supply.
GetTxBlockRate
GetNumTxnsDSEpoch
DSBlockListing – This API returns the error: "Invalid params", "data": "invalid type: integer 1, expected a string at line 1 column 2".
GetDsBlock

cc @DrZoltanFazekas @bzawisto @JamesHinshelwood

JamesHinshelwood:
cc @maxconway  - I believe some of these are already tracked by your issues?

maxconway:
GetDSblockRate and GetTxBlockRate are already merged and tested - is there somewhere else I should be noting them down?

GetTotalCoinSupply, GetTotalCoinSupplyAsInt, GetNumTxnsDSEpoch I've already done outline implementations for and got tests together, they're in my todo list to double check and merge in

Re DSBlockListing, yep, that's my bad, looks like I write it to accept a string when it should be taking an int. I've created an issue for this: #1593 

Regarding the others: GetShardingStructure, GetNumTransactions, GetPrevDifficulty, there's quite a few others in the same state as well, so I'm very open to help working out which ones to do first. I'll do a little table in the next comment to list every API I know about

chetan-zilliqa:
Please proceed as you see fit. My intention was to test the persistence converter tool and call the API to verify responses for legacy data. In the process, I created the tickets, but I wasn't aware that you were already tracking them. There's no priority from my side.:-) Thanks for the detailed tracking.

PR #1826, opened 2024-11-14
Implement GetNodeType

### LINKED ISSUE #1538: GetNodeType
### TODO list

- [x] Write tests
- [ ] Draft implementation
- [ ] Finalize and merge

maxconway:
- Previous node types (Seed/Lookup/Not in network...) don't appear to be relevant in ZQ2, so this now returns Leader or Validator instead
- Docs updated to match
- Test added
- Closes #1538
- References #1590

### DIFF
--- a/docs/api/zilliqa/getnodetype.doc.md
+++ b/docs/api/zilliqa/getnodetype.doc.md
@@ -14,9 +14,8 @@ NotDocumented
 
 Returns node type. The possible return values are:
 
-- `"Not in network, synced till epoch [epoch number]"` if the server has not joined the network and is synced until a specific epoch.
-- `"Seed"` if the server is in lookup node mode and is an archival lookup node.
-- `"Lookup"` if the server is in lookup node mode
+- `"Leader"` if we are currently the lead node
+- `"Validator"` otherwise
 
 # Curl
 
@@ -32,7 +31,7 @@ curl -d '{
 # Response
 
 ```json
-{ "id": "1", "jsonrpc": "2.0", "result": "Seed" }
+{ "id": "1", "jsonrpc": "2.0", "result": "Validator" }
 ```
 
 # Arguments
@@ -43,4 +42,3 @@ curl -d '{
 | `jsonrpc` | string | Required | `"2.0"`           |
 | `method`  | string | Required | `"GetNodeType"`   |
 | `params`  | string | Required | Empty string `""` |
-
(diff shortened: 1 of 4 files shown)


docs/api/zilliqa/getnodetype.doc.md
```diff
@@ -14,9 +14,8 @@ NotDocumented
 
 Returns node type. The possible return values are:
 
-- `"Not in network, synced till epoch [epoch number]"` if the server has not joined the network and is synced until a specific epoch.
-- `"Seed"` if the server is in lookup node mode and is an archival lookup node.
-- `"Lookup"` if the server is in lookup node mode
+- `"Leader"` if we are currently the lead node
+- `"Validator"` otherwise
```
JamesHinshelwood, line 18: I think we should return (a subset of) the same values that ZQ1 returned. Clients will be expecting to receive one of the existing values, so I'm cautious about inventing new ones. @chetan-zilliqa Can you advise which ZQ1 value is most appropriate for ZQ2 nodes?

maxconway: If I try it on api.zilliqa.com I get:

````
curl -s --request POST \
  --url https://api.zilliqa.com/ \
  --header 'Content-Type: application/json' \
  --data '{
    "id": "1",
    "jsonrpc": "2.0",
    "method": "GetNodeType",
        "params": [""]
}' | jq
{
  "id": "1",
  "jsonrpc": "2.0",
  "result": "Seed"
}
````

So should I just hardcode it to always return Seed then? And update the documentation to say that it always returns seed for backwards compatibility reasons?

chetan-zilliqa: Only two values would make sense here IMO.
seed - If a node is an API node.
validator - For all other nodes.

maxconway: So it should be okay to hardcode Seed then, right? because GetNodeType doesn't take arguments, it always applies to the node answering the request, and that's guaranteed to be an API node precisely because it is answering the request?

maxconway: This is now updated to always return Seed

JamesHinshelwood (approved):

docs/api/zilliqa/getnodetype.doc.md
JamesHinshelwood: Makes sense to me :+1: Thanks

PR #1830, opened 2024-11-14
Remove MDNS

rrw-zilliqa:
(this PR applies on top of https://github.com/Zilliqa/zq2/pull/1829 )

GCP only partly supports mdns and local networks don't support it at all; having it present encourages libp2p to think it can locate peers using mdns and do work that it doesn't need to do. It also leads to an odd situation with local networks (where nodes have no globally visible address) where all packets have to be passed using illegitimate mdns broadcasts because we refuse to accept any of the addresses identify identifies (they are all classed as non-global and our code discards them).

This patch removes mdns, simplifying our libp2p config. zq2-richard is at block 106,000 and some with it, so I hope it's not too deleterious ..


rrw-zilliqa:
Duplicate of JH's recent PR

PR #1831, opened 2024-11-15
Fix contract state api returning parsing error

chetan-zilliqa:


maxconway:
This is going to conflict quite heavily with #1793 I think. #1793 splits GetSmartContractSubstate and GetSmartContractState up again, removing the get_smart_contract_state_internal function because the implementations changed quite a bit.

Any thoughts on which we want to merge first?

chetan-zilliqa:
maybe I will remove the `get_smart_contract_state_internal` changes from the PR which are conflicting. Then we can modify later on top of your change.

maxconway (approved):
I'm happy with this, but someone else should probably take a gander too because I was the last one to touch it. @bzawisto?

chetan-zilliqa:
Thanks @bzawisto for the review comment on the clone. I will merge it now as I have to create a build to convert the persistence. Sorry for not handling the review comment in the current PR.
Created a new issue for this -  - https://github.com/Zilliqa/zq2/issues/1837

z2/src/converter.rs
```diff
@@ -252,7 +252,6 @@ pub async fn convert_persistence(
     )?;
 
     if convert_accounts {
-        // Calculate an estimate for the number of accounts by taking the first 100 accounts, calculating the distance
```
JamesHinshelwood, line 255: Please don't delete this :)

maxconway (approved):

its-saeed (approved):

saeed-zil (approved):

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #1832, opened 2024-11-15
Add support for scilla transtiions to converter

chetan-zilliqa:
There are some Println statements that I will remove afterwards. Currently, they are kept for debugging purposes.


z2/Cargo.toml
```diff
@@ -59,6 +59,7 @@ revm = {version = "14.0.3", features = ["optional_balance_check"]}
 rs-leveldb = "0.1.5"
 rustls = "0.23.16"
 serde = {version = "1.0.215", features = ["derive"]}
+scilla-parser = "2.0.0"
```
JamesHinshelwood, line 62: Please bump the version of `scilla-parser` in `zilliqa/Cargo.toml` too so we don't have two versions lying around.

z2/src/converter.rs
```diff
@@ -206,13 +208,55 @@ fn get_contract_code(zq1_db: &zq1::Db, address: Address) -> Result<Code> {
...
+            println!("Transitions:");
+            for transition in transitions {
+                println!("  Transition Name: {}", transition.name);
+                for param in &transition.params {
+                    println!("    Param Name: {}, Type: {}", param.name, param.ty);
+                }
+            }
+        }
+        _ => {}
+    }
```
JamesHinshelwood: Is this some debugging that was left behind?

z2/src/converter.rs
```diff
@@ -206,13 +208,55 @@ fn get_contract_code(zq1_db: &zq1::Db, address: Address) -> Result<Code> {
...
+    let contract = Contract::from_str(&code)?;
+
+    let scilla_code = Code::Scilla {
+        code: code.clone(),
         init_data: init_data_vec,
         types: BTreeMap::default(),
-        transitions: vec![],
-    })
+        transitions: contract
+            .transitions
```
JamesHinshelwood, line 221: Elsewhere in ZQ2 we rely on the Scilla interpreter's check output to calculate the `transitions` for a contract. Whereas here we are using the Rust parser. While these parsers should be the same in theory, it worries me a bit that they will differ for some edge cases. I would prefer if we used the Scilla interpreter here as well if possible.

Or if that's not possible, can we add a test suite to the Rust Scilla parser which takes every contract from testnet and mainnet and asserts that the parsed result is the same as what the OCaml implementation returns?

z2/src/converter.rs
chetan-zilliqa: I kept it intentionally because I am planning to test persistence conversion with all these PR combined. Currently, it is hard to test and verify if we have more than one branch. The print statement, i will remove later on.

z2/Cargo.toml
chetan-zilliqa: thanks , done

JamesHinshelwood (approved):
LGTM, thanks for switching the parser :)

JamesHinshelwood (approved):

PR #1835, opened 2024-11-15
BREAKING CHANGE: Calculate view jump based on high_qc updated_at time

86667:
These changes fix the linked bug in which restarting a node multiple times whilst consensus is not met causes the view to jump forward too far. Here, we determine the view jump number off of the `high_qc` which will not have changed since the original node shut down.

Unfortunately this is another breaking change to the db schema.


zilliqa/src/db.rs
```diff
@@ -259,6 +259,7 @@ impl Db {
                 view INTEGER,
                 view_updated_at BLOB,
```
JamesHinshelwood, line 260: Do we still use the `view_updated_at`?

JamesHinshelwood (dismissed):

zilliqa/src/db.rs
86667: Removed

infra/config_rpc_disabled.toml
```diff
@@ -1,19 +1,32 @@
 otlp_collector_endpoint = "http://otel-collector:4317"
 
 [[nodes]]
+# These (public key, peerId, stake, rewardAddress) tuples correspond to the private keys of all four nodes in `docker-compose.yaml`.
```
86667, line 4: bringing this config inline with `./config.toml`

JamesHinshelwood (approved):

PR #1847, opened 2024-11-18
Implement GetSoftConfirmedTransaction

### LINKED ISSUE #1546: GetSoftConfirmedTransaction
### TODO list

- [x] Write tests
- [x] Draft implementation
- [ ] Finalize and merge

maxconway:
- "Soft" isn't mentioned elsewhere in the ZQ2 codebase so this just wraps GetTransaction
- Adds a new test, and lengthenes the test for GetTransaction, but keeps them separate in case we need to modify one or both in future
- Looks like there might be a bug with stringification of addresses somewhere leading to mixed case results, but it feels outside the scope of this commit so I normalise via to_lowercase in the test
- Closes #1546

### DIFF
--- a/zilliqa/tests/it/zil.rs
+++ b/zilliqa/tests/it/zil.rs
@@ -667,6 +667,14 @@ async fn get_transaction(mut network: Network) {
     );
     assert_eq!(parsed_response.amount.to_string(), "200000000000000");
     assert_eq!(parsed_response.gas_limit.0, 50000);
+
+    let response_soft_confirmed: Value = wallet
+        .provider()
+        .request("GetSoftConfirmedTransaction", [transaction_id])
+        .await
+        .expect("Failed to call GetTransaction API");
+
+    assert_eq!(response, response_soft_confirmed);
 }
 
 // We need to restrict the concurrency level of this test, because each node in the network will spawn a TCP listener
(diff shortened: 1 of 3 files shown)


zilliqa/tests/it/zil.rs
```diff
@@ -2257,9 +2269,78 @@ async fn get_smart_contract_sub_state(mut network: Network) {
...
+        .as_str()
+        .expect("Failed to get ID from response");
+
+    // Wait for the transaction to be mined
+    network.run_until_block(&wallet, 1.into(), 50).await;
+
+    // Use the GetTransaction API to retrieve the transaction details
+    let response: Value = wallet
+        .provider()
+        .request("GetTransaction", [transaction_id])
```
JamesHinshelwood, line 2307: Isn't this the wrong API? Thought FWIW I'd probably prefer just adding an extra request to the existing `GetTransaction` test which asserts that the response from `GetTransaction` and `GetSoftConfirmedTransaction` are the same.

JamesHinshelwood (approved):

zilliqa/tests/it/zil.rs
maxconway: Right, I've combined them, and good spot on the copy paste error

JamesHinshelwood (approved):

PR #1851, opened 2024-11-19
Remove mDNS

JamesHinshelwood:
It seems to interact weirdly with autonat in real networks.


shawn-zil:
Just a note that this seems to have completely broken local docker.

PR #1854, opened 2024-11-20
Add mDNS again

JamesHinshelwood:
Reverts Zilliqa/zq2#1851

Fixes local docker runs until @JamesHinshelwood sends a PR that removes mDNS without breaking local development. Sorry!


rrw-zilliqa:
Ah! Sorry - I hoped that wouldn't happen - add 172.<whatever> to the local network list and that should fix it?

PR #1867, opened 2024-11-20
Events requested by Blockscout

DrZoltanFazekas:
The same events can also be used by the UCCB validator oracle to keep track of changes of the committee and the validators' stake. The events in this PR replace the `StakerAdded` and `StakerRemoved` events implemented in https://github.com/Zilliqa/zq2/pull/1178


zilliqa/src/contracts/deposit.sol
```diff
@@ -421,6 +435,13 @@ contract Deposit {
         }
     }
 
+    // Returns the next block number at which new stakers are added,
+    // existing ones removed and/or deposits of existing stakers change
+    function nextUpdate() public view returns (uint256 blockNumber) {
+        if (latestComputedEpoch > currentEpoch())
+            blockNumber = latestComputedEpoch * blocksPerEpoch;
```
JamesHinshelwood, line 442: Why is this not `blockNumber = (currentEpoch() + 2) * blocksPerEpoch`? That's the logic we use in `deposit` and `unstake` so shouldn't we be consistent?

JamesHinshelwood (approved):

zilliqa/src/contracts/deposit.sol
DrZoltanFazekas: Not really imo. If there are future changes then `latestComputedEpoch` determines when they become effective. This is either at the beginning of `currentEpoch() + 2` or if this function is called later, maybe already at the beginning of `currentEpoch() + 1`.

DrZoltanFazekas: Note that the function is `public`, i.e. it's not only intended for internal use when we emit the events.

JamesHinshelwood: Oh yes, I misunderstood the point of this function. Fair enough :)

JamesHinshelwood (approved):

PR #1875, opened 2024-11-22
SQLite optimisations

### LINKED ISSUE #1503: Change how we store state to reduce state growth
Even in absence of transactions the persisted state is growing by approx. 250-300 MB/day. This is because rewards are issued to validators on each block and we produce 86,400 bocks per day and we store state trie updates. We shall consider a more efficient way of storing the state and only construct the trie on the fly to compute its root. Please add your thought @JamesHinshelwood

### LINKED ISSUE #1748: Block intervals increase over time, even with as few as 10 validators
In the `devnet` with only 8 validators producing empty blocks, the average block time was 1.2s at the beginning, but by the time the network reached 30k blocks, it increased noticeably:
```
  view height timestmp                                   proposer      gas txns highqc
--------------------------------------------------------------------------------------
 29843  29362 23:32:53 0xb2d01a549f84d04346f4bb454ef6ae2553129526        0    0  29842 
 29844  29363 23:32:55 0x977a797a0546cdb1082c2368921af455dfdc7a5a        0    0  29843 2s since last block
 29845  29364 23:32:58 0xa598a7946e88196f9aa6c05837a554d2318e4c12        0    0  29844 3s since last block
 29846  29365 23:33:00 0x977a797a0546cdb1082c2368921af455dfdc7a5a        0    0  29845 2s since last block
 29847  29366 23:33:03 0x35293af1338a53ee22ef3a784e6f42e4d05702b5        0    0  29846 3s since last block
 29848  29367 23:33:05 0x5e579ddd88a40c0738453d284c25adc0a5178cbc        0    0  29847 2s since last block
 29849  29368 23:33:06 0xa598a7946e88196f9aa6c05837a554d2318e4c12        0    0  29848 
 29850  29369 23:33:08 0x977a797a0546cdb1082c2368921af455dfdc7a5a        0    0  29849 2s since last block
 29851  29370 23:33:12 0x35293af1338a53ee22ef3a784e6f42e4d05702b5        0    0  29850 4s since last block
 29852  29371 23:33:14 0x977a797a0546cdb1082c2368921af455dfdc7a5a        0    0  29851 2s since last block
 ```
 
For comparison, the `prototestnet` with 11 validators has produced over 3.5 million blocks yet maintained an average block time of 1.2s

Since individual block times are 5s or even longer, the `devnet` has a lot of missed views (timeouts).

It's unclear whether the block times are increasing over time i.e. with the number of blocks produced by the network.

### PR DESCRIPTION
**Pull request by shawn-zil**
This PR does a few things:

First, it uses `WITHOUT ROWID` when creating tables. Only works on **new** databases. Measurements show about 2% filesize impact for each table. Digging into the [docs](https://www.sqlite.org/withoutrowid.html):
> The WITHOUT ROWID optimization is likely to be helpful for tables that have non-integer or composite (multi-column) PRIMARY KEYs and that do not store large strings or BLOBs.

Second, it turns on several SQLite options:
- `journal_mode=wal` generally improves overall read/write operations by using an on-disk journal.
- `journal_size_limit=32M` reduces journal file growth; 32M chosen based on observations.
- `synchronous=normal` reduces unnecessary filesystem I/O; small risk of data-loss in a crash.
- `temp_store=memory` stores temporary/intermediate indices in RAM; for complex queries.
- `page_size=32768` packs more data into a database page, reducing fragmentation; 64K had no extra impact; only impacts **new** databases; existing databases need to be vacuumed with the new page size.

Third, it implements transparent LZ4 de/compression on a number of `ToSql/FromSql` traits to further reduce the storage used by selected blobs. This is done in a backwards-compatible way, where non-compressed blocks will be passed-thru. LZ4 was chosen as it runs faster than ZSTD:
```
ZSTD 12.423262ms 1078
LZ4F 13.213884ms 1583
LZ4 304.681µs 1541
```

Net impact is > 20% reduction in DB size, producing empty blocks, across 1 day. I've left `perftest` running over the weekend to collect data.

- Closing #1503 for now.
- Closing #1748 for now.

### DIFF
--- a/zilliqa/src/db.rs
+++ b/zilliqa/src/db.rs
@@ -205,6 +205,12 @@ impl Db {
             None => (Connection::open_in_memory()?, None),
         };
 
+        // SQLite performance tweaks
+        connection.pragma_update(None, "journal_mode", "wal")?;
+        connection.pragma_update(None, "synchronous", "normal")?;
+        connection.pragma_update(None, "temp_store", "memory")?;
+        connection.pragma_update(None, "journal_size_limit", 1 << 22)?;
+
         connection.trace(Some(|statement| tracing::trace!(statement, "sql executed")));
 
         Self::ensure_schema(&connection)?;
--- a/zilliqa/src/db.rs
+++ b/zilliqa/src/db.rs
@@ -52,7 +52,7 @@ macro_rules! sqlify_with_bincode {
     };
 }
 
-macro_rules! sqlify_with_bincode_and_zstd {
+macro_rules! sqlify_with_bincode_and_lz4 {
     ($type: ty) => {
         impl ToSql for $type {
             fn to_sql(&self) -> rusqlite::Result<ToSqlOutput<'_>> {
@@ -98,10 +98,13 @@ macro_rules! make_wrapper {
     };
 }
 
-sqlify_with_bincode!(AggregateQc);
-sqlify_with_bincode!(QuorumCertificate);
-sqlify_with_bincode!(NodeSignature);
-sqlify_with_bincode_and_zstd!(SignedTransaction);
+// Careful about choosing which blobs to LZ4
+// e.g. it may impact search on a column that is LZ4-ed.
+
+sqlify_with_bincode_and_lz4!(AggregateQc);
+sqlify_with_bincode_and_lz4!(QuorumCertificate);
+sqlify_with_bincode_and_lz4!(NodeSignature);
+sqlify_with_bincode_and_lz4!(SignedTransaction);
 
 make_wrapper!(Vec<ScillaException>, VecScillaExceptionSqlable);
 sqlify_with_bincode!(VecScillaExceptionSqlable);
--- a/zilliqa/src/db.rs
+++ b/zilliqa/src/db.rs
@@ -56,21 +56,20 @@ macro_rules! sqlify_with_bincode_and_zstd {
     ($type: ty) => {
         impl ToSql for $type {
             fn to_sql(&self) -> rusqlite::Result<ToSqlOutput<'_>> {
-                let v = bincode::serialize(self)
+                let data = bincode::serialize(self)
                     .map_err(|e| rusqlite::Error::ToSqlConversionFailure(e))?;
-                let v = lz4::block::compress(v.as_slice(), None, true)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: zilliqa/src/db.rs [hunk 1]
```diff
@@ -231,11 +266,11 @@ impl Db {
                 gas_limit INTEGER NOT NULL,
                 qc BLOB NOT NULL,
                 agg BLOB,
-                is_canonical BOOLEAN NOT NULL);
+                is_canonical BOOLEAN NOT NULL) WITHOUT ROWID;
```
JamesHinshelwood, line 269: What's the downside of `WITHOUT ROWID`?

zilliqa/src/db.rs
```diff
@@ -32,21 +32,23 @@ use crate::{
     transaction::{EvmGas, Log, SignedTransaction, TransactionReceipt},
 };
 
-macro_rules! sqlify_with_bincode {
+macro_rules! sqlify_with_bincode_and_lz4 {
```
JamesHinshelwood, line 35: How much does compressing these values save? Can we measure it and see if it non-negligible?

My worry is that if compressing does shrink the value size a lot, then we should actually be changing how the values are encoded, since there is clearly some waste. And if compressing doesn't change the size, then we shouldn't slow down DB reads and writes by running it.

Or to state my thoughts in another way - Compression isn't 'magic' and it can only remove 'redundant' data. So why can't we just remove that redundant data (if it exists) from the encoded values ourselves?

JamesHinshelwood (commented):
LGTM, but I would like to explore the trade-offs around value compression more before merging. Feel free to schedule a call if I'm not making any sense :)

zilliqa/src/db.rs
shawn-zil: From the docs, as long as the Primary Key is not too big, then there is not much of a downside. The recommendation is to keep it under 1/20th of a page size. In our case, the page size is configured as 32KB (4KB default), which would be 1.6KB (200 bytes default). In either case, our Primary Key is under that size, iirc.

zilliqa/src/db.rs
shawn-zil: For empty blocks, the compression brings about nett ~1.5% size reduction overall. If we don't want to go overboard with compression, then we can remove them. In that case, we will get ~15% reduction from the WITHOUT ROWID and pragmas.

PR #1879, opened 2024-11-22
Deposit contract POP to sign over from address and chain_id

86667:
Sign over `bls_public_key`, `from` and `chain_id` when proving possession of bls pub key.

We sign over both public key and address to allow for the sender of the deposit transaction to be a different account to the key which is staking. 

This new deposit authentication mechanism will be used in `deposit_v3`, therefore giving us an opportunity to build and test a basic contract upgrade mechanism which are included in this PR. 

Tools for testing `deposit_v3` via z2 commands can be found in [this branch](https://github.com/Zilliqa/zq2/tree/deposit_v3_changes). I'll keep them on this branch until we deploy deposit_v3 in the coming weeks.


shawn-zil (commented):
My opinion is that we're confusing 2 different issues - possession of the key, and whether it is applicable to a particular chain. When Theo brought this up, I felt that it should be handled separately. 

Proof of possession of a key is about authentication and which chain it is on is an authorisation problem. It didn't sit well with me that we're trying to do a 2-in-1. I think that a alternate solution might be to do a PoP, with a separate challenge/response/check for the authorisation bits.

However, I don't feel strongly enough about it to block this PR.

86667:
> My opinion is that we're confusing 2 different issues - possession of the key, and whether it is applicable to a particular chain. When Theo brought this up, I felt that it should be handled separately.
> 
> Proof of possession of a key is about authentication and which chain it is on is an authorisation problem. It didn't sit well with me that we're trying to do a 2-in-1. I think that a alternate solution might be to do a PoP, with a separate challenge/response/check for the authorisation bits.
> 
> However, I don't feel strongly enough about it to block this PR.

I think youre correct in that the PoP is authentication only and so perhaps using its name here is incorrect/misleading. 

Im not sure I agree however that we need a PoP as well as a separate signature over the authentication data, we are after all proving that we know the corresponding private key in the authentication proof.  We would be proving knowledge of our private key twice. Unless you have some other mechanism for the authentication proof in mind?

@DrZoltanFazekas how do you feel about this non-explicit PoP which includes authentication data? Am i correct in thinking that the PoP is used only here and is not required later in the aggregate signing protocol, and so therefore can safely be removed?

I would be happy to rename to something general such as `bls_deposit_signature` signing over the public key, from address of the tx and chain id.

DrZoltanFazekas:
> I think youre correct in that the PoP is authentication only and so perhaps using its name here is incorrect/misleading.
> 
> Im not sure I agree however that we need a PoP as well as a separate signature over the authentication data, we are after all proving that we know the corresponding private key in the authentication proof. We would be proving knowledge of our private key twice. Unless you have some other mechanism for the authentication proof in mind?
> 
> @DrZoltanFazekas how do you feel about this non-explicit PoP which includes authentication data? Am i correct in thinking that the PoP is used only here and is not required later in the aggregate signing protocol, and so therefore can safely be removed?
> 
> I would be happy to rename to something general such as `bls_deposit_signature` signing over the public key, from address of the tx and chain id.

The PoP is only used in the `deposit()` function to register the validators' BLS pubkeys. What do you mean by "can safely be removes" @86667 ?

To prevent that the PoP can be misused by others to impersonate our BLS pubkey and bind it to their control address (which is set to `msg.sender` in the `deposit()` function), we have to bind the PoP to the legitimate `msg.sender`, the one who is allowed to use the PoP in a deposit transaction. We can rename it if you want since it's not the standard BLS PoP anymore.

IIRC the shards will have a unique chain id and you can't manipulate the chain id since it's part of the signed transaction (EIP-155). It's true that we're using the same chain id on internal test networks (devnet, perftest etc.) so our team mates could drain our test account's balance by replaying the deposit transaction they saw on another internal test network, but only if the nonce on the network they're replaying the transaction on is lower than on the original network the transaction was submitted to. Am I missing something?

zilliqa/src/precompiles/pop_verify.rs
```diff
@@ -13,6 +13,7 @@ use crate::exec::PendingState;
 
 pub struct PopVerify;
 
+/// A custom Proof of Possession in which the bls public key, an evm address and a chain_id are signed over
 // keep in-sync with zilliqa/src/contracts/deposit.sol
 impl PopVerify {
     const POP_VERIFY_GAS_PRICE: u64 = 1_000_000u64; // FIXME: Gas Price?
```
shawn-zil, line 19: A nit. But since you're working on this, maybe you could add in the changes in #1335 which has been there for months.

86667:
@shawn-zil we decided to remove the Pop altogether and instead pass a standard signature signing over the required data. Zoltan is happy for this signature to inexplicably provide a proof of possession. 

I've included what I think is a fair gas cost for signature verification based off of gas costs in Ethereum.

zilliqa/src/contracts/deposit_v3.sol
```diff
@@ -0,0 +1,685 @@
...
+        for (uint256 i = 0; i < 20; i++) {
+            message[blsPubKey.length + 8 + i] = bytes20(msg.sender)[i];
+        }
+
+        // Verify bls signature
+        if (!_blsVerify(message, blsPubKey, signature)) {
+            revert RogueKeyCheckFailed();
+        }
+
+        Staker storage staker = $._stakersMap[blsPubKey];
```
shawn-zil, line 496: nit: does it make sense to move this a few lines down, where `staker` is first used?

zilliqa/src/contracts/deposit_v3.sol
```diff
@@ -0,0 +1,685 @@
...
+        }
+        if (peerId.length != 38) {
+            revert UnexpectedArgumentLength("peer id", 38);
+        }
+        if (signature.length != 96) {
+            revert UnexpectedArgumentLength("signature", 96);
+        }
+        DepositStorage storage $ = _getDepositStorage();
+
+        uint64 chain_id = uint64(block.chainid);
```
shawn-zil, line 474: nit: lint/format warning.

shawn-zil (approved):
Would be good to test this out, particularly the upgrade.

zilliqa/src/contracts/deposit_v3.sol
```diff
@@ -0,0 +1,684 @@
...
+        }
+
+        for (uint256 i = 0; i < 8; i++) {
+            message[blsPubKey.length + i] = bytes1(
+                uint8(chainId >> (8 * (7 - i)))
+            );
+        }
+
+        for (uint256 i = 0; i < 20; i++) {
+            message[blsPubKey.length + 8 + i] = bytes20(msg.sender)[i];
```
JamesHinshelwood, line 488: Can we use `abi.encodePacked` for encoding these parameters to bytes instead?

86667: Good idea. https://github.com/Zilliqa/zq2/pull/2015

PR #1902, opened 2024-11-25
Update eth_getBlockByNumber and eth_getBlockByHash to match spec

maxconway:
- Removed view, mixhash, quorumCertificate, since they're not in Etherium spec
- Added comments for why placeholders are placeholders


zilliqa/src/api/types/eth.rs
```diff
@@ -134,9 +134,9 @@ pub struct Header {
     #[serde(serialize_with = "hex")]
     pub nonce: [u8; 8],
     #[serde(serialize_with = "hex")]
-    pub sha_3_uncles: B256,
+    pub sha_3_uncles: B256, // Uncles do not exist in ZQ2
     #[serde(serialize_with = "hex")]
-    pub logs_bloom: [u8; 256],
+    pub logs_bloom: [u8; 256], // Zilliqa blocks do not have logs of their own
```
JamesHinshelwood, line 139: This should have a value. I'm not 100% certain but I think this should be the bitwise-OR of the bloom filters of all the logs in the block.

maxconway: I've pushed a change to implement this

maxconway: Sorry for the wall of text, but I've looked at this and it's kind of tricky to do in a useful way. 

The aim of a bloom filter is to place it as a sort of cache in front of an expensive lookup to avoid doing that lookup (the majority of the time) when the key isn't in the index. 

However, in our case as far as I can see we're not keeping a bloom filter updated in the block, so my implementation was to query every log from every transaction to build it up for each hit of the endpoint. 

This was very slow, and it doesn't really make any sense because if we're already fetching everything there's no point having a bloom filter. So I could probably make it a bit faster, but there still doesn't seem to be much point. 

If we want to return this field usefully we should probably add it to the blocks themselves and keep it updated, but until then, I think the right thing is to return a placeholder of all 1s. This just indicates that any log may be in this block, so it should be backwards compatible with any caller, and while it's not as fast as a proper bloom filter, it'll be faster than what we have now. 

Let me know if this makes sense @JamesHinshelwood - I've pushed the all 1s filter approach

JamesHinshelwood (approved):

PR #1910, opened 2024-11-27
Send transaction in batches

bzawisto:
Transactions are not immediately broadcast after being received from API but added to a new vector instead. 
Mempool is regularly polled for anything in the batch and, if criteria is met (chosen arbitrarily), transactions are broadcast to other peers.

There's a new timer - the old one wasn't kicking out frequently (and if the traffic was stable - never).


zilliqa/src/pool.rs
```diff
@@ -289,14 +309,40 @@ impl TransactionPool {
...
+        if self.first_broadcast_txn_add_time == SystemTime::UNIX_EPOCH {
+            self.first_broadcast_txn_add_time = SystemTime::now();
+        }
+    }
+
+    pub fn pull_txns_to_broadcast(&mut self) -> Result<Vec<SignedTransaction>> {
+        const MAX_TIME_SINCE_FIRST_INSERT: Duration = Duration::milliseconds(50);
+        const MAX_BATCH_SIZE: usize = 20;
+
+        if self.transactions_to_broadcast.len() > MAX_BATCH_SIZE
```
shawn-zil, line 337: Instead of having a minimum batch size, and needing to check for 'staleness', I would probably have a maximum batch size limit (just in case there are too many) and pull whatever is available up to that max size, at the point to broadcast.

zilliqa/src/node_launcher.rs
```diff
@@ -290,7 +293,12 @@ impl NodeLauncher {
                 r = self.reset_timeout_receiver.next() => {
                     let sleep_time = r.expect("reset timeout stream should be infinite");
                     trace!(?sleep_time, "timeout reset");
-                    sleep.as_mut().reset(Instant::now() + sleep_time);
+                    consensus_sleep.as_mut().reset(Instant::now() + sleep_time);
+                },
+
+                () = &mut mempool_sleep => {
+                    self.node.lock().unwrap().process_transactions_to_broadcast()?;
+                    mempool_sleep.as_mut().reset(Instant::now() + Duration::from_millis(20));
```
shawn-zil, line 301: Is 20ms too regular? I don't know. Personally, I would prefer less regular but larger batches e.g. 40 batches of 100 txns rather than 100 batches of 40 txns. There is a cost to do the underlying gossip.

shawn-zil: Yea, my hunch is that it might work better. It may also have less routing overhead i.e. less big flows, rather than more small flows.

JamesHinshelwood:
This adds some complexity in the mempool and node that makes me worry a bit. Before we decide on merging this, please could we do some testing to prove that batching `NewTransaction`s provides a meaningful speed-up over sending messages individually?

zilliqa/src/pool.rs
```diff
@@ -289,14 +293,41 @@ impl TransactionPool {
...
+
         TxAddResult::AddedToMempool
     }
 
+    fn store_broadcast_txn(&mut self, txn: SignedTransaction) {
+        self.transactions_to_broadcast.push_back(txn);
+    }
+
+    pub fn pull_txns_to_broadcast(&mut self) -> Result<Vec<SignedTransaction>> {
+        const MAX_BATCH_SIZE: usize = 100;
```
shawn-zil, line 315: Might want to make this 400, which would be a theoretical max of 400/100ms (4K/s)

JamesHinshelwood:
Last I recall, we were testing the difference this made in a test network. Did we get any results?

JamesHinshelwood:
Will close this for now. We can re-open later if we see the need based on performance tests.

chetan-zilliqa (dismissed):

zilliqa/src/message.rs
```diff
@@ -263,6 +263,7 @@ pub enum ExternalMessage {
     BlockResponse(BlockResponse),
     ProcessProposal, // deprecated since 0.7.0
     NewTransaction(SignedTransaction),
+    BatchedTransactions(Vec<SignedTransaction>),
```
shawn-zil, line 266: I believe that this will break syncing, unless all the nodes are running the same version as the enum values for the subsequent messages will be different.

PR #1915, opened 2024-11-28
Use redb instead of SQLite

JamesHinshelwood:
The performance of redb in devnet compared to SQLite is much faster. I don't necessarily blame SQLite for this, since it could easily be our transaction management, SQL queries, lack of indices, etc. that are making the existing implementation slow. However, I do feel confident saying that this interface makes it clearer what work the database is doing and makes it harder to accidentally do slow things.


shawn-zil:
Can we run it for a week and measure the block-slowdown timings (if any)?

JamesHinshelwood:
Benchmark when rebased onto #1941 + #1945 
```
produce-full/produce-full
                        time:   [1.4581 s 1.4839 s 1.4965 s]
                        thrpt:  [0.6682  elem/s 0.6739  elem/s 0.6858  elem/s]
                 change:
                        time:   [-16.643% -14.095% -11.524%] (p = 0.00 < 0.05)
                        thrpt:  [+13.025% +16.407% +19.966%]
                        Performance has improved.
```

flamegraph

PR #1937, opened 2024-12-03
Fixed missing Genesis deposit balances.

shawn-zil:
This PR does two things:
- Fixes config.toml for local/docker, due to TCP changes.
- Sets the Deposit contract to the total deposits at genesis, which was missing.

Before this fix:
```
curl -X POST   -H "Content-Type: application/json"   --data '{"jsonrpc":"2.0","method":"eth_getBalance","params":["0x000000000000000000005a494C4445504F534954", "latest"],"id":1}' http://localhost:4201/
{"jsonrpc":"2.0","id":1,"result":"0x0"}
```

After this fix:
```
curl -X POST   -H "Content-Type: application/json"   --data '{"jsonrpc":"2.0","method":"eth_getBalance","params":["0x000000000000000000005a494C4445504F534954", "latest"],"id":1}' http://localhost:4201/
{"jsonrpc":"2.0","id":1,"result":"0x2116545850052128000000"}
```

The amount 0x2116545850052128000000 is 40000000000000000000000000ZIL, which corresponds to the total deposited amount in the genesis contract for local/docker.

### DIFF
--- a/zilliqa/src/state.rs
+++ b/zilliqa/src/state.rs
@@ -9,7 +9,6 @@ use alloy::{
     primitives::{Address, B256},
 };
 use anyhow::{anyhow, Result};
-use contract_addr::DEPOSIT;
 use eth_trie::{EthTrie as PatriciaTrie, Trie};
 use ethabi::Token;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


DrZoltanFazekas (approved):

zilliqa/src/state.rs
```diff
@@ -176,6 +185,12 @@ impl State {
         )?;
         state.force_deploy_contract_evm(deposit_data, Some(contract_addr::DEPOSIT))?;
 
+        // Set DEPOSIT contract to total deposited at genesis
+        state.mutate_account(contract_addr::DEPOSIT, |a| {
```
JamesHinshelwood, line 189: I think we should do this instead by changing the deposit contract constructor to require a `msg.value` equal to the deposits specified and then pass the value in the transaction call above.

shawn-zil: Alright. I'll try that in the follow up.

shawn-zil: Fixed in #1948

PR #1946, opened 2024-12-03
Deposit contract upgradeable via EIP 1967

86667:
Use EIP 1967 proxy for deposit contract upgrade-ability.

We initialise the deposit contract with `/contracts/deposit_v1.sol` and update immediately to `/contracts/deposit_v2.sol`. The plan for future upgrades would be to call `upgrade_deposit_contract()` with a new deposit implementation only after a certain block height is reached.


zilliqa/src/contracts/deposit_v1.sol
```diff
@@ -0,0 +1,291 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
86667, line 1: Reviewer please compare this file again current main's `zzq2/zilliqa/src/contracts/deposit.sol`

zilliqa/src/contracts/deposit_v2.sol
```diff
@@ -166,44 +169,81 @@ contract Deposit {
...
+        address newImplementation
+    ) internal virtual override onlyOwner {}
+
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() {
+        _disableInitializers();
+    }
+
+    function initialize(
+        address initialOwner,
```
DrZoltanFazekas, line 234: The initializer is only needed in v1. Subsequent version have a reinitializer. We can remove the entire `initialize()` function from v2.

zilliqa/src/contracts/deposit_v1.sol
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+
+    // automatically incrementing the version number allows for
+    // upgrading the contract without manually specifying the next
+    // version number in the source file - use with caution since
+    // it won't be possible to identify the actual version of the
+    // source file without a hardcoded version number, but storing
+    // the file versions in separate folders would help
+    function reinitialize() public reinitializer(version() + 1) {}
```
DrZoltanFazekas, line 268: The reinitializer is only needed in v2 and subsequent versions, we can remove the `reinitialize()` function in v1.

zilliqa/src/contracts/deposit_v1.sol
```diff
@@ -0,0 +1,293 @@
...
+    function __Deposit_init(address initialOwner) internal onlyInitializing {
+        __Ownable2Step_init_unchained();
+        __Ownable_init_unchained(initialOwner);
+        __UUPSUpgradeable_init_unchained();
+    }
+
+    function _authorizeUpgrade(
+        address newImplementation
+    ) internal virtual override onlyOwner {}
+
```
DrZoltanFazekas, line 195: I would remove `onlyOwner()` here add a single line into the `_authorizeUpgrade()` function:
```
require(msg.sender == address(0), "system contract must be upgraded by the system");
```
and then we can completely remove `Ownable2StepUpgradeable` from the base contracts.

zilliqa/src/contracts/deposit_v2.sol
```diff
@@ -166,44 +169,81 @@ contract Deposit {
...
+
+    function __Deposit_init(address initialOwner) internal onlyInitializing {
+        __Ownable2Step_init_unchained();
+        __Ownable_init_unchained(initialOwner);
+        __UUPSUpgradeable_init_unchained();
+    }
+
+    function _authorizeUpgrade(
+        address newImplementation
+    ) internal virtual override onlyOwner {}
```
DrZoltanFazekas, line 226: Same as in v1, instead of making the contract ownable I'd check if the caller is the zero address i.e. the system (our Rust code)

zilliqa/src/contracts/deposit_v1.sol
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    function version() public view returns (uint64) {
+        return _getInitializedVersion();
+    }
+
+    function __Deposit_init(address initialOwner) internal onlyInitializing {
+        __Ownable2Step_init_unchained();
+        __Ownable_init_unchained(initialOwner);
+        __UUPSUpgradeable_init_unchained();
```
DrZoltanFazekas, line 189: This pattern with the `__ContractName_init()` is only used in base contracts i.e. if other contracts inherited from the `Deposit` contract but this is not the case. So we can get rid of this function and call `__UUPSUpgradeable_init_unchained();`  in the `initialize()` function. As I mentioned below, I'd remove `Ownable2StepUpgradeable` from the base contracts so that we don't need the other two calls.

zilliqa/src/contracts/deposit_v1.sol
DrZoltanFazekas: We will have to include all versions of the contract into every release and upgrade to the next higher version at the specified block heights as you mentioned above. Therefore I'd suggest to use hardcoded version numbers i.e. `deposit_v<n>.sol` in the file name and `reinitializer(<n>)` in the modifier i.e. `reinitilizer(2)` above.

zilliqa/src/contracts/mod.rs
```diff
@@ -1,17 +1,51 @@
...
 
     static CONTRACT: Lazy<Contract> =
-        Lazy::new(|| contract("src/contracts/deposit.sol", "Deposit"));
+        Lazy::new(|| contract("src/contracts/deposit_v1.sol", "DepositInit"));
     pub static CONSTRUCTOR: Lazy<Constructor> =
         Lazy::new(|| CONTRACT.abi.constructor().unwrap().clone());
+    pub static INITIALIZE: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("initialize").unwrap().clone());
+    pub static REINITIALIZE: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("reinitialize").unwrap().clone());
```
DrZoltanFazekas, line 17: Remove the `reinitialize` function from v1

zilliqa/src/contracts/mod.rs
```diff
@@ -1,17 +1,51 @@
...
+    pub static UPGRADE_TO_AND_CALL: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("upgradeToAndCall").unwrap().clone());
+    pub static VERSION: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("version").unwrap().clone());
+    pub static GET_STAKERS: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("getStakers").unwrap().clone());
+    pub static BYTECODE: Lazy<Vec<u8>> = Lazy::new(|| CONTRACT.bytecode.clone());
+    pub static OWNER: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("owner").unwrap().clone());
```
DrZoltanFazekas, line 26: If we get rid of `Ownable2StepUpgradeable` then there is not `owner` function

zilliqa/src/contracts/mod.rs
```diff
@@ -1,17 +1,51 @@
...
+    pub static REINITIALIZE: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("reinitialize").unwrap().clone());
+    pub static UPGRADE_TO_AND_CALL: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("upgradeToAndCall").unwrap().clone());
+    pub static VERSION: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("version").unwrap().clone());
 
     pub static BYTECODE: Lazy<Vec<u8>> = Lazy::new(|| CONTRACT.bytecode.clone());
+    pub static OWNER: Lazy<Function> =
+        Lazy::new(|| CONTRACT.abi.function("owner").unwrap().clone());
```
DrZoltanFazekas, line 48: Same as above

zilliqa/src/state.rs
```diff
@@ -351,7 +389,9 @@ pub mod contract_addr {
     pub const INTERSHARD_BRIDGE: Address = Address::new(*b"\0\0\0\0\0\0\0\0ZQINTERSHARD");
     /// Address of the shard registry - only present on the root shard.
     pub const SHARD_REGISTRY: Address = Address::new(*b"\0\0\0\0\0\0\0\0\0\0\0\0\0ZQSHARD");
-    pub const DEPOSIT: Address = Address::new(*b"\0\0\0\0\0\0\0\0\0\0ZILDEPOSIT");
+    /// Address of EIP 1967 proxy for Deposit contract
+    pub const DEPOSIT_PROXY: Address = Address::new(*b"\0\0\0\0\0ZILDEPOSITPROXY");
+    pub const DEPOSIT_PROXY_OWNER: Address = Address::new(*b"ZILDEPOSITPROXYOWNER");
```
DrZoltanFazekas, line 394: As mentioned above, can't we use the zero address instead which can't be impersonated. neither by EOAs nor by other contracts to maliciously upgrade the deposit contract?

DrZoltanFazekas (commented):
Nice work! Besides a few minor comments, my main question is whether we can simplify everything by getting rid of `Ownable2StepUpgradeable` and  requiring `msg.sender == address(0)` for the authorization of upgrades (since the only is not used for anything else).

DrZoltanFazekas (approved):

PR #1948, opened 2024-12-04
Fix genesis deposit balances in Solidity, instead of Rust.

shawn-zil:


shawn-zil:
Closing this as it has diverged significantly from #1946

86667:
Rebased! Github wants a new PR due to force-push. Re-opened here - https://github.com/Zilliqa/zq2/pull/1953

PR #1949, opened 2024-12-04
Fix converter crashes

bzawisto:
A few fixes to fix crashes during conversion of transition/errrors and other fields. In addition I pulled all the scilla errors from zq1 which already existed (and I think will exist in scilla for zq2).


JamesHinshelwood (approved):
Thanks

PR #1953, opened 2024-12-04
Fix genesis deposit balances in Solidity.

86667:
Clone of https://github.com/Zilliqa/zq2/pull/1948

```
curl -X POST   -H "Content-Type: application/json"   --data '{"jsonrpc":"2.0","method":"eth_getBalance","params":["0x00000000005A494C4445504F53495450524F5859", "latest"],"id":1}' http://localhost:4201/
{"jsonrpc":"2.0","id":1,"result":"0x2116545850052128000000"}
```

### DIFF
--- a/zilliqa/src/contracts/deposit_v1.sol
+++ b/zilliqa/src/contracts/deposit_v1.sol
@@ -137,7 +137,7 @@ contract DepositInit is UUPSUpgradeable {
         uint256 _maximumStakers,
         uint64 _blocksPerEpoch,
         InitialStaker[] memory initialStakers
-    ) public initializer {
+    ) public payable initializer {
         __UUPSUpgradeable_init_unchained();
         DepositStorage storage $ = _getDepositStorage();
 
@@ -192,6 +192,11 @@ contract DepositInit is UUPSUpgradeable {
 
             emit StakerAdded(blsPubKey, block.number, amount);
         }
+
+        require(
+            msg.value == committee().totalStake,
+            "stake value does not match total"
+        );
     }
 
     function currentEpoch() public view returns (uint64) {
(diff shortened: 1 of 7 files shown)


zilliqa/src/contracts/deposit_v1.sol
```diff
@@ -137,7 +137,7 @@ contract DepositInit is UUPSUpgradeable {
         uint256 _maximumStakers,
         uint64 _blocksPerEpoch,
         InitialStaker[] memory initialStakers
-    ) public initializer {
+    ) public payable initializer {
```
JamesHinshelwood, line 140: Don't we need to change v2 as well?

zilliqa/src/exec.rs
```diff
@@ -454,13 +454,14 @@ impl State {
         &mut self,
         creation_bytecode: Vec<u8>,
         override_address: Option<Address>,
+        amount: Option<u128>,
```
JamesHinshelwood, line 457: nit: `amount: None` doesn't make sense. Callers should just pass `0` if they don't want to send anything.

JamesHinshelwood (approved):

zilliqa/src/contracts/deposit_v1.sol
shawn-zil: v2 doesn't seem to have this function.

zilliqa/src/exec.rs
shawn-zil: Small fix, I'll do that now.

zilliqa/src/contracts/deposit_v1.sol
86667: v1 does the initialising, further versions inherit the previous version's storage and so do not need to initialise.

JamesHinshelwood: Makes sense, thanks for explaining.

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #1954, opened 2024-12-04
Improve full block production benchmark

JamesHinshelwood:
We create a dummy network with 2 validators - 'big' which has a large proportion of the stake and 'tiny' which has a small amount of stake. The intention is that 'big' will always be the block proposer, because the proposer is selected in proportion to the validators' relative stake. However, 'tiny' will still get to have a vote on this proposal, despite its vote not being needed to reach a supermajority. The benchmark will execute the following in each iteration:

1. Get 'big' to process the previous vote and propose a block
2. Get 'tiny' to vote on this block
3. Get 'big' to vote on this block

Step 2 is important, because we want to measure the time it takes a validator to vote on a block it hasn't seen before. In step 3, 'big' will skip most of the block validation logic because it knows it built the block itself.

### DIFF
--- a/zilliqa/benches/it.rs
+++ b/zilliqa/benches/it.rs
@@ -14,6 +14,7 @@ use revm::primitives::{Bytes, TxKind};
 use tempfile::tempdir;
 use tokio::sync::mpsc;
 use zilliqa::{
+    cfg::{Amount, GenesisDeposit, NodeConfig},
     consensus::Consensus,
     crypto::{Hash, SecretKey},
     db::Db,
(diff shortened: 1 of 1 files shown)


JamesHinshelwood:
Current timings on my machine for reference:
```
produce-full/produce-full
                        time:   [1.6645 s 1.7765 s 1.8760 s]
                        thrpt:  [0.5331  elem/s 0.5629  elem/s 0.6008  elem/s]
```

86667 (approved):
:raised_hands: 

Would you mind leaving some instructions somewhere for usage? Especially around producing those cool fire graphs.

PR #1961, opened 2024-12-05
Update prototestnet config file

bzawisto:


frankmeds (commented):
which key has been used to generate these values?

PR #1962, opened 2024-12-05
fix: missed this out in #1942.

shawn-zil:


JamesHinshelwood (approved):
Thanks! We should probably de-dup this code at some point.

PR #1966, opened 2024-12-05
feat: Add solhint and fix warnings

86667:


.github/workflows/ci.yaml
```diff
@@ -26,9 +26,9 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: "16"
-      - run: cd evm_scilla_js_tests
```
86667, line 29: Directory of shell is reset for each `run` command so this `cd` has no effect on following commands

JamesHinshelwood (approved):

PR #1967, opened 2024-12-06
DB blocks query and availiability timeout tweaks

bzawisto:


JamesHinshelwood (dismissed):

zilliqa/src/db.rs
```diff
@@ -584,7 +584,7 @@ impl Db {
             .lock()
             .unwrap()
             .query_row_and_then(
-                "SELECT height FROM blocks WHERE is_canonical = TRUE ORDER BY height DESC LIMIT 1",
+                "SELECT height from (SELECT height, is_canonical FROM blocks ORDER BY height DESC) WHERE is_canonical = 1 LIMIT 1",
```
JamesHinshelwood, line 587: Worth adding a comment here stating that the brackets make sure SQLite uses the index, so someone clever doesn't come along later and remove them :)

JamesHinshelwood (approved):

PR #1969, opened 2024-12-06
Add ability to enable/disable APIs via configuration

JamesHinshelwood:
By default all JSON-RPC API methods are now disabled. API methods must be enabled with the `enabled_apis` property in configuration. Entire namespaces can be enabled at once with a string (e.g. `"eth"`) or individual methods can be enabled with an object of the form `{ namespace = "eth", apis = ["blockNumber"] }`. For the purposes of this configuration, any method without a namespace (i.e. without an underscore) is assumed to be under the `zilliqa` namespace.

Nodes are also now able to start the JSON-RPC server on more than one port, each of which can have different configuration.


README.md
```diff
@@ -46,6 +46,15 @@ Pass the path to configuration files on the command line with `-c` or `--config-
 If multiple configuration files are provided, they will be merged together.
 If a configuration key occurs in more than one configuration file, the process will exit with an error.
 
+By default, a node will not expose any RPC methods.
+To enable APIs, you must set `enabled_apis` under the [[nodes]] object in the configuration file.
+Each item in `enabled_apis` must either be:
+
+* A string such as `"eth"`, which enables all API methods under the `eth_` namespace.
+* An object of the form `{ namespace = "eth", apis = ["blockNumber"] }`, which enables specific API methods.
```
JamesHinshelwood, line 54: Is this the right level of granularity for enabling APIs? This lets you enable an entire namespace or specific API methods. One option that is deliberately missing is "Enable all APIs in this namespace, except for these". Is this okay?

README.md
```diff
@@ -46,6 +46,15 @@ Pass the path to configuration files on the command line with `-c` or `--config-
 If multiple configuration files are provided, they will be merged together.
 If a configuration key occurs in more than one configuration file, the process will exit with an error.
 
+By default, a node will not expose any RPC methods.
```
JamesHinshelwood, line 49: Is disabling all APIs a good default? I figured it was best for the default to be maximally safe and to force the node operator to consider what APIs they want to enable, but I can see arguments the other way.

z2/src/chain/node.rs
```diff
@@ -515,17 +515,29 @@ impl ChainNode {
...
+            json!(["erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa"])
+        } else {
+            // Only enable `eth_blockNumber` for other nodes.
+            json!([
+                {
+                    "namespace": "eth",
+                    "apis": ["blockNumber"],
+                },
+            ])
+        };
```
JamesHinshelwood, line 529: Does it make sense for us to enable all APIs on API nodes and only have `eth_blockNumber` enabled on other nodes?

zilliqa/src/node_launcher.rs
```diff
@@ -125,30 +125,28 @@ impl NodeLauncher {
...
-            let middleware = tower::ServiceBuilder::new().layer(HealthLayer).layer(cors);
-            let port = config.json_rpc_port;
-            let server = jsonrpsee::server::ServerBuilder::new()
-                .set_http_middleware(middleware)
-                .set_id_provider(EthIdProvider)
-                .build((Ipv4Addr::UNSPECIFIED, port))
-                .await;
+        trace!("Launching JSON-RPC server");
+        // Construct the JSON-RPC API server. We inject a [CorsLayer] to ensure web browsers can call our API directly.
+        let cors = CorsLayer::new()
```
86667, line 130: Do we want to build this if `enabled_apis` is empty? It may add complexity when running multiple nodes locally as different `config.json_rpc_port` will have to be set for each node

z2/src/chain/node.rs
JamesHinshelwood: Yeah possibly - I guess I was just being as conservative as possible but its true that all `eth_` APIs are _probably_ safe?

zilliqa/src/node_launcher.rs
JamesHinshelwood: Good idea - I'll skip this if `enabled_apis` is empty, thanks!

infra/config_rpc_disabled.toml
```diff
@@ -1,6 +1,7 @@
 otlp_collector_endpoint = "http://otel-collector:4317"
 
 [[nodes]]
+enabled_apis = [ "admin", "erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa" ]
```
86667, line 4: This config should be empty! As the equivalent of `disable_rpc`

JamesHinshelwood: I'm dumb, fixed - Thanks!

z2/src/chain/node.rs
86667: Most of them come in handy when testing, we can always remove later :)

README.md
86667: It would be useful to be able to disable only some methods. Happy to merge as is though.

zilliqa/src/node_launcher.rs
```diff
@@ -125,7 +125,7 @@ impl NodeLauncher {
 
         let rpc_module = api::rpc_module(Arc::clone(&node));
 
-        if !config.disable_rpc {
+        if !config.enabled_apis.is_empty() {
```
JamesHinshelwood, line 128: After discussing with @DrZoltanFazekas , we think its important to be able to expose the API on more than one port with different configuration. I will make that change tomorrow.

z2/src/chain/node.rs
```diff
@@ -515,17 +515,28 @@ impl ChainNode {
...
+        let public_api = if self.role == NodeRole::Api {
+            // Enable all APIs, except `admin_` for API nodes.
+            json!({ "port": 4201, "enabled_apis": ["erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa"] })
+        } else {
+            // Only enable `eth_blockNumber` for other nodes.
+            json!({"port": 4201, "enabled_apis": [ { "namespace": "eth", "apis": ["blockNumber"] } ] })
+        };
+        // 4202 is not exposed, so enable everything for local debugging.
+        let private_api = json!({ "port": 4202, "enabled_apis": ["admin", "erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa"] });
+        let api_servers = json!([public_api, private_api]);
```
JamesHinshelwood, line 528: @DrZoltanFazekas @frankmeds Please review and resolve this if you're happy.

86667 (approved):

PR #1973, opened 2024-12-09
fix: DEVOPS-1772 Remove logs from previous containers

frankmeds:


pavlops (approved):

JamesHinshelwood (commented):
Will we still be able to view the logs of containers that crashed and a new one has taken its place?

PR #1994, opened 2024-12-10
Handle non-whitelisted contracts by fully reverting txn if gas limit is 21k

bzawisto:
As title says, we want to implement the following behaviour:

1. If gas = 21k and whitelisted -> allow it to run with gas::max_64()
2. if gas = 21k and NOT whitelisted -> revert entire txn
3. Otherwise, apply normal rules

### DIFF
--- a/zilliqa/src/precompiles/scilla.rs
+++ b/zilliqa/src/precompiles/scilla.rs
@@ -338,8 +338,36 @@ pub fn scilla_call_handle_register<I: ScillaInspector>(
         }
 
         let gas = Gas::new(inputs.gas_limit);
-        let outcome =
-            scilla_call_precompile(&inputs, gas.limit(), &mut ctx.evm.inner, &mut ctx.external);
+        let gas_exempt = ctx
+            .external
+            .scilla_call_gas_exempt_addrs
+            .contains(&inputs.caller);
+
+        // The behaviour is different for contracts having 21k gas and/or deployed with zq1
+        // 1. If gas == 21k and gas_exempt -> allow it to run with gas_left()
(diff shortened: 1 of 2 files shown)


zilliqa/src/precompiles/scilla.rs
```diff
@@ -338,8 +338,36 @@ pub fn scilla_call_handle_register<I: ScillaInspector>(
...
-            scilla_call_precompile(&inputs, gas.limit(), &mut ctx.evm.inner, &mut ctx.external);
+        let gas_exempt = ctx
+            .external
+            .scilla_call_gas_exempt_addrs
+            .contains(&inputs.caller);
+
+        // The behaviour is different for contracts having 21k gas and/or deployed with zq1
+        // 1. If gas == 21k and gas_exempt -> allow it to run with gas_left()
+        // 2. if gas == 21k and NOT gas_exempt -> mark entire txn as failed (not only the current precompile)
+        // 3. Otherwise, let it run with what it's given and let the caller decide
```
JamesHinshelwood, line 349: It seems very odd to me that the behaviour is different if you pass 21000 gas vs 21001 gas. Why?

DrZoltanFazekas: To distinguish the incorrect contracts imported from ZQ1 that used exactly 21000 from the correct ones deployed later. We need a workaround for the imported contracts that may be forgotten in the whitelist to prevent that their Scilla execution fails without reverting the whole transaction which is clearly a bug in `ScillaConnector.sol` with bad consequences for Plunderswap and other protocols.

JamesHinshelwood (approved):

PR #2000, opened 2024-12-11
feat: DEVOPS-1774 Whitelist EVM contracts imported from ZQ1 that use scilla_call

frankmeds:


z2/src/chain/instance.rs
```diff
@@ -18,7 +18,7 @@ pub struct ChainInstance {
 
 impl ChainInstance {
     pub async fn new(config: NetworkConfig) -> Result<Self> {
-        let chain = <Chain as std::str::FromStr>::from_str(&config.name.clone())?;
+        let chain = Chain::parse(&config.name.clone())?;
```
JamesHinshelwood, line 21: nitpicking a bit, but `Chain::from_str` should work fine here?

z2/src/chain/node.rs
```diff
@@ -516,6 +516,20 @@ impl ChainNode {
         let eth_chain_id = self.eth_chain_id.to_string();
         let bootstrap_public_ip = selected_bootstrap.machine.external_address;
 
+        let contracts = self.chain()?.get_whitelisted_evm_contracts();
+        let whitelisted_evm_contract_addresses = if contracts.is_empty() {
+            String::from("[ ]")
+        } else {
+            format!(
```
JamesHinshelwood, line 523: This code would be a lot cleaner if we can just pass a `serde_json::Value` into the `Tera` `Context`.

I have made a similar change here, which might be worth moving into this PR if you agree?

JamesHinshelwood (approved):

PR #2004, opened 2024-12-12
Use systemd watchdog to restart a node

shawn-zil:
This uses an application layer detection to determine if a node is stuck. If so, it triggers `systemd` to restart the service.


shawn-zil:
Turns out that this is the wrong mechanism, as the service runs under docker, with an external systemd. So, closing this and reworking it to perform an internal restart of the node instead.

PR #2012, opened 2024-12-13
Fix for failing scilla_call with whitelisted contracts

bzawisto:
If `scilla_call` fails and it's a whitelisted contract -> mart entire txn as failed.


JamesHinshelwood:
Replaced by #2016

PR #2013, opened 2024-12-13
feat: DEVOPS-1769 Automate starting validator from checkpoint

frankmeds:


pavlops (dismissed):

pavlops (approved):

z2/src/chain/node.rs
```diff
@@ -548,6 +548,54 @@ impl ChainNode {
...
+                    crate::utils::string_decimal_to_hex(&checkpoint_file.replace(".dat", ""))?;
+
+                let json_response = self
+                    .chain
+                    .run_rpc_call(
+                        "eth_getBlockByNumber",
+                        &Some(format!("[\"{}\", false]", checkpoint_hex_block)),
+                        30,
+                    )
+                    .await?;
```
JamesHinshelwood, line 566: @DrZoltanFazekas Do you think its okay for the infra tooling to obtain the checkpoint's block hash itself like this or should we force the user to specify it themselves from a source they trust?

PR #2014, opened 2024-12-13
Slow node start with checkpoint sync

bzawisto:
Storage elements are inserted into memory and once the threshold is met - they are copied into final (sqlite) db with batch inserts. This speeds loading up by ~25%.

### DIFF
--- a/zilliqa/src/db.rs
+++ b/zilliqa/src/db.rs
@@ -414,7 +414,7 @@ impl Db {
             }
         }
 
-        // Helper function used for inserting entries from memory (which backs storage trie) into state_trie db (for now it's sql-lite)
+        // Helper function used for inserting entries from memory (which backs storage trie) into persistent storage
         let db_flush = |db: Arc<TrieStorage>, cache: Arc<MemoryDB>| -> Result<()> {
             let mut cache_storage = cache.storage.write();
             let (keys, values): (Vec<_>, Vec<_>) = cache_storage.drain().unzip();


zilliqa/src/db.rs
```diff
@@ -414,8 +414,41 @@ impl Db {
             }
         }
 
+        // Helper function used for inserting entries from memory (which backs storage trie) into state_trie db (for now it's sql-lite)
```
JamesHinshelwood, line 417:
```suggestion
        // Helper function used for inserting entries from memory (which backs storage trie) into persistent storage
```
(so we don't have an outdated comment when we merge redb :) )

zilliqa/src/db.rs
```diff
@@ -414,8 +414,41 @@ impl Db {
...
+                let mut cache_storage = cache.storage.write();
+                let mut storage_iter = cache_storage.iter();
+                debug!("Doing write to db with total items {}", cache_storage.len());
+                loop {
+                    let mut keys = Vec::with_capacity(max_batch_size);
+                    let mut values = Vec::with_capacity(max_batch_size);
+
+                    storage_iter
+                        .by_ref()
+                        .take(max_batch_size)
```
JamesHinshelwood, line 429: Doesn't `TrieStorage::insert_batch` handle the batching by itself anyway? Why do we need to repeat it here?

i.e. if I pass more than ~32000 key-value pairs to `insert_batch` it splits them up by itself?

zilliqa/src/db.rs
```diff
@@ -414,8 +414,41 @@ impl Db {
             }
         }
 
+        // Helper function used for inserting entries from memory (which backs storage trie) into state_trie db (for now it's sql-lite)
+        let db_flush =
+            |db: Arc<TrieStorage>, cache: Arc<MemoryDB>, max_batch_size: usize| -> Result<()> {
+                let mut cache_storage = cache.storage.write();
+                let mut storage_iter = cache_storage.iter();
```
JamesHinshelwood, line 421: Instead of doing `.iter()` then `.clear()` can we just `.drain()` here?

zilliqa/src/db.rs
```diff
@@ -474,6 +509,15 @@ impl Db {
                     "Invalid checkpoint file: account trie root hash mismatch: calculated {}, checkpoint file contained {}", hex::encode(account_trie.root_hash()?.as_slice()), hex::encode(account_trie_root)
                 ));
             }
+            if processed_storage_items > maximum_sql_parameters {
```
JamesHinshelwood, line 512: I don't think the maximum SQL parameters matters here. The reason we're occasionally flushing to disk is to limit the memory usage of `mem_storage`? So can we instead just pick an arbitrary large-ish number for the maximum number of kv pairs we keep in memory at once?

zilliqa/src/db.rs
```diff
@@ -1166,45 +1216,12 @@ pub struct TrieStorage {
...
-            "INSERT OR REPLACE INTO state_trie (key, value) VALUES (?1, ?2)",
-            (key, &value),
-        )?;
-        let _ = self.cache.lock().unwrap().insert(key.to_vec(), value);
-        Ok(())
-    }
-
-    fn insert_batch(&self, keys: Vec<Vec<u8>>, values: Vec<Vec<u8>>) -> Result<(), Self::Error> {
+impl TrieStorage {
+    pub fn write_batch(
```
JamesHinshelwood, line 1220: Maybe nitpicking - Is this called anywhere other than `insert_batch` in the `eth_trie::DB` implementation? Why do we need to separate it out and add an extra function call?

JamesHinshelwood (approved):
Nice idea, thanks! A few comments that might be worth considering before we merge.

zilliqa/src/db.rs
JamesHinshelwood: Ah yeah that's a fair point :+1:

JamesHinshelwood (approved):
Thanks!

PR #2016, opened 2024-12-13
Fail the whole EVM transaction if an internal Scilla call fails

JamesHinshelwood:
(if the contract is whitelisted)

This PR also adds support for specifying 'forks' in configuration which lets us diverge execution behaviour based on the block height.


z2/src/chain.rs
```diff
@@ -151,6 +152,22 @@ impl Chain {
         }
     }
 
+    pub fn get_forks(&self) -> Vec<Value> {
+        match self {
+            Chain::Zq2ProtoTestnet => vec![
+                json!({ "at_height": 0, "failed_scilla_call_from_gas_exempt_caller_causes_revert": false }),
+                // estimated: 2024-12-18 11:57:19Z
```
DrZoltanFazekas, line 159: Just a typo in the comment, should be `2024-12-18T11:57:19Z`
but let's activate the fork on the proto-testnet on Tuesday
and on the proto-mainnet 2 days later, on Thursday

JamesHinshelwood: I've updated, thanks!

86667 (approved):

PR #2017, opened 2024-12-13
API updates

maxconway:
- Adds deprecation status tags to deprecated endpoints (mainly ones pertaining to DS blocks)
- Removes partially implemented status tag gettxblock, which is now implemented


rrw-zilliqa (approved):
Looks good to me :-)

PR #2018, opened 2024-12-15
Bump outdated evm_js_tests modules.

saeed-zil:
* Use `pnpm` instead of `npm`!
* Remove web3.js dependency. (web3.js hardhat plugin is outdated and doesn't support the latest version of web3.js)
* Remove useless tests.
* Fix compilation errors.
* Use a newer version of `nodejs` (18.20)


scripts/js_test.sh
```diff
@@ -22,16 +22,17 @@ export NVM_DIR="$HOME/.nvm"
 [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
 
 echo "Installing nvm"
-nvm install 16.0
+nvm install 18.20
 
 echo "Using nvm 16"
```
JamesHinshelwood, line 27:
```suggestion
echo "Using nvm 18"
```
(or just remove this message)

Shouldn't we just update to the latest version of node now?

JamesHinshelwood (dismissed):
Thanks!

chetan-zilliqa (approved):

PR #2027, opened 2024-12-16
feat: new view cache

86667:


zilliqa/src/consensus.rs
```diff
@@ -554,7 +559,14 @@ impl Consensus {
...
             {
-                return Ok(Some(self.build_new_view()?));
+                if let Some((_, ExternalMessage::NewView(new_view))) =
+                    self.new_view_message_cache.as_ref()
+                {
+                    if new_view.view == self.get_view()? {
+                        return Ok(self.new_view_message_cache.clone());
+                    }
+                }
+                self.new_view_message_cache = None;
```
JamesHinshelwood, line 569: If we clear this straight away, won't we only ever send a single repeated `NewView`?

JamesHinshelwood: Maybe a better check is something like `if self.view == self.new_view_message_cache.1.view`? (i.e. If I've sent and cached a `NewView` in this view already, then re-send it?

86667: That is what this code intends to do!

If were in the same view then `return` with the `NewView` message. If were not in the same view then clear the cache.

Its more ugly than `self.new_view_message_cache.1.view` to avoid unwrapping of Options

86667: I've moved the clear cache line inside of the `if let` to be safer.

JamesHinshelwood (approved):
Thanks!

Issue #1878, opened 2024-11-22
Validators should not serve `BlockRequests`


DrZoltanFazekas:
Syncing nodes can't distinguish the current validators from other nodes since their latest block contains an outdated validator set. But ~~they~~ validators can advertise that they don't have any blocks so that nodes don't send them request for blocks that they want to download.

shawn-zil:
@DrZoltanFazekas 
If I understood this issue, it is that nodes which are busy catching up (syncing), should not be queried for blocks, since their database is incomplete anyway. The requestor distributes the requests across different peers, and does not know which ones are syncing.

Is this right?

Also, am I correct in thinking that nodes which are busy syncing, should probably not participate in other requests (e.g. voting) either since they're still syncing.

shawn-zil:
When we alter the Validators so that they will always respond with a 'null' response (to indicate that they don't have any blocks) the requesting node retries the range with a different peer. This is desired behaviour.

However, it creates a potential dead-lock. By definition, the non-Validators are generally 'behind' the Validators. In the rare case where restarted Validators are trying to sync, they may not be able to if the only nodes that have the latest blocks are other Validators and if the latest blocks have not propagated to non-validators.

As seen in this example where the Validators are requesting for blocks [770..772] but the non-Validators only have < 771.
```
zilliqanode2  | 2024-12-19T03:08:42.714937Z  WARN zilliqa::node: 292: block_store::BlockRequest : sending empty block response.
zilliqanode2  | 2024-12-19T03:08:42.716387Z  WARN zilliqa::block_store: 1014: block_store::request_blocks() : Could not find peers for views RangeMap { ranges: [770..772] }
zilliqanode2  | 2024-12-19T03:08:43.067254Z  INFO zilliqa::node: 967: block response from: 12D3KooWPXw2dXBRH1bT4vcNos9f6W2KoFTiarqptBuTzxaXg7zu len: 0 availability: None
zilliqanode3  | 2024-12-19T03:08:43.069067Z  INFO zilliqa::node: 967: block response from: 12D3KooWNtUCTB4J5MzXc3wawYic4m4pRwTLSL4BzTQc2Csk3VfL len: 1 availability: Some([Latest(10), CachedViewRange(0..771, None)])
```

shawn-zil:
The BlockRequest mechanism seems to be broken somewhere as the blocks are being requested, serviced and cached.

```
[...]

PR #2030, opened 2024-12-17
Get local docker network running again.

shawn-zil:
Took a while to debug this issue - but I observed a lot of TCP retransmissions occurring in the present local docker network using wireshark, which hinted at an underlying networking issue.

Turns out that the nodes were communicating with each other over the gateway acting as a NAT-router. So, they were seeing the gateway IP/port as the remote IP/port. This is fine if it was used for responses (aka NAT) but would have problems if a client tried to initiate a connection to this ephemeral IP/port that isn't listening on the gateway.

This PR does two things:
- It makes node3 a dependency for all other nodes, as it is the bootstrap node.
- It manually adds the remote IP/ports to the list of known peers exploiting Identify.
- It manually adds the external IP/port (as observed from the remote node) to the list of external addresses.

Local docker cluster is starting up again.

I will be working on my #1878 changes on top of this branch. Would be good to get this fix into main.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -361,7 +361,7 @@ dependencies = [
  "serde_json",
  "tokio",
  "tokio-stream",
- "tower 0.5.2",
+ "tower 0.5.1",
  "tracing",
  "wasmtimer",
 ]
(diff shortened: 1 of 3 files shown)


86667 (approved):

shawn-zil:
Just stashing it here that Rust's libp2p implementation implies that Kademlia needs to be hooked up to Identify, manually, which is largely what got added here.
https://docs.rs/libp2p-kad/latest/libp2p_kad/#important-discrepancies

PR #2032, opened 2024-12-17
Deposit_v3 tooling

86667:
- Update `deposit.md` instructions
- Update `convert-key` to generate new deposit auth sig
- Configure `contract_upgrade_block_heights` config var via z2
- Update z2 Validator and Deployer to handle deposit auth sig


z2/docs/deposit.md
```diff
@@ -48,8 +54,9 @@ $ echo '{"secret_key":"96252e38af375be21d9eb30a6b88abc3836acecaeb2240731fb42e029
   z2 deposit --chain zq2-prototestnet \
   --peer-id  12D3KooWGu8PBoj6vMPafnhA2P7sLumSV1NhQJZ2W2AGiBgc5ATW \
   --private-key 96252e38af375be21d9eb30a6b88abc3836acecaeb2240731fb42e0299e14419 \
-  --reward-address 0xe29a3e99a6997B1571DA24d6517e7b3acaFB5d9e \
+  --reward-address 0x3946f9872247af2eb4fe44c81c463e801925b8d4 \
```
DrZoltanFazekas, line 57: We should use the original reward address `0xe29a3e99a6997B1571DA24d6517e7b3acaFB5d9e` in the example, it's not supposed to be the same as the signing address

86667: My bad you did mention this - must have gotten lost in the threads earlier. Updated

z2/docs/deposit.md
```diff
@@ -22,21 +23,26 @@ Usage: z2 deposit --chain <CHAIN_NAME> --public-key <BLS_PUBLIC_KEY> --peer-id <
 * `--private-key <PRIVATE_KEY_OF_WALLET_WITH_MIN_10_MIL_FUND>`: The private key of the wallet that has a minimum stake amount of 10 million.
 * `--amount <AMOUNT_IN_MILLION_ZIL>`: The amount in ZIL to deposit. The valid range is from 10 million to 255 million ZIL, allowing a deposit of up to 255 million ZIL.
 * `--reward-address <REWARD_ADDRESS>`: Specifies the address to receive rewards. You can generate a new wallet address to receive the rewards.
-* `--pop-signature <BLS_POP_SIGNATURE>`: The BLS proof-of-possession signature of the validator node.
+* `--signing-address <SIGNING_ADDRESS>`: Specifies the address which signs cross-chain events.
+* `--deposit-auth-signature <DEPOSIT_AUTH_SIGNATURE>`: The BLS proof-of-possession signature of the validator node.
```
frankmeds, line 27: shall `--deposit-auth-signature` still be called `The BLS proof-of-possession signature of the validator node`?

@DrZoltanFazekas @JamesHinshelwood ?

z2/src/bin/z2.rs
```diff
@@ -444,9 +444,12 @@ struct DepositStruct {
     /// Specify the staking reward address
     #[clap(long, short)]
     reward_address: String,
+    /// Specify the signing address
+    #[clap(long, short)]
+    signing_address: String,
     /// Specify the Validator Proof-of-Possession
```
frankmeds, line 450: shall `--deposit-auth-signature` still be called `Validator proof-of-possession`?

@DrZoltanFazekas @JamesHinshelwood ?

z2/docs/deposit.md
```diff
@@ -22,21 +23,26 @@ Usage: z2 deposit --chain <CHAIN_NAME> --public-key <BLS_PUBLIC_KEY> --peer-id <
...
 
 **Note**: The `--private-key` parameter should be the private key of a wallet that has secured a minimum stake amount of 10 million ZILs.
 
 ### Generating Required Values
-To generate the `public-key`, `pop-signature` and `peer-id`, use the following command inside the zq2 folder. Please pass `PRIVATE_KEY_OF_VALIDATOR` to the command input.
+To generate the `public-key`, `deposit-auth-signature` and `peer-id`, use the following command inside the zq2 folder. Please pass `PRIVATE_KEY_OF_VALIDATOR` and `CHAIN_ID` to the command input. 
+
+By default this tool signs over the address derived from the given secret key to generate the `deposit-auth-signature`, you may want to override this for example if deploying via a Delegation contract. To override pass in `"address": "<CONTROL_ADDRESS>"`.
```
frankmeds, line 34: to have the input parameter clearer to an user, can we call it `control_address`?

@DrZoltanFazekas @JamesHinshelwood

z2/docs/deposit.md
```diff
@@ -22,21 +23,26 @@ Usage: z2 deposit --chain <CHAIN_NAME> --public-key <BLS_PUBLIC_KEY> --peer-id <
...
+{"address":"0x3946f9872247af2eb4fe44c81c463e801925b8d4","deposit_auth_signature":"a53efd8bad058e4e551b7e9681613a278b782acfe05fb98d536bc95029278704adbb85891cdc7fa384ab7d5008fdd42c0ea70404ab4ec07bcf5e738c92b2be88debdc33014852ead1d9976fcbf7760043615ba74f36181fc87db21f8f8997a44","bls_public_key":"825124961d51c99816848875fa505b75f2e62e69937fe9bfa5fa97711845abd667f05bdc3756f7dba6b7e9e0467a3804","peer_id":"12D3KooWGu8PBoj6vMPafnhA2P7sLumSV1NhQJZ2W2AGiBgc5ATW","tx_pubkey":{"Ecdsa":["3056301006072A8648CE3D020106052B8104000A03420004B7C457DC36C75EADA5675629F1CE0FA93534FB76ADFC49840CC050AE2995FC87764AEB8975D049D19FDA6BFF2B3FF51608034A3FC6708F476A0C9306BA5CBE14",true]}}
```
frankmeds, line 44: from this example looks like the address in output to the `convert key` is also the one in input to the signing address, is this correct?

Can we add an example also using an address as input?

@DrZoltanFazekas @JamesHinshelwood

z2/src/chain.rs
```diff
@@ -151,6 +152,17 @@ impl Chain {
         }
     }
 
+    pub fn get_contract_upgrades_block_heights(&self) -> ContractUpgradesBlockHeights {
+        match self {
+            Self::Zq2Devnet => ContractUpgradesBlockHeights {
+                deposit_v3: Some(3600),
```
frankmeds, line 158: is `deposit_v3` expected in devnet to be always at block 3600?

z2/src/bin/z2.rs
DrZoltanFazekas: let's rename it to BLS signature as @JamesHinshelwood suggested above

z2/docs/deposit.md
DrZoltanFazekas: Makes sense to make it more specific than `address` especially because there are two other addresses, the reward address and the signing address and people could get easily confused

z2/docs/deposit.md
DrZoltanFazekas: That's correct, since `z2 deposit` in the example uses the same `PRIVATE_KEY_OF_VALIDATOR` as `convert-key` so the address computed by `convert-key` is the one that z2 will send the deposit transaction from

z2/src/chain.rs
DrZoltanFazekas: In devnet it can be set to 3600 or even less in order to reduce the time we're waiting for the upgrade to happen.

86667: The value determines at which height a particular contract upgrade will occur. So assuming no network `reset` the value for a particular contract (`deposit_v3` in this case) will never change. 

However more contract upgrades may be added at later dates, requiring a network `upgrade`.

PR #2034, opened 2024-12-18
Write database version file to disk

JamesHinshelwood:
This will let us make breaking changes to the database format in the future, without having to write ugly code which has to guess what version of data it is reading.

### DIFF
--- a/zilliqa/src/db.rs
+++ b/zilliqa/src/db.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::BTreeMap,
     fmt::Debug,
-    fs::{self, File},
+    fs::{self, File, OpenOptions},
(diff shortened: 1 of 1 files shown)


DrZoltanFazekas (dismissed):

86667 (dismissed):
This doesnt allow us to ever update the version as the error will always throw. I suppose it forces a re-sync rather than bugs surfacing later on. 

For backwards compatible changes we should allow for different versions up to a block height (using `config.forks`) and then update this version field in the table. Not sure how that would work with  this hard coded `CURRENT_DB_VERSION`. 

Just thinking out loud.

86667 (approved):

Issue #1877, opened 2024-11-22
Dedicated API node for querying large Scilla contract states


DrZoltanFazekas:
Based on #1603 , we should run a customized API node on which the 10MiB does not apply to make Zillion and other apps that fetch the entire SSNList state work. We should not include it in the API load balancer, it should be a node used exclusively by a set of apps that need this feature and we want to support.

shawn-zil:
I've provided a node config `max_rpc_response_size` for use in changing the response payload size (default: 10MB).

PR #2035, opened 2024-12-18
Added max_rpc_response_size config

shawn-zil:
To help address issue #1877 - provide a config setting that allows the maximum RPC response size to be configurable (default: 10MB).


z2/src/setup.rs
```diff
@@ -546,6 +546,7 @@ impl Setup {
                 state_rpc_limit: state_rpc_limit_default(),
                 failed_request_sleep_duration: failed_request_sleep_duration_default(),
                 enable_ots_indices: false,
+                max_rpc_response_size: max_rpc_response_size_default(),
```
frankmeds, line 549: can have sense having in the config file a `max_rpc_response_size_in_mb` or `max_rpc_response_size_in_kb`?

@JamesHinshelwood ?

frankmeds (approved):

z2/src/setup.rs
shawn-zil: Happy to leave it as is, or do it either way.

frankmeds: no strong opinion my side @shawn-zil

shawn-zil: @JamesHinshelwood commented to leave it as Bytes (during the call-chat today). So, I'll leave it as-is and mark this comment as closed @frankmeds

frankmeds (approved):

PR #2042, opened 2024-12-19
Broadcast repeated new view messages to speed up high QC sharing

JamesHinshelwood:
In some cases where the entire network is down, many nodes will have disjoint values of their high QC. The only way for them to sync the high QCs up is to wait for a supermajority of nodes to become the leader of a missed view, meaning it receives `NewView` messages from everyone else and can updates its own high QC.

To speed this up, we now broadcast `NewView`s. We still send the first `NewView` message directly to the leader, but we broadcast repeated messages.

We also change the `NewView` processing logic to ensure nodes will update their high QC based on received messages, even if they are not the leader of the view.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -563,7 +563,9 @@ impl Consensus {
                     self.new_view_message_cache.as_ref()
                 {
                     if new_view.view == self.get_view()? {
-                        return Ok(self.new_view_message_cache.clone());
+                        // When re-sending new view messages we broadcast them, rather than only sending them to the
+                        // view leader. This speeds up network recovery when many nodes have different high QCs.
+                        return Ok(Some((None, ExternalMessage::NewView(new_view.clone()))));
                     }
                     // If new_view message is not for this view then it must be outdated
                     self.new_view_message_cache = None;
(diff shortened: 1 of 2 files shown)


zilliqa/src/consensus.rs
```diff
@@ -563,7 +563,9 @@ impl Consensus {
                     self.new_view_message_cache.as_ref()
                 {
                     if new_view.view == self.get_view()? {
-                        return Ok(self.new_view_message_cache.clone());
+                        // When re-sending new view messages we broadcast them, rather than only sending them to the
+                        // view leader. This speeds up network recovery when many nodes have different high QCs.
+                        return Ok(Some((None, ExternalMessage::NewView(new_view.clone()))));
```
86667, line 568: We could get away with just caching the `NewView` rather than the `(leader, ExternalMessage)` which would simplify things. Im happy to do as a clean up later

zilliqa/src/consensus.rs
```diff
@@ -1822,9 +1810,23 @@ impl Consensus {
...
+        // What happens when there are multiple QCs with different parents?
+        // if we are not the leader of the round in which the vote counts
+        if !self.are_we_leader_for_view(new_view.qc.block_hash, new_view.view) {
+            trace!(new_view.view, "skipping new view, not the leader");
+            return Ok(None);
+        }
+
+        let mut current_view = self.get_view()?;
+        // if the vote is too old and does not count anymore
+        if new_view.view < current_view {
```
86667, line 1826: should this go at the top of the fn?

86667 (approved):

PR #2046, opened 2024-12-19
fix: hex encode deposit auth sig

86667:


frankmeds (changes requested):
can you please replace
    println!("\t--staking-address {} \\", ZERO_ACCOUNT);
with 
    println!("\t--signing-address {} \\", ZERO_ACCOUNT);

z2/src/deployer.rs
```diff
@@ -293,13 +300,16 @@ pub async fn run_deposit(config_file: &str, node_selection: bool) -> Result<()>
...
-                    node.chain_id(),
-                    node_ethereum_address.secret_key.to_evm_address(),
-                )
-                .to_string(),
+            &hex::encode(
+                node_ethereum_address
+                    .secret_key
+                    .deposit_auth_signature(
+                        node.chain_id(),
+                        SecretKey::from_hex(&genesis_private_key)?.to_evm_address(),
```
86667, line 308: Notice different address passed here

z2/src/deployer.rs
```diff
@@ -293,13 +300,16 @@ pub async fn run_deposit(config_file: &str, node_selection: bool) -> Result<()>
...
             &node_ethereum_address.peer_id,
             &node_ethereum_address.bls_public_key,
-            &node_ethereum_address
-                .secret_key
-                .deposit_auth_signature(
-                    node.chain_id(),
-                    node_ethereum_address.secret_key.to_evm_address(),
-                )
-                .to_string(),
+            &hex::encode(
```
JamesHinshelwood, line 303: Isn't it kind of silly to convert the signature to a string here, only for `Validator::new` to immediately parse it again to a signature?

JamesHinshelwood (approved):

z2/src/deployer.rs
86667: Yes, Im going to clean but wanted to get out the working version to unblock today

DrZoltanFazekas: yes, let's have a working version so that we can deposit and continue with the test on the devnet and you can refactor it later

DrZoltanFazekas (dismissed):

frankmeds (approved):

JamesHinshelwood (approved):

PR #2049, opened 2024-12-20
feat: DEVOPS-1775 Adjust order of deployment operations to minimise node downtime when upgrading

frankmeds:
in this PR:
- feat: DEVOPS-1787 Remove rpc check on z2 deployer update
- feat: DEVOPS-1775 Adjust order of deployment operations to minimise node downtime when upgrading

### DIFF
--- a/z2/src/chain/node.rs
+++ b/z2/src/chain/node.rs
@@ -364,20 +364,21 @@ impl ChainNode {
         self.import_config_files().await?;
         self.run_provisioning_script().await?;
 
+        // TODO implement a more effective check
         // Check the node is making progress
-        if self.role != NodeRole::Apps {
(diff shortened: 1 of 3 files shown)


pavlops (approved):

z2/src/chain/node.rs
```diff
@@ -364,20 +364,21 @@ impl ChainNode {
         self.import_config_files().await?;
         self.run_provisioning_script().await?;
 
+        // TODO implement a more effective check
```
JamesHinshelwood, line 367: What doesn't work with this check?

frankmeds: network as prototestnet or protomainnet after upgrade, nodes do not provide responses for a time between 6 to 10 minutes making this check de facto failing and not useful

PR #2102, opened 2025-01-07
feat: deposit_v4 with cheaper depositTopUp

86667:
Work done by `updateLatestComputedEpoch()` within functions called by delegators, ie `depositTopUp()`, `unstake()` and `withdraw()` is expensive in some circumstances, ie when a deposit has occurred and one or more epochs passed since. This is because the new staker key is written to future committees. 

This work aims to remove costly writes to `_committee[]` from the above functions and instead have all writes done in `deposit()`. To achieve this i've changed how `_committee` works slightly such that we only write to `_committee[] `if there are changes and do not copy data if data hasnt changed.

#### Before

deposit owner1: Gas used 410503

Rolling ahead 2 epochs  
deposit owner2: Gas used 644369
  
Rolling ahead 1 epochs
1st top up owner 1: Gas used 179666
1st top up owner 2: Gas used 16149
  

Rolling ahead 2 epochs
2nd top up owner 1: Gas used 273779
2nd top up owner 2: Gas used 16158

#### After

deposit owner1: Gas used 334749
   
Rolling ahead 2 epochs
deposit owner2: Gas used 446585

Rolling ahead 1 epochs
1st top up owner 1: Gas used 14174
1st top up owner 2: Gas used 14174

Rolling ahead 2 epochs
2nd top up owner 1: Gas used 13987
2nd top up owner 2: Gas used 13990


86667:
Migrated with aother solution to a single draft PR - https://github.com/Zilliqa/zq2/pull/2226

PR #2110, opened 2025-01-08
StakerUpdated and StakerMoved events

86667:


zilliqa/src/contracts/deposit_v4.sol
```diff
@@ -592,7 +598,8 @@ contract Deposit is UUPSUpgradeable {
                 // We need to remember to update the moved staker's `index` too.
                 futureCommittee.stakers[lastStakerKey].index = futureCommittee
                     .stakers[blsPubKey]
-                    .index;
+                    .index;  
+                emit StakerMoved(blsPubKey, deleteIndex, blocksPerEpoch() * (currentEpoch() + 2));
```
DrZoltanFazekas, line 602: Shouldn't it be `emit StakerMoved(lastStakerKey, deleteIndex, nextUpdate());` instead?

86667: Yep that works. Pushed

zilliqa/src/contracts/deposit_v4.sol
```diff
@@ -592,7 +598,8 @@ contract Deposit is UUPSUpgradeable {
                 // We need to remember to update the moved staker's `index` too.
                 futureCommittee.stakers[lastStakerKey].index = futureCommittee
                     .stakers[blsPubKey]
-                    .index;
+                    .index;  
+                emit StakerMoved(blsPubKey, deleteIndex, nextUpdate());
```
DrZoltanFazekas, line 602: `blsPubKey` is the staker who being deleted here, but we want to include the key of the last staker who is moved in the deleted staker's place, hence I think it should be `emit StakerMoved(lastStakerKey, deleteIndex, nextUpdate());`

86667: Woops! good spot

PR #2111, opened 2025-01-08
Optimise common cases in `RangeMap::with_range`

JamesHinshelwood:
Most calls to `with_range` just add a single element to the end of the largest existing range. Previously, this required inserting an extra 1-wide range at the end of the list, then iterating through the list again to merge the two final ranges.

Instead, we just add an explicit check for this case (and the case) where the new range is strictly greater than the largest existing range.

There are certainly more efficient ways to solve the general case of `with_range` but this fixes the immediate problem with block syncing speed on my machine, so I would propose merging it anyway.

### DIFF
--- a/zilliqa/src/range_map.rs
+++ b/zilliqa/src/range_map.rs
@@ -7,7 +7,6 @@ use std::{
 };
 
 use serde::{Deserialize, Serialize};
-use tracing::*;
 
 /// A block map - a reasonably efficient, easily implementable representation of a collection of ranges.
 /// Feel free to make this generic - we only ever need the u64 variant so I didn't bother.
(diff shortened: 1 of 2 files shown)


zilliqa/src/range_map.rs
```diff
@@ -177,20 +177,44 @@ impl RangeMap {
...
+            return self;
+        }
+
+        let last = self.ranges.last_mut().unwrap();
+        // Optimise the common case where the new range overlaps with the current greatest range. Note that `end`
+        // is exclusive, since `Range`s are half-open, but we still check `range.start <= last.end`. In the case
+        // where `range.start == last.end`, the ranges are not overlapping, but adjacent and we can still merge
+        // them immediately.
+        if range.start >= last.start && range.start <= last.end {
+            last.end = range.end;
```
DrZoltanFazekas, line 195: Is it possible that `range.end < last.end` ?
If it is, we should only set `last.end = range.end` if `range.end > last.end` imo

JamesHinshelwood: Oh yeah, good spot, thank you - Just to be clear, you are talking about the case where `range` is entirely encompassed by the existing `last` and thus there is nothing to do?

DrZoltanFazekas: Yes, exactly that case

DrZoltanFazekas (approved):

PR #2122, opened 2025-01-09
feat: DEVOPS-1819 Regenerate all the chain specs

frankmeds:


JamesHinshelwood (approved):
We should add some script which re-generates these for all networks.

PR #2123, opened 2025-01-10
Revert "Fix balance of created Scilla contracts if they already have a non-zero balance"

JamesHinshelwood:
Reverts Zilliqa/zq2#2113

We should instead add this change with a flag in the fork configuration.


rrw-zilliqa (approved):
Looks good to me :-)

PR #2125, opened 2025-01-10
Fix for issue where statuscode 255 was being returned even for succesful transactions

maxconway:
`receipt.accepted==Some(false)` is now ignored, and if `receipt.success==true` the transaction is reported dispatched

@saeed-zil I noticed you already assigned yourself the associated issue after I pushed this, my bad.
Please feel free to reject it if's likely to cause a conflict with something you've already got in the works :)


saeed-zil (approved):
LGTM :)

maxconway:
Looks like the test failure is actually in main, just messaged @JamesHinshelwood about it because it seems to come from https://github.com/Zilliqa/zq2/pull/2130

JamesHinshelwood (approved):

PR #2128, opened 2025-01-10
Get Proof implementation

bzawisto:
This PR introduces eth_getProof implementation.

### DIFF
--- a/.github/workflows/deps.yaml
+++ b/.github/workflows/deps.yaml
@@ -59,11 +59,11 @@ jobs:
         uses: actions/checkout@v4
 
       - name: download Cargo.lock from update job
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v5
         with:
           name: Cargo-lock
       - name: download cargo-update log from update job
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v5
         with:
           name: cargo-updates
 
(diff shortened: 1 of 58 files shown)


zilliqa/src/state.rs
```diff
@@ -382,6 +395,43 @@ impl State {
             &bincode::serialize(&account)?,
         )?)
     }
+
+    pub fn get_proof(&self, address: Address, storage_keys: &[B256]) -> Result<Proof> {
+        if !self.has_account(address)? {
+            return Ok(Proof::default());
+        };
+
+        // get_proof() requires &mut so clone state and don't mutate the origin
```
JamesHinshelwood, line 404: Why not make this `&mut self` instead? You already make a clone of it in the API implementation (by calling `.at_root`).

zilliqa/src/state.rs
```diff
@@ -382,6 +395,43 @@ impl State {
...
+            .accounts
+            .get_proof(Self::account_key(address).as_slice())?;
+
+        let mut storage_trie = state.get_account_trie(address)?;
+        storage_trie.root_hash()?;
+
+        let storage_proofs = {
+            let mut storage_proofs = Vec::new();
+            for key in storage_keys {
+                let key = Self::account_storage_key(address, *key);
```
JamesHinshelwood, line 419: @DrZoltanFazekas Are we happy that this exposes our trie encoding as part of our external API (and thus makes it hard to ever change it)?

I ask because I don't think we are encoding things into the trie the same way as Ethereum does and thus there's no guarantee that what we're doing is sensible.

maxconway (approved):
LGTM

PR #2159, opened 2025-01-14
Add logging to API call entry points

86667:


JamesHinshelwood (commented):
It will probably be easier to do this in the `declare_module` macro instead?

86667:
> It will probably be easier to do this in the `declare_module` macro instead?

Good point, thanks

maxconway (approved):
LGTM

JamesHinshelwood (approved):

PR #2155, opened 2025-01-14
feat: add option to adjust contract balance on deployment if already funded

saeed-zil:
Please let me know:
* if you have a better name for the config
* which config file(s) should I change accordingly?

PR #2162, opened 2025-01-15
feat: introduce fork deltas for enhanced fork configuration

saeed-zil:
The target branch will be changed to the main after #2155 is merged.


zilliqa/src/contracts/deposit_v1.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.24;
```
JamesHinshelwood, line 2: Why are we making these changes in this PR?

zilliqa/src/cfg.rs
```diff
@@ -455,11 +499,20 @@ impl Forks {
...
 pub struct Fork {
+    pub at_height: u64,
+    pub failed_scilla_call_from_gas_exempt_caller_causes_revert: bool,
+    pub call_mode_1_sets_caller_to_parent_caller: bool,
+    pub scilla_messages_can_call_evm_contracts: bool,
+    pub scilla_contract_creation_increments_account_balance: bool,
+}
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+pub struct DeltaFork {
```
JamesHinshelwood, line 510: nit: I'd prefer `ForkDelta`

zilliqa/src/cfg.rs
```diff
@@ -397,6 +397,50 @@ impl Default for ConsensusConfig {
     }
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct DeltaForks(Vec<DeltaFork>);
+impl Default for DeltaForks {
```
JamesHinshelwood, line 402: Do we need this `Default` impl? Why?

zilliqa/src/cfg.rs
```diff
@@ -366,7 +366,7 @@ pub struct ConsensusConfig {
     /// Forks in block execution logic. Each entry describes the difference in logic and the block height at which that
     /// difference applies.
     #[serde(default)]
-    pub forks: Forks,
+    pub forks: DeltaForks,
```
JamesHinshelwood, line 369: How about we add `genesis_fork: Fork` which is always required and `forks: ForkDeltas` which can mutate those forks at higher blocks?

This way we can be sure that we always have a complete description of the execution behaviour and don't have to rely on the duplicate `Default` implementations we have here (which I think is error-prone).

JamesHinshelwood (commented):
I think the way this works is slightly messy/error-prone. It would be better to guarantee in the configuration that we list one full `Fork` for the genesis block.

zilliqa/src/contracts/deposit_v1.sol
saeed-zil: It was due to a failure to CI that Tomos resolved, I think.

zilliqa/src/cfg.rs
```diff
@@ -392,28 +419,21 @@ impl Default for ConsensusConfig {
...
             contract_upgrade_block_heights: ContractUpgradesBlockHeights::default(),
-            forks: Default::default(),
+            forks: vec![],
+            genesis_fork: genesis_fork_default(),
         }
     }
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(try_from = "Vec<Fork>", into = "Vec<Fork>")]
```
JamesHinshelwood, line 401: I believe the `serde` impls and `From/To` impls are now unused?

zilliqa/src/cfg.rs
```diff
@@ -535,13 +567,23 @@ pub fn total_native_token_supply_default() -> Amount {
     Amount::from(21_000_000_000_000_000_000_000_000_000)
 }
 
+pub fn genesis_fork_default() -> Fork {
```
JamesHinshelwood, line 570: Shouldn't we copy the old default, with the comment? (where all of these are `true` instead)

>     /// The default implementation of [Forks] returns a single fork at the genesis block, with the most up-to-date
>     /// execution logic.

zilliqa/src/cfg.rs
```diff
@@ -568,3 +610,130 @@ impl Default for ContractUpgradesBlockHeights {
...
+                scilla_messages_can_call_evm_contracts: None,
+                scilla_contract_creation_increments_account_balance: Some(true),
+            }],
+            ..Default::default()
+        };
+
+        let forks = config.get_forks().unwrap();
+        assert_eq!(forks.0.len(), 2);
+        assert_eq!(forks.0[0].at_height, 0);
+        assert_eq!(forks.0[1].at_height, 10);
```
JamesHinshelwood, line 648: I think these tests should be using `Forks::get` instead to get the fork at a specific block height.

That also gives us the chance to test the boundary cases to ensure we don't have any off by one errors (I was worried about this when implementing `get`)

JamesHinshelwood (commented):
LGTM, thanks! Just a few small niggles left I think :)

JamesHinshelwood (approved):

PR #2164, opened 2025-01-15
(fix) Pass strings to scilla as strings, rather than arrays of their bytes

### LINKED ISSUE #2163: Passing a string through the scilla_call precompile causes the txn to fail
- Write a scilla transition which takes a string (in this case '/ipfs/something').
 - Try to call this via the scilla call precompile.
 - Observe the txn fails 

(the ERC721 EVM proxy is a good example of this).

This is because we try to encode the string as a series of bytes and then attempt to parse it as JSON, discover it isn't valid JSON and fault out.

rrw-zilliqa:
Fixes #2163


JamesHinshelwood:
Would a transaction have been able to be mined with the previous behaviour? If so, we need to conditionalise this with fork configuration.

rrw-zilliqa:
> Would a transaction have been able to be mined with the previous behaviour? If so, we need to conditionalise this with fork configuration.

I don't believe so - the old code returned an array of bytes; if you could find an array of bytes which was valid JSON, then potentially yes, but I can't think of one. Happy either way, though - let me know if you want to add it?

JamesHinshelwood:
I would ideally like to test this with the block replay support that I've recently implemented to verify it doesn't break existing transactions before we merge it.

PR #2165, opened 2025-01-16
feat: DEVOPS-1791 rename explorer endpoint to otterscan

frankmeds:


JamesHinshelwood:
Why? I worry this will lead to lots of confusion when explorer.zq2-protomainnet.zilliqa.com disappears.

PR #2168, opened 2025-01-17
Add benchmarks for Scilla and ERC-20 transfers

JamesHinshelwood:
Previously we would generate pre-emptively generate 3.2 million transactions, just in case we ran enough iterations to need them. Instead, we use Criterion's `iter_batched` method to generate the transactions as they are needed. This ensures we don't generate more transactions than we need and that we don't consume loads of memory. The two downsides of this approach are:

* We don't generate transactions in parallel any more, but this doesn't seem to make much difference to the total benchmarking time now that we aren't over-generating transactions.
* The generated flamegraphs now include the transaction generation calls, but drilling down and ignoring them is only one keypress.

I've also renamed the benchmark from `produce-full` to `full-blocks-evm-transfers` since I think it better represents what we are testing.

### DIFF
--- a/zilliqa/benches/it.rs
+++ b/zilliqa/benches/it.rs
@@ -1,16 +1,12 @@
-use std::{
-    cell::{RefCell, RefMut},
-    env,
-    path::PathBuf,
-    sync::Arc,
-    time::Duration,
-};
+use std::{env, iter, path::PathBuf, sync::Arc, time::Duration};
 
 use alloy::{
     consensus::TxLegacy, network::TxSignerSync, primitives::Address, signers::local::LocalSigner,
 };
 use bitvec::{bitarr, order::Msb0};
-use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion, Throughput};
+use criterion::{
+    black_box, criterion_group, criterion_main, BatchSize, Criterion, SamplingMode, Throughput,
+};
 use eth_trie::{MemoryDB, Trie};
 use libp2p::PeerId;
 use pprof::criterion::{Output, PProfProfiler};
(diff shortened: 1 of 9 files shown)


zilliqa/benches/it.rs
JamesHinshelwood, line 1: Will investigate benchmark backfilling and resolve this comment when done. Leaving here to avoid merging :)

86667 (approved):

86667 (dismissed):

JamesHinshelwood:
Sorry @86667 - I had to update the gas limit to ensure the ERC-20 benchmark actually executed the transfers. Could you check again?

zilliqa/benches/it.rs
```diff
@@ -308,20 +465,29 @@ fn a_big_process_vote(big: &mut Consensus, vote: Vote) -> Proposal {
...
     let (_, tiny_vote) = tiny
         .proposal(from, black_box(proposal), false)
         .unwrap()
         .unwrap();
     // We assert 'tiny' actually voted but don't do anything with its vote.
     assert!(matches!(tiny_vote, ExternalMessage::Vote(_)));
+
+    // Get the last transaction receipt and make sure it succeeded. We assume that all other transactions succeeded if
+    // this one did.
+    if let Some(last_txn) = last_txn {
```
86667, line 487: :+1:

86667 (approved):

zilliqa/benches/it.rs
86667: Why do we not need to confirm this in `c_big_process_block` also?

JamesHinshelwood: They process the same block and I didn't want to run this check more than necessary - It isn't what we actually want to benchmark, just a sanity check that our benchmark is working.

PR #2170, opened 2025-01-17
Altered behaviour of GetTransactionStatus

### LINKED ISSUE #2167: Incorrect status returned by `GetTransactionStatus`
The status should be 3 according to https://dev.zilliqa.com/zilliqa2/api/apis/api/zilliqa/GetTransactionStatus/ but the value returned is 1 e.g.:
```
cast rpc GetTransactionStatus 0x7edd27517945f2dbe5725593d85694e6840e42b8959ff6d99ef471c34ec26110 --rpc-url https://api.zq2-prototestnet.zilliqa.com

{"ID":"7edd27517945f2dbe5725593d85694e6840e42b8959ff6d99ef471c34ec26110","_id":null,"amount":"50000000000000","data":"","epochInserted":"10290041","epochUpdated":"10290041","gasLimit":"1200","gasPrice":"2000000016","lastModified":"1737059340905672","modificationState":2,"status":1,"nonce":"49","senderAddr":"0x036beea18d938a51beea396095f78f96f6a8a95d2e8043d7f82ae9576e4792a6af","signature":"0x3fb5586ebfcc9b36c99bb2110494c9919aefa5bcb9b9bcd1f540206d67ddcc9e524e36c0ad19f64643154ff33ec0526079dce8906fa141b99694e3089e1b0cd0","success":true,"toAddr":"0xea207bde83659a939727365282d784075d259f61","version":"21954561"}
```

maxconway:
I'm very open to different logic if anyone knows better exactly what the behaviour should be here.

Intended to fix #2167


maxconway:
See discussion in #2167

zilliqa/src/api/types/zil.rs
```diff
@@ -679,35 +679,35 @@ pub struct TransactionReceiptResponse {
...
-    InsufficientGas = 22,
-    MempoolAlreadyPresent = 23,
-    MempoolSameNonceLowerGas = 24,
-    //
-    InvalidToAccount = 25,
-    FailContractAccountCreation = 26,
-    NonceTooLow = 27,
+    // PresentGasExceeded = 5,
+    // PresentValidConsensusNotReached = 6,
+    // // RareDropped
```
JamesHinshelwood, line 690: I'd be in favour of just deleting these, since we've decided they will never be useful.

zilliqa/src/api/types/zil.rs
```diff
@@ -800,41 +807,55 @@ impl TransactionStatusResponse {
...
-                } else {
-                    TxnStatusCode::Error
-                }
-            };
-        let modification_state = if receipt.accepted.is_none() { 0 } else { 2 };
+        let status_code = if !receipt.errors.is_empty() {
+            TxnStatusCode::Error
+        } else if finalized {
+            TxnStatusCode::Confirmed
+        } else if (block.is_some() && !finalized) || (pending.is_some() && pending.unwrap()) {
```
JamesHinshelwood, line 814: Maybe a nitpick, but could the logic be simplified by passing `state: { Queued, Pending, Finalized }` from the caller instead? Its quite hard to follow what flags lead to different states here.

zilliqa/src/api/zilliqa.rs
```diff
@@ -1494,15 +1494,28 @@ fn get_transaction_status(
...
-        .ok_or(jsonrpsee::types::ErrorObject::owned(
-            RPCErrorCode::RpcDatabaseError as i32,
-            "Block not found".to_string(),
-            jsonrpc_error_data.clone(),
-        ))?;
+    let block = node.get_block(receipt.block_hash)?;
+    let chain_tip_number = node.get_chain_tip();
+
+    let finalized = if let Some(block) = &block {
+        chain_tip_number - block.number() > 2
```
JamesHinshelwood, line 1501: You can't use this to decide finalization. In the presence of missed views/forks, there may be more than 2 blocks between the latest block and the finalized block. `resolve_block_number` has the correct logic.

zilliqa/src/api/types/zil.rs
maxconway: Done

zilliqa/src/api/types/zil.rs
maxconway: Done

zilliqa/src/api/zilliqa.rs

zilliqa/src/api/zilliqa.rs
```diff
@@ -1494,16 +1494,39 @@ fn get_transaction_status(
...
+                if newest_finalized_block.is_some()
+                    && block.number() >= newest_finalized_block.unwrap().number()
+                {
+                    TransactionState::Finalized
+                } else {
+                    TransactionState::Pending
+                }
+            }
+            None => {
+                let mempool = node.consensus.txpool_content()?;
```
JamesHinshelwood, line 1517: I seem to recall `txpool_content` clones a bunch of data and thus isn't particularly fast, especially if the pool is quite large. Can we add/use some kind of `get_transaction_by_hash` method to the pool instead?

zilliqa/src/api/zilliqa.rs
```diff
@@ -1494,16 +1494,39 @@ fn get_transaction_status(
...
-        .get_block(receipt.block_hash)?
-        .ok_or(jsonrpsee::types::ErrorObject::owned(
-            RPCErrorCode::RpcDatabaseError as i32,
-            "Block not found".to_string(),
-            jsonrpc_error_data.clone(),
-        ))?;
 
-    let res = TransactionStatusResponse::new(transaction, receipt, block)?;
-    Ok(res)
+    let block = node.get_block(receipt.block_hash)?;
```
JamesHinshelwood, line 1498: When is the `receipt` not `None` but `block.is_none()`? My guess would be that they are both `None` at the same time (when the transaction is not yet mined)?

maxconway: Good point, I've done a bit or a refactor around this, to deal with missing transaction receipts properly

zilliqa/src/api/zilliqa.rs
maxconway: I've refactored this to possibly look a bit faster by introducing new methods into pool.rs and consensus.rs.

The pool seems to calculate `pending` on the fly and then `queued` is just everything else, so I couldn't really speed up the pending calculation, but I've at least reduced the queued transaction lookup from a vec iteration to a BTreeMap lookup. I don't have an intuition for how many transactions are typically pending or queued at once so I don't know how much difference that'll make

JamesHinshelwood: The pool could contain quite a few transactions (on the order of thousands) in some situations. Doing a linear search for a specific key to work out if it is pending or not seems quite wasteful.

The criteria for whether a transaction is pending or not should be as simple as `is_pending = txn.nonce == state.get_account(txn.from_addr).nonce`.

maxconway: Done, I've implemented that approach

zilliqa/src/pool.rs
```diff
@@ -163,6 +169,23 @@ impl TransactionPool {
         Ok(None)
     }
 
+    /// Returns whether the transaction is pending or queued
+    pub fn get_pending_or_queued(
+        &self,
+        state: &State,
+        txn: &VerifiedTransaction,
+    ) -> Result<Option<PendingOrQueued>> {
+        if txn.tx.nonce().ok_or(anyhow!("Transaction missing nonce"))?
```
JamesHinshelwood, line 178: I think we also need to check if the hash is included in the mempool here. Otherwise all transactions with the right nonce will be 'pending' even if they aren't in the pool at all?

maxconway: As in something like `self.hash_to_index.contains_key(&txn.hash) && txn.tx.nonce().ok_or(anyhow!("Transaction missing nonce"))? == state.get_account(txn.signer)?.nonce`? I'm not sure I get the relationship between a transaction nonce and an account nonce so I'm not sure exactly what the line is doing.

JamesHinshelwood (approved):

zilliqa/src/pool.rs
maxconway: Right, I've pushed a change so that transactions without nonces are never returned as pending, always queued or none

JamesHinshelwood: Hmm just thinking about it now - I think transactions without a nonce should always be **pending**, since they are always ready to be mined?

maxconway: Done :)

JamesHinshelwood (approved):

PR #2171, opened 2025-01-17
fix: Kademlia bootstrap only on startup

86667:
Currently we call Kademlia bootstrap every time there is a `DialFailure`, which can be often even in a healthy network. Instead we should call it only at startup and let libp2p do the rest.  

Also fixes running of local network configs: 
- Since removing mDNS nodes could not communicate with each other without a bootstrap address
- Updated `config_rpc_disabled` to be inline with `config_docker`


JamesHinshelwood (commented):
`docker-compose` should use your local version, not 0.5.3.

With `docker compose up --build`, this should work.

86667:
@JamesHinshelwood thanks. I've changed this PR from being about fixing `docker-compose` (which it turns out already works) to fixing running via the instructions in the README (### Example invocations for a network with 4 nodes)

zilliqa/src/p2p_node.rs
```diff
@@ -234,8 +234,14 @@ impl P2pNode {
...
             if self.swarm.local_peer_id() != peer {
+                // Add bootstrap node's address and call Kademlia bootstrap
+                self.swarm
+                    .behaviour_mut()
+                    .kademlia
+                    .add_address(peer, address.clone());
+                self.swarm.behaviour_mut().kademlia.bootstrap()?;
+
                 self.swarm.dial(address.clone())?;
-                self.swarm.add_peer_address(*peer, address.clone());
```
JamesHinshelwood, line 238: Why don't we add the peer address to the swarm? I'm not sure, but I guess this would also let other behaviours (e.g. gossipsub) learn about the bootstrap peer?

JamesHinshelwood (approved):

zilliqa/src/p2p_node.rs
86667: I removed it because we're explicitly adding to Kademlia above. I suppose youre right it doesnt hurt to do both for the other protocols

86667:
Config changes included here - https://github.com/Zilliqa/zq2/pull/2293

PR #2205, opened 2025-01-21
Implement three trace APIs

### LINKED ISSUE #2117: Implement `trace_filter` RPC method
https://www.quicknode.com/docs/ethereum/trace_filter

### LINKED ISSUE #2192: Implement trace_block

### LINKED ISSUE #2197: Implement trace_transaction

maxconway:
- trace_block
- trace_filter
- trace_transaction

Plus tests and docs

Closes #2117 
Closes #2192 
Closes #2197


zilliqa/src/api/trace.rs
```diff
@@ -109,8 +162,124 @@ fn trace_call_many(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
...
+            }
+        }
+    }
+
+    // Apply pagination if specified
+    let traces = match (filter.after, filter.count) {
+        (Some(after), Some(count)) => all_traces
+            .into_iter()
+            .skip(after as usize)
+            .take(count as usize)
```
JamesHinshelwood, line 275: Isn't it wasteful to only use these parameters after you've already collected `all_traces`? Should we use them above? Or equivalently, make `all_traces` an `Iterator` which you collect the correct number of here?

JamesHinshelwood (dismissed):

zilliqa/src/api/trace.rs
maxconway: Done, good point. I did a procedural approach here and used a loop label here because that was the smaller change from what we already had, but it feels a bit gnarly to me so I'm very open to switching to the iterator approach if you'd prefer?

zilliqa/src/api/trace.rs
```diff
@@ -94,8 +96,59 @@ fn debug_trace_transaction(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<
...
+        // Apply the transaction
+        let result = state.apply_transaction(txn, block.header, &mut inspector, true)?;
+
+        // Build trace results
+        if let TransactionApplyResult::Evm(result, ..) = result {
+            let builder = inspector.into_parity_builder();
+            let trace = builder.into_trace_results_with_state(&result, &trace_types, &pre_state)?;
+            traces.push(trace);
+        }
+        // Skip non-EVM transactions but continue processing the block
```
JamesHinshelwood, line 148: Is this comment still relevant?

zilliqa/src/api/trace.rs
```diff
@@ -139,6 +309,27 @@ fn trace_replay_transaction(params: Params, node: &Arc<Mutex<Node>>) -> Result<T
...
 
 /// trace_transaction
-fn trace_transaction(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
-    todo!("Endpoint not implemented yet");
+fn trace_transaction(params: Params, node: &Arc<Mutex<Node>>) -> Result<Option<TraceResults>> {
+    let mut params = params.sequence();
+    let txn_hash: B256 = params.next()?;
+    let txn_hash: Hash = txn_hash.into();
+
+    // Get the transaction receipt first to check if it exists and was mined
```
JamesHinshelwood, line 317: Is this necessary for tracing? Do other eth clients work this way? (i.e. they don't let you trace unmined transactions)

JamesHinshelwood (dismissed):

zilliqa/src/api/trace.rs
maxconway: Good shout, the same check is already inside trace_evm_transaction

JamesHinshelwood (dismissed):
LGTM

PR #2206, opened 2025-01-21
Fix contract address retrieval for Zilliqa transactions.

saeed-zil:


JamesHinshelwood (commented):
Didn't @bzawisto fix this already?

saeed-zil:
> Didn't @bzawisto fix this already?

I already asked @bzawisto before starting this ticket, and he said no. Maybe he can add more details.

saeed-zil:
@JamesHinshelwood The main branch returned the wrong address when I ran a node with persistence.

zilliqa/src/api/zilliqa.rs
```diff
@@ -376,16 +376,36 @@ fn get_contract_address_from_transaction_id(
...
+            .get_transaction_by_hash(hash)?
+            .ok_or_else(|| anyhow!("Txn Hash not Present"))?;
+        (receipt, signed_transaction)
+    };
 
     let contract_address = receipt
         .contract_address
         .ok_or_else(|| anyhow!("ID is not a contract txn"))?;
 
+    let contract_address = match signed_transaction.tx {
```
JamesHinshelwood, line 394: Please add this as a method to `TxZilliqa` instead. @bzawisto made a similar change in #2190 and there might be more places we need this logic too. I don't think it makes sense to duplicate the logic everywhere.

PR #2221, opened 2025-01-23
Implement `eth_feeHistory` and `eth_maxPriorityFeePerGas` APIs

saeed-zil:
Return values such that user's `maxPriorityFeePerGas` and `maxFeePerGas` are the same as a legacy tx's `gas_price`. This is so that the following translations work:
```
effectiveGasPrice = min(maxFeePerGas, baseFeePerGas + maxPriorityFeePerGas)
totalValidatorReward = effectiveGasPrice * gasUsed
```
Where `baseFeePerGas = 0` since there is no token burn.

So then for `eth_feeHistory` we return values such that client's calculation for Eip1559 fee parameters of 
```
maxFeePerGas = (nextBaseFee * x) + maxPriorityFeePerGas
maxPriorityFeePerGas = median(rewards)
```
Work out as being `gas_price`. 

I have left in the reward percentile calculation code for potential future usage and so that the pattern of data returned is consistent with other EVM blockhchains.


zilliqa/src/api/eth.rs
```diff
@@ -1039,10 +1039,114 @@ fn blob_base_fee(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
...
+    let reward_percentiles: Option<Vec<f64>> = params.optional_next()?;
+    expect_end_of_params(&mut params, 2, 3)?;
+
+    let node = node.lock().unwrap();
+    let newest_block_number = node
+        .resolve_block_number(newest_block)?
+        .ok_or_else(|| anyhow!("block not found"))?
+        .number();
+
+    if newest_block_number < block_count {
```
JamesHinshelwood, line 1102: Should we put some limit on the block count? I think the specification even says this is limited to a range of 1024 blocks?

zilliqa/src/api/eth.rs
```diff
@@ -1039,10 +1039,114 @@ fn blob_base_fee(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
...
+
 /// eth_feeHistory
 /// Returns the collection of historical gas information
-fn fee_history(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
-    Err(anyhow!("API method eth_feeHistory is not implemented yet"))
+fn fee_history(params: Params, node: &Arc<Mutex<Node>>) -> Result<eth::FeeHistory> {
+    let mut params = params.sequence();
+    let block_count: u64 = params.next()?;
+    let newest_block: BlockNumberOrTag = params.next()?;
+    let reward_percentiles: Option<Vec<f64>> = params.optional_next()?;
```
JamesHinshelwood, line 1093:
```suggestion
    let reward_percentiles: Vec<f64> = params.optional_next()?.unwrap_or_default();
```
to save the extra branching later?

zilliqa/src/api/eth.rs
```diff
@@ -1039,10 +1039,114 @@ fn blob_base_fee(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
...
+                    .map(|tx_hash| {
+                        let tx = node
+                            .get_transaction_by_hash(*tx_hash)?
+                            .ok_or_else(|| anyhow!("transaction not found: {}", tx_hash))?;
+                        Ok(tx.tx.gas_price_per_evm_gas())
+                    })
+                    .collect::<Result<Vec<_>>>()?
+                    .into_iter()
+                    .sorted_unstable()
+                    .collect::<Vec<_>>();
```
JamesHinshelwood, line 1126: This makes two allocations. I would instead `collect` into a mutable `Vec`, then just call `Vec::sort_unstable`.

zilliqa/src/api/eth.rs
```diff
@@ -1039,10 +1039,114 @@ fn blob_base_fee(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
     Err(anyhow!("API method eth_blobBaseFee is not implemented yet"))
 }
 
+/// Calculates the gas rewards for a block based on specified percentiles.
+///
+/// # Parameters
+/// - `block`: The block for which the rewards are being calculated.
+/// - `reward_percentiles`: A slice of percentiles (0-100) at which to calculate the rewards.
+/// - `transactions_sorted_gas_used`: A vector of gas used by transactions, sorted in ascending order.
```
JamesHinshelwood, line 1047: This doesn't correspond to the actual parameter?

zilliqa/src/api/eth.rs
```diff
@@ -1039,10 +1039,114 @@ fn blob_base_fee(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
...
+        .iter()
+        .map(|&percentile| percentile / 100.0 * (block_gas_used as f64))
+        .collect();
+
+    let mut rewards = vec![0.0; reward_percentiles.len()];
+    let mut sum_gas = 0.0;
+    let mut percentile_index = 0;
+
+    for gas_used in sorted_gas_prices.iter().map(|&g| g as f64) {
+        sum_gas += gas_used;
```
JamesHinshelwood, line 1066: I don't see how this makes sense. `gas_price` is an amount in ZIL. `gas_used` is an amount of `EvmGas`. I think we are mixing them up here.

I suggest we keep the `EvmGas` newtype for longer to avoid this happening?

zilliqa/src/api/eth.rs
saeed-zil: I believe we should avoid including `reward` in the response if `reward_percentiles` is absent. Doesn't this imply that we need at least one branch?

JamesHinshelwood: You are right, sorry

zilliqa/src/api/types/eth.rs
```diff
@@ -196,6 +198,7 @@ impl Header {
                 .duration_since(SystemTime::UNIX_EPOCH)
                 .unwrap_or_default()
                 .as_secs(),
+            base_fee_per_gas: 0,
```
JamesHinshelwood, line 201: Why is this zero? Shouldn't it be set to the gas price?

zilliqa/src/api/eth.rs
```diff
@@ -1035,8 +1035,95 @@ fn blob_base_fee(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
...
+    };
+
+    let block_count = block_count.min(1024);
+
+    if block_count == 0 {
+        return Ok(FeeHistory::default());
+    }
+
+    let newest_block: BlockNumberOrTag = params.next()?;
+    let reward_percentiles: Option<Vec<f64>> = params.optional_next()?.unwrap_or_default();
```
JamesHinshelwood, line 1054: Did you mean to add `unwrap_or_default` here? I believe this means the API is now expecting a `Option<Option<Vec<f64>>>` which doesn't make much sense.

zilliqa/src/api/eth.rs
```diff
@@ -1035,8 +1035,95 @@ fn blob_base_fee(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
...
+
+            let reward = if let Some(reward_percentiles) = reward_percentiles.as_ref() {
+                let mut effective_priority_fees = block
+                    .transactions
+                    .iter()
+                    .map(|tx_hash| {
+                        let tx = node
+                            .get_transaction_by_hash(*tx_hash)?
+                            .ok_or_else(|| anyhow!("transaction not found: {}", tx_hash))?;
+                        Ok(tx.tx.effective_priority_fee(block.header.base_fee_per_gas))
```
JamesHinshelwood, line 1091: Is it correct to use the priority fee here? This will be zero for all non-EIP-1559 transactions.

JamesHinshelwood: I think not, because if I call `eth_feeHistory` on a Ethereum mainnet block from before EIP-1559, there are non-zero values in `reward`.

zilliqa/src/api/eth.rs
```diff
@@ -1035,8 +1035,95 @@ fn blob_base_fee(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
...
+
+                let fees_len = effective_priority_fees.len() as f64;
+
+                reward_percentiles
+                    .iter()
+                    .map(|x| {
+                        let i = (x * fees_len / 100_f64) as usize;
+                        effective_priority_fees.get(i).cloned().unwrap_or_default()
+                    })
+                    .collect()
```
JamesHinshelwood, line 1105: Some comments here would be nice. It took me a while to work out what this was doing and why it is correct.

JamesHinshelwood (changes requested):

zilliqa/src/api/types/eth.rs
saeed-zil: @DrZoltanFazekas stated, "The base fee is always zero, and the priority fee is the gas price."

zilliqa/src/api/eth.rs
saeed-zil: What if I change `effective_priority_fee` to:

```rust
    pub fn effective_priority_fee(&self, base_fee_per_gas: u128) -> u128 {
        match self {
            SignedTransaction::Eip1559 { tx, .. } => tx
                .max_priority_fee_per_gas
                .min(tx.max_fee_per_gas - base_fee_per_gas),
            _ => self.gas_price_per_evm_gas(),
        }
    }
```

zilliqa/src/api/types/eth.rs
DrZoltanFazekas: Came to the above conclusion by testing on an Ethereum block before EIP-1559.

zilliqa/src/transaction.rs
```diff
@@ -294,6 +294,15 @@ impl SignedTransaction {
         }
     }
 
+    pub fn effective_priority_fee(&self, base_fee_per_gas: u128) -> u128 {
+        match self {
+            SignedTransaction::Eip1559 { tx, .. } => tx
+                .max_priority_fee_per_gas
+                .min(tx.max_fee_per_gas - base_fee_per_gas),
```
JamesHinshelwood, line 301: The alloy implementation is different:
https://docs.rs/alloy-consensus/0.11.1/src/alloy_consensus/transaction/eip1559.rs.html#198

Who is right?

JamesHinshelwood (commented):
A couple more comments, sorry. Also the tests are failing.

zilliqa/src/transaction.rs
```diff
@@ -294,6 +294,15 @@ impl SignedTransaction {
         }
     }
 
+    pub fn effective_priority_fee(&self, base_fee_per_gas: u128) -> u128 {
```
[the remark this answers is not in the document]
saeed-zil: Sure!

zilliqa/src/transaction.rs
saeed-zil: `(tx.minerReward = MIN(tx.maxPriorityFeePerGas, tx.maxFeePerGas-block.baseFee))` is implemention in the document you mentioned not to refer to here :)

JamesHinshelwood: What document?

saeed-zil:
I discovered #2441 while examining the only breaking test. Therefore, I need to resolve it before merging this code.

saeed-zil:
@JamesHinshelwood, I removed the breaking test from this PR to allow for its merging. It will be re-added in #2441.

zilliqa/src/message.rs
```diff
@@ -558,6 +558,7 @@ pub struct BlockHeader {
     pub timestamp: SystemTime,
     pub gas_used: EvmGas,
     pub gas_limit: EvmGas,
+    pub base_fee_per_gas: u128,
```
JamesHinshelwood, line 561: Can we keep this as a fixed constant (the same value as `consensus_config.gas_price`). If we're adding it to the block header we will have a backwards compatibility issue, since we're changing the message format.

JamesHinshelwood (changes requested):
Everything LGTM, except I don't think we need to add a field to `BlockHeader`.

Sorry for the long wait!

zilliqa/src/api/types/eth.rs
DrZoltanFazekas: Here's my test on Ethereum mainnet:

I took a pre-EIP-1559 block with a single transaction that paid 50 gwei gas price: https://etherscan.io/tx/0xefb6c796269c0d1f15fdedb5496fa196eb7fb55b601c0fa527609405519fd581

I called `eth_feeHistory` which returned zero `baseFeePerGas` and the gas price of the transaction as the effective priority fee per gas (aka `reward`):
```
cast rpc eth_feeHistory '["0x1","0xf4241",[100]]' --raw --rpc-url https://eth.llamarpc.com/
{"baseFeePerBlobGas":["0x0","0x0"],"baseFeePerGas":["0x0","0x0"],"blobGasUsedRatio":[0],"gasUsedRatio":[0.006684509000532214],"oldestBlock":"0xf4241","reward":[["0xba43b7400"]]}
```

This led to my conclusion that we should treat/set these values as follows: 
1. base fee = 0
1. priority fee = gas price

@JamesHinshelwood Please prove me wrong :-)

PR #2224, opened 2025-01-24
Fix deposit_v4 audit findings

86667:
- Set contract `VERSION=4`
- Fix Solidity version to `0.8.28`
- Remove double storage lookup of balance in `unstake()`. This saves 494 gas.
- Initialise `indices` in `getStakersData()` so that the function does not error


DrZoltanFazekas (changes requested):
Can you please also remove the two `//TODO` comments?

DrZoltanFazekas (approved):

PR #2225, opened 2025-01-24
Fix crash while removing values from maps containing complex types

bzawisto:
It happened when there was a map containing maps as values, e.g `map<string, map<string, ...>>`. It triggered the assertion and led to node crash.

### DIFF
--- a/zilliqa/src/scilla.rs
+++ b/zilliqa/src/scilla.rs
@@ -931,39 +931,12 @@ impl ActiveCall {
             } else {
                 // Remove multiple elements from storage having the same prefix specified by `indices`
 
-                // Collect all paths of indices up to a value that is not of a map type
-                fn collect_indices(
-                    map: BTreeMap<Vec<u8>, StorageValue>,
-                    path: Vec<Vec<u8>>,
-                    all_paths: &mut Vec<Vec<Vec<u8>>>,
-                ) {
-                    for (key, value) in map {
-                        let mut path = path.clone();
-                        path.push(key);
-                        match value {
-                            StorageValue::Map { map, .. } => {
-                                collect_indices(map, path, all_paths);
-                            }
-                            StorageValue::Value(_) => {
-                                all_paths.push(path.clone());
-                            }
-                        }
-                    }
-                }
-
-                let load_map_from_prefix =
-                    self.state
-                        .load_storage_by_prefix(self.sender, &name, &indices)?;
-
-                let mut all_indices = Vec::new();
-                let path = Vec::new();
-
-                collect_indices(load_map_from_prefix, path, &mut all_indices);
-
-                for path in all_indices {
-                    self.state
-                        .set_storage(self.sender, &name, &path, StorageValue::Value(None))?
-                }
+                let complete_empty_map = StorageValue::Map {
+                    map: BTreeMap::new(),
+                    complete: true,
+                };
+                self.state
+                    .set_storage(self.sender, &name, &indices, complete_empty_map)?;
             }
         } else if indices.len() == depth {
             let Some(ValType::Bval(value)) = value.val_type else {
--- a/zilliqa/src/scilla.rs
+++ b/zilliqa/src/scilla.rs
@@ -931,7 +931,7 @@ impl ActiveCall {
             } else {
                 // Remove multiple elements from storage having the same prefix specified by `indices`
 
-                // Collect all paths of indexes up to a value that is not of a map type
+                // Collect all paths of indices up to a value that is not of a map type
                 fn collect_indices(
                     map: BTreeMap<Vec<u8>, StorageValue>,
                     path: Vec<Vec<u8>>,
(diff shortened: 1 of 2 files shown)


zilliqa/tests/it/zil.rs
```diff
@@ -340,6 +340,12 @@ pub fn scilla_test_contract_code() -> String {
         field welcome_msg : String = "default"
         field welcome_map : Map Uint32 (Map Uint32 String) = Emp Uint32 (Map Uint32 String)
 
+        transition removeHello()
+          delete welcome_map[one][two];
```
JamesHinshelwood, line 344: Don't you want `delete welcome_map[one];` to test the change you made here?

zilliqa/src/scilla.rs
```diff
@@ -923,10 +923,48 @@ impl ActiveCall {
         }
 
         if ignore_value {
-            // We only supporting deleting a single value of a map.
-            assert_eq!(indices.len(), depth);
```
JamesHinshelwood, line 927: This property is relied upon in other places too. Now it doesn't hold, we can't make these assumptions any more and need to implement these methods too:
https://github.com/Zilliqa/zq2/blob/69aa81e4ba4bdf8496a3b89a5e3b1541064e870e/zilliqa/src/exec.rs#L1256
https://github.com/Zilliqa/zq2/blob/69aa81e4ba4bdf8496a3b89a5e3b1541064e870e/zilliqa/src/exec.rs#L1297

zilliqa/src/scilla.rs
```diff
@@ -923,10 +923,48 @@ impl ActiveCall {
...
+
+                let mut all_indices = Vec::new();
+                let path = Vec::new();
+
+                collect_indices(load_map_from_prefix, path, &mut all_indices);
+
+                for path in all_indices {
+                    self.state
+                        .set_storage(self.sender, &name, &path, StorageValue::Value(None))?
+                }
```
JamesHinshelwood, line 966: The database should support deleting all keys with a common prefix more efficiently than this?

JamesHinshelwood (changes requested):

zilliqa/src/scilla.rs
```diff
@@ -923,10 +923,21 @@ impl ActiveCall {
...
+                *storage_slot = None;
+            } else {
+                // Remove multiple elements from storage having the same prefix specified by `indices`
+
+                let complete_empty_map = StorageValue::Map {
+                    map: BTreeMap::new(),
+                    complete: true,
+                };
+                self.state
+                    .set_storage(self.sender, &name, &indices, complete_empty_map)?;
```
JamesHinshelwood, line 939:
```suggestion
                let complete_empty_map = StorageValue::Map {
                    map: BTreeMap::new(),
                    complete: true,
                };
                self.state
                    .set_storage(self.sender, &name, &indices, StorageValue::complete_map())?;
```

zilliqa/tests/it/zil.rs
```diff
@@ -2959,6 +2965,109 @@ async fn get_smart_contract_sub_state(mut network: Network) {
...
+            50_000,
+            None,
+            Some(call),
+        )
+        .await;
+        let event = &txn["receipt"]["event_logs"][0];
+        assert_eq!(event["_eventname"], "removeHello");
+    }
+
+    // Check and confirm the entry does not exist anymore
```
JamesHinshelwood, line 3049: If possible, I'd like to add another assertion here that calls `GetContractState` and asserts that the map now looks like `{ 1 : {} }` (and not `{}`).

JamesHinshelwood (dismissed):
LGTM, thanks. Only 2 small comments.

zilliqa/tests/it/zil.rs
JamesHinshelwood: Oh I guess I assumed that Scilla could support empty maps so if I had `{ 1 : { 2 : "hello" } }`, then these would be the results
* `delete map[one]` -> `{1 : {}}`
* `delete map` -> `{}`

But I guess not?

JamesHinshelwood: Oh I guess I assumed that Scilla could support empty maps so if I had `{ 1 : { 2 : "hello" } }`, then these would be the results
* `delete map[one]` -> `{1 : {}}`
* `delete map` -> `{}`

But I guess not?

JamesHinshelwood: Oh I guess I assumed that Scilla could support empty maps so if I had `{ 1 : { 2 : "hello" } }`, then these would be the results
* `delete map[one]` -> `{1 : {}}`
* `delete map` -> `{}`

But I guess it removes all empty maps?

JamesHinshelwood: Oh I guess I assumed that Scilla could support empty maps so if I had `{ 1 : { 2 : "hello" } }`, then these would be the results
* `delete map[one]` -> `{1 : {}}`
* `delete map` -> `{}`

But I guess it removes all empty maps?

JamesHinshelwood: Confirmed on testnet that this is the correct behaviour - Thanks for explaining and sorry for the noise :)

JamesHinshelwood (approved):

PR #2228, opened 2025-01-24
Alter Zilliqa API to treat latest finalized block as latest block

### LINKED ISSUE #1964: Zilliqa API should query state from the finalized block rather than the latest block
The ZQ1 API made the guarantee that data returned from the Zilliqa API would be consistent. However, ZQ2 does not have single-block finality any more so in order to maintain this guarantee we should return state from the finalized block instead of the latest block. We should also not (yet) return a response for transaction receipts, etc. if they are mined after the finalized block.

maxconway:
Alter Zilliqa API to treat latest finalized block as latest block.
This includes changes so that functions ignore transactions on later blocks

Plus supporting changes:
- convenience functions in node.rs
- many changes to tests
- changes to test utility functions

Closes #1964


zilliqa/src/api/zilliqa.rs
```diff
@@ -506,29 +506,25 @@ fn get_blockchain_info(_: Params, node: &Arc<Mutex<Node>>) -> Result<BlockchainI
...
     let ds_block_rate = tx_block_rate / TX_BLOCKS_PER_DS_BLOCK as f64;
 
     // num_txns_ds_epoch
-    let current_epoch = node.get_chain_tip() / TX_BLOCKS_PER_DS_BLOCK;
+    let current_epoch =
+        node.get_latest_finalized_block_number()?.unwrap_or(0) / TX_BLOCKS_PER_DS_BLOCK;
     let current_epoch_first = current_epoch * TX_BLOCKS_PER_DS_BLOCK;
     let mut num_txns_ds_epoch = 0;
-    for i in current_epoch_first..node.get_chain_tip() {
+    for i in current_epoch_first..node.get_latest_finalized_block_number()?.unwrap_or(0) {
```
JamesHinshelwood, line 519: I don't think it would be wrong to push the `.unwrap_or(0)` into `get_latest_finalized_block(_number)` instead. We can safely make the assumption at that point that the genesis block (at height 0) is always finalized.

JamesHinshelwood (approved):
LGTM - Consider moving the `.unwrap_or(0)` down the call-stack to remove the duplication?

zilliqa/src/api/zilliqa.rs

JamesHinshelwood (approved):

PR #2243, opened 2025-01-29
Issue #2242 load checkpoints as stream

### LINKED ISSUE #2242: Stream decompress checkpoints
Checkpoint decompression currently decompresses a file to `.part` and then loads it and deletes the temporary.
If multiple nodes try to load a checkpoint at the same time, they'll conflict. Given that we read checkpoints serially, we can just as easily read them from the decompressed datastream.

rrw-zilliqa:
Fix #2242

### DIFF
--- a/zilliqa/benches/it.rs
+++ b/zilliqa/benches/it.rs
@@ -68,6 +68,11 @@ fn process_empty(c: &mut Criterion) {
                 consensus.minimum_stake = "1"
                 consensus.eth_block_gas_limit = 1000000000
                 consensus.gas_price = "1"
+                consensus.genesis_fork.at_height = 0
+                consensus.genesis_fork.failed_scilla_call_from_gas_exempt_caller_causes_revert = true
+                consensus.genesis_fork.call_mode_1_sets_caller_to_parent_caller = true
+                consensus.genesis_fork.scilla_messages_can_call_evm_contracts = true
+                consensus.genesis_fork.scilla_contract_creation_increments_account_balance = true
                 consensus.genesis_accounts = [
                     [
                         "0x0000000000000000000000000000000000000000",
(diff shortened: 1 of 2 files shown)


zilliqa/benches/it.rs
```diff
@@ -66,6 +66,11 @@ fn process_empty(c: &mut Criterion) {
                 consensus.minimum_stake = "1"
                 consensus.eth_block_gas_limit = 1000000000
                 consensus.gas_price = "1"
+                consensus.genesis_fork.at_height = 0
+                consensus.genesis_fork.failed_scilla_call_from_gas_exempt_caller_causes_revert = true
+                consensus.genesis_fork.call_mode_1_sets_caller_to_parent_caller = true
+                consensus.genesis_fork.scilla_messages_can_call_evm_contracts = true
+                consensus.genesis_fork.scilla_contract_creation_increments_account_balance = true
```
JamesHinshelwood, line 73: Sorry - Please can we merge https://github.com/Zilliqa/zq2/pull/2244 instead? These were meant to have defaults.

zilliqa/src/db.rs
```diff
@@ -361,13 +361,8 @@ impl Db {
...
         // Decompress file and write to temp file
         let input_filename = path.as_ref();
-        let temp_filename = input_filename.with_extension("part");
-        decompress_file(input_filename, &temp_filename)?;
-
-        // Read decompressed file
-        let input = File::open(&temp_filename)?;
-
-        let mut reader = BufReader::with_capacity(128 * 1024 * 1024, input); // 128 MiB read chunks
+        let buf_reader: BufReader<File> = BufReader::new(File::open(input_filename)?);
```
JamesHinshelwood, line 364: I recall @bzawisto deliberately changing the buffer size to 128 MiB at some point in the past. Any reason not to keep it?
```suggestion
        let buf_reader = BufReader::with_capacity(128 * 1024 * 1024, File::open(input_filename)?); // 128 MiB read chunks
```

JamesHinshelwood (commented):
Please can we rebase once #2244 is merged? Sorry for letting main break

zilliqa/benches/it.rs
rrw-zilliqa: I've removed these now 2244 is merged.

zilliqa/src/db.rs
rrw-zilliqa: No, though I don't think it's likely to particularly change anything any longer - a big BufReader probably helps when writing the result to disk but I'm not sure it helps much when reading, since the buffer cache will look ahead in any event. I did try to measure it, but restoring a mainnet checkpoint takes so long I gave up ... harmless, anyway, so done :-)

JamesHinshelwood (approved):

PR #2281, opened 2025-02-03
fix: create NewView on startup and error log proposal failures

86667:


JamesHinshelwood (approved):

zilliqa/src/consensus.rs
```diff
@@ -417,6 +417,9 @@ impl Consensus {
             peers.add_peers(recent_peer_ids);
         }
 
+        // Build NewView so that we can immediately contribute to consensus moving along if it has halted
+        consensus.build_new_view()?;
```
DrZoltanFazekas, line 421: Why is it correct to send a `NewView` when we restart a validator node??

DrZoltanFazekas (changes requested):
I'm not convinced we should send a `NewView` on startup - let's discuss here or on a call before merging this...

zilliqa/src/consensus.rs
86667: This doesnt send the `NewView`, it only adds it to the `new_view_message_cache`. 

In the case of consensus moving along then nothing happens, it gets cleared or overwritten with a new `NewView` in the future. If however consensus in the network is stalled then after `consensus_timeout` (5s) has passed the node will send this `NewView` in the re-send logic.

Without this change a node will only ever send a `NewView` if it has timed-out and incremented its own view. This means that in scenarios where there is a long timeout we have to wait for a supermajority to be back online and for them all to timeout once.

DrZoltanFazekas: 5 sec afer the node started it resends the `NewView` only if it had sent the same `NewView` before it went down? It won't send a` NewView` for sure in case it had sent a `Vote` in the same view before it went down?

86667: That's true, it is possible for a node to `Vote` on a view with a long timeout, go down, come back up, wait 5s but still be inside the exp back-off timeout, then send a `NewView` for the same view as the one it Voted for.

Ive edited the code so that the `NewView` is only built in scenarios when a view jump has occurred. This way the node can help move consensus if it has not moved after 5 seconds but could not have already made any actions for this view because it went down with a view number less than the current.

86667: If this is too complex and not worth it then I'm more than happy to remove. It is useful only in extreme downtime scenarios only

DrZoltanFazekas (approved):

PR #2283, opened 2025-02-03
Correct log index returns in etherium apis

### LINKED ISSUE #2188: Receipts contain incorrect log indices
For instance the proto-testnet block `0x9b08fd0e1278a9a15fcc645fed8ff5c6cdc9cf9440095b38bd92a97ee0bb4735`  imported from ZQ1 contains 2 transactions. The first transaction has 2 logs i.e. their log indices are 0x0 and 0x1 so the second transaction's log index must be 0x2 which is the case on ZQ1
```
cast receipt 0x02a2c597dc4039d6bc0dcd63cd85b496da891c74125281393b488f2354d00c8a --json --rpc-url https://dev-api.zilliqa.com | jq '.logs.[].logIndex'

"0x2"
```
but not on ZQ2 where it is 0x0
```
cast receipt 0x02a2c597dc4039d6bc0dcd63cd85b496da891c74125281393b488f2354d00c8a --json --rpc-url https://api.zq2-prototestnet.zilliqa.com | jq '.logs.[].logIndex'

"0x0"

```

It might be an API issue or a persistence conversion issue

maxconway:
Fixes log index returns

Logs need to be indexed starting at 0 at the start of each block (versus previously they started at 0 for each receipt)

This commit does it all on the fly, counting through blocks as needed at the point of returning from the APIs, rather than by adding to the log structures. This keeps the logs themselves cleaner but at the cost of some messy code and some amount of performance penalty.

Closes #2188


zilliqa/src/api/eth.rs
```diff
@@ -946,7 +934,19 @@ async fn subscribe(
                 if !filter.filter_block_hash(receipt.block_hash.into()) {
                     continue;
                 }
-                for (log_index, log) in receipt.logs.into_iter().enumerate() {
+
+                // We track log index plus one because we have to increment before we use the log index, and log indexes are 0-based.
```
JamesHinshelwood, line 938: Are we not able to pass the fully correct log index in the subscription rather than recalculating it here?

My only worry is that this will be called for each log that is mined if anyone has an open subscription, even without any API interactions, so performance might be more of a consideration here.

JamesHinshelwood (commented):
LGTM - Just one comment in the subscription implementation that I would like to discuss.

zilliqa/src/api/eth.rs
maxconway: Yep so I think if we want to do that, the closest place to calculate it would be in broadcast_commit_receipts(). Because we need the whole block to calculate log indices, subscribe_to_receipts() gets its receipts from consensus.new_receipts, and that gets them from broadcast_commit_receipts(). But broadcast_commit_receipts() is quite small and is called from inside execute_block(), which also creates the receipts and adds them to the db, so at that point I'd be tempted to just calculate the correct log indices at that point and have it done all once and for all, if you're up for that?

JamesHinshelwood (approved):

zilliqa/src/api/eth.rs
maxconway: I've created a new issue to track this required change: #2473

PR #2284, opened 2025-02-03
Do not execute txs on ZQ1 blocks

86667:
Set ZQ1 block state root hash to zero when converting persistence. For proto networks with incorrect state root hashes  we now set the state root hash to zero upon fetching it.

With there now being blocks without any State in the db we return an error when an attempt to execute transactions on these blocks is made via an API call.

### DIFF
--- a/zilliqa/src/cfg.rs
+++ b/zilliqa/src/cfg.rs
@@ -209,6 +209,8 @@ impl Default for NodeConfig {
             failed_request_sleep_duration: failed_request_sleep_duration_default(),
             enable_ots_indices: false,
             max_rpc_response_size: max_rpc_response_size_default(),
+            prune_interval: u64_max(),
+            proto_network_persistence_block_height: None,
         }
     }
 }
--- a/zilliqa/src/cfg.rs
+++ b/zilliqa/src/cfg.rs
@@ -1,6 +1,6 @@
 use std::{ops::Deref, str::FromStr, time::Duration};
 
-use alloy::primitives::Address;
+use alloy::{primitives::Address, rlp::Encodable};
 use anyhow::{Result, anyhow};
 use libp2p::{Multiaddr, PeerId};
 use rand::{Rng, distributions::Alphanumeric};
@@ -209,8 +209,6 @@ impl Default for NodeConfig {
             failed_request_sleep_duration: failed_request_sleep_duration_default(),
             enable_ots_indices: false,
             max_rpc_response_size: max_rpc_response_size_default(),
-            prune_interval: u64_max(),
-            proto_network_persistence_block_height: None,
         }
     }
 }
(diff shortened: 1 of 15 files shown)


86667:
> So I think we've still got some functions in the API, e.g. `impl TxBlock` in api/types/zil.rs, which return state root hashes by calling `block.state_root_hash()`, which gets a state root hash directly from the block header. Will this address that too or should we also add some calls to `is_zq1_block` there to zero the hashes if appropriate?

the code has changed as discussed in a meeting since this comment was made but its still worth answering to confirm - yes these changes will change all ZQ1 blocks to have state root hashes of zeros.

JamesHinshelwood:
@86667 Note that the final ZQ1 block shouldn't have a state root hash of zero since that's the only block where we do actually know the state.

zilliqa/src/cfg.rs
```diff
@@ -159,6 +159,10 @@ pub struct NodeConfig {
     /// Maximum allowed RPC response size
     #[serde(default = "max_rpc_response_size_default")]
     pub max_rpc_response_size: u32,
+    /// The block height at which ZQ1 converted persistence ends. After this block ZQ2 blocks begin.
+    /// This value should be required only for proto networks to distinguise between ZQ1 and ZQ2 blocks. In future converted networks all ZQ1 blocks will be distinguishable by their zeroed state root hash.
```
JamesHinshelwood, line 163: > In future converted networks all ZQ1 blocks will be distinguishable by their zeroed state root hash

I would caution against this comment for two reasons:

1. When pruning is implemented ZQ2 blocks might have some kind of 'zero' state hash to indicate we haven't stored their state.
2. Eventually ZQ1 blocks won't have a zero state root hash.

I don't want people to rely upon the property that `block.is_from_zq1() <=> block.state_root_hash.is_zero()`

zilliqa/src/cfg.rs
```diff
@@ -159,6 +159,10 @@ pub struct NodeConfig {
     /// Maximum allowed RPC response size
     #[serde(default = "max_rpc_response_size_default")]
     pub max_rpc_response_size: u32,
+    /// The block height at which ZQ1 converted persistence ends. After this block ZQ2 blocks begin.
+    /// This value should be required only for proto networks to distinguise between ZQ1 and ZQ2 blocks. In future converted networks all ZQ1 blocks will be distinguishable by their zeroed state root hash.  
+    #[serde(default)]
+    pub proto_network_persistence_block_height: Option<u64>,
```
JamesHinshelwood, line 165: Could it make sense to represent this property as a fork instead? i.e. we insert a fork at the first ZQ2 block with a new property that says "this block is executable" and all blocks before it aren't?

zilliqa/src/cfg.rs
86667: Okay understood thanks

zilliqa/src/cfg.rs
86667: Done. That's cleaner nice suggestion

maxconway (approved):
LGTM

maxconway (approved):

PR #2285, opened 2025-02-03
Fix local docker compose runs

### LINKED ISSUE #2114: Support multiple bootstrap nodes
The job of bootstrap nodes aka boot nodes is to facilitate peer discovery. It shall be possible to specify all bootstrap nodes with a singe DNS name instead of their IP addresses. However, bootstrap nodes like every other node in the p2p protocol need a unique peer id.

JamesHinshelwood:
This adds support for multiple bootstrap peers. We use this in the local docker compose set up to ensure all nodes have at least one other node to talk to at startup, which they can use to find their own external address via autonat.

This also means we can remove the incorrect code where we always trust our observed external address once a single peer reports one. Instead, the external address candidates are passed to autonat and confirmed later.

The configuration change is backwards compatible, because we support either a single entry or a list of bootstrap peers now.

Fixes #2114.

### DIFF
--- a/zilliqa/src/p2p_node.rs
+++ b/zilliqa/src/p2p_node.rs
@@ -236,7 +236,7 @@ impl P2pNode {
             self.swarm.add_external_address(external_address.clone());
         }
 
-        if let Some((peer, address)) = &self.config.bootstrap_address {
+        for (peer, address) in &self.config.bootstrap_address.0 {
             if self.swarm.local_peer_id() != peer {
                 self.swarm.dial(address.clone())?;
                 self.swarm.add_peer_address(*peer, address.clone());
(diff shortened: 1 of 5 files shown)


frankmeds (approved):

shawn-zil:
@JamesHinshelwood i think one side-effect of this is that, while local docker is able to run (because each node is now connected to the 2 bootstraps) syncing is now only syncing against 2 nodes, not 4 - when I spin up a new node.

zilliqa/src/p2p_node.rs
```diff
@@ -254,20 +254,6 @@ impl P2pNode {
                         SwarmEvent::NewListenAddr { address, .. } => {
                             info!(%address, "P2P swarm listening on");
                         }
-                        SwarmEvent::Behaviour(BehaviourEvent::Identify(identify::Event::Received { peer_id, info: identify::Info{ listen_addrs, observed_addr, protocols, .. }, .. })) => {
```
shawn-zil, line 257: I believe that this is actually needed - https://docs.rs/libp2p-kad/latest/libp2p_kad/#important-discrepancies
> Peer Discovery with Identify In other libp2p implementations, the [Identify](https://github.com/libp2p/specs/tree/master/identify) protocol might be seen as a core protocol. Rust-libp2p tries to stay as generic as possible, and does not make this assumption. This means that the Identify protocol must be manually hooked up to Kademlia through calls to [Behaviour::add_address](https://docs.rs/libp2p-kad/latest/libp2p_kad/struct.Behaviour.html#method.add_address). If you choose not to use the Identify protocol, and do not provide an alternative peer discovery mechanism, a Kademlia node will not discover nodes beyond the network’s [boot nodes](https://docs.libp2p.io/concepts/glossary/#boot-node). Without the Identify protocol, existing nodes in the kademlia network cannot obtain the listen addresses of nodes querying them, and thus will not be able to add them to their routing table.

shawn-zil: This might explain why I'm seeing only syncing against 2 nodes on docker, now.

PR #2293, opened 2025-02-05
Config toml updates

86667:
- Write newest config builds for each network
- Updated `config_rpc_disabled` to be inline with `config_docker` and have a bootstrap address 
- Remove `otlp_collector_endpoint` from configs. Allow adding via z2 join


z2/resources/chain-specs/zq2-devnet.toml
```diff
@@ -1,6 +1,8 @@
 p2p_port = 3333
 bootstrap_address = [ "12D3KooWFvZum3gpEQkj6o9Tg71uLuQZJ4hqXL32c8wPTcpkRnaP", "/dns/bootstrap.zq2-devnet.zilliqa.com/tcp/3333" ]
 
+otlp_collector_endpoint = "http://localhost:4317"
```
JamesHinshelwood, line 4: I don't think we want this. Its up to validators if they want to collect metrics. We shouldn't assume they will (nor that their collector will be running on port 4317).

@pavlops @frankmeds Thoughts?

frankmeds: yes, let's add this as z2 join param...... if this is required in the config.toml

86667: Sure I can do that now

JamesHinshelwood:
Superseded by #2295

PR #2299, opened 2025-02-06
Validators ignore some sync requests

shawn-zil:
With a configurable chance for validators to ignore sync requests, rather than hard-coded 100% complete blocking, there is chance that some leakage will happen - needed to ensure blocks don't end up being trapped within validators.

There is also a question of whether this feature is still needed, due to the new sync operation. However, having this as a configuration, allows the node operator to turn it on/off as an option. It defaults to servicing 100% requests.


JamesHinshelwood (commented):
I'd be reluctant to do this unless we see that serving sync requests on validators is causing concrete performance problems. And even if that's the case, my first wish would be for us to make serving those sync requests faster rather than adding this extra source of non-determinism to our networks.

shawn-zil:
OK. Closing unless needed in the future.

PR #2300, opened 2025-02-06
Added global and per sender limits to txns pool

bzawisto:
Two new parameters have been added:
- max number of txns pool can hold
- max number of pending txns kept per sender
- max number of slots for all senders


rrw-zilliqa:
Does this not lead to `sender_txn_counter` tending to occupy O(number of accounts) ? Probably not disastrous, but maybe worth avoiding ...

z2/resources/chain-specs/zq2-devnet.toml
```diff
@@ -51,3 +51,7 @@ consensus.gas_price = "4_761_904_800_000"
 consensus.contract_upgrades = { deposit_v5 = { height = 0, reinitialise_params = { withdrawal_period = 300 } } }
 
 api_servers = [{ port = 4201, enabled_apis = [{ namespace = "eth", apis = ["blockNumber"] }] }, { port = 4202, enabled_apis = ["admin", "debug", "erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa"] }]
+
+txn_pool.maximum_global_size = 100000
+txn_pool.maximum_txn_count_per_sender = 200
+txn_pool.total_slots_for_all_senders = 50000
```
JamesHinshelwood, line 57: Why do we need to include these in any config files if we're just setting them to the default values anyway?

zilliqa/src/cfg.rs
```diff
@@ -517,6 +520,39 @@ impl Default for ConsensusConfig {
...
 }
 
+#[derive(Clone, Serialize, Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct TxnPoolConfig {
+    /// Maximum number of transactions transaction pool can hold
+    pub maximum_global_size: u64,
+    /// Maximum number of transactions per single sender
+    pub maximum_txn_count_per_sender: u64,
+    /// total slots for all senders
```
JamesHinshelwood, line 530: What is a 'slot'?

zilliqa/src/pool.rs
```diff
@@ -313,6 +331,35 @@ impl TransactionPool {
...
+            // Check per sender counter
+            let sender_counter = self
+                .sender_txn_counter
+                .get(&txn.signer)
+                .copied()
+                .unwrap_or_default();
+            if sender_counter + 1 > self.config.maximum_txn_count_per_sender {
+                return ValidationFailed(ValidationOutcome::TransactionCountExceededForSender);
+            }
         }
```
JamesHinshelwood, line 363: Can we just add this code to the `else` block above and remove `is_replacement`?

zilliqa/src/cfg.rs
JamesHinshelwood: From the impl it looks like this is a limit on the number of unique senders in the pool.

Why is that necessary when we already have the `maximum_global_size`?

JamesHinshelwood (approved):

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):

PR #2301, opened 2025-02-06
feat: DEVOPS-1860 support multi bootstrap

frankmeds:


JamesHinshelwood:
Do we also need to update `config.tera.toml`?

frankmeds:
@JamesHinshelwood this change is related to the infrastructure and does not bring any impact on z2. Will follow another PR related only the z2 changes

PR #2305, opened 2025-02-06
Add z2 commands for depositTopUp, Unstake and Withdraw

86667:


z2/src/validators.rs
```diff
@@ -46,31 +46,16 @@ impl Validator {
 }
 
 #[derive(Debug)]
-pub struct StakeDeposit {
-    validator: Validator,
-    amount: u8,
+pub struct ClientConfig {
```
86667, line 49: Ugly diff for simple change: created a `ClientConfig` and moved its members from  `StakeDeposit`

frankmeds (approved):

frankmeds (approved):

frankmeds (approved):

PR #2307, opened 2025-02-07
Missing retry turnaround

shawn-zil:
While a node can still theoretically sync up without this fix, given enough time, but it will be rather annoying to have to re-download all the pre-existing blocks because it missed the turnaround.

FWIW, I think that we should merge this and fix it into 0.6.1.


86667 (approved):
Makes sense :+1:

PR #2316, opened 2025-02-10
Speed up sync

shawn-zil:
Speeds up Phase 1, by firing 10 concurrent requests at a time, processing 10 responses, then firing the next 10, etc. The speedup in this sense is fairly linear (at least up to 10), as expected. Since Phase 2 is CPU-bound, firing concurrent requests is unnecessary.

When processing the Phase 1 responses, it also splits the chain segment into sub-segments, based on the estimated response payload size. So, if there are large blocks, it will split the segment into several sub-segments for Phase 2 processing.

This removes support for V1 compatibility. **Do not merge this until the hard-fork to >= 0.6.0**


zilliqa/src/message.rs
```diff
@@ -550,6 +550,8 @@ pub struct BlockHeader {
     pub timestamp: SystemTime,
     pub gas_used: EvmGas,
     pub gas_limit: EvmGas,
+    /// Used for syncing only, defaults to None
+    pub sync_size_estimate: Option<usize>,
```
86667, line 554: Bit of a nit: This being in the type which we use in the db and within logic elsewhere feels messy. Im wondering if its worth using a different `SyncBlockHeader` type which is a `BlockHeader` along with this  `sync_size_estimate` value, and only writing the `BlockHeader` as is required for consesnsus

86667 (commented):
Looks good. The whole V1/V2 situation isnt super clear to me, I would find it useful to have a quick explainer in your docs at the top of the file if you wouldnt mind

zilliqa/src/message.rs
shawn-zil: This does make sense. Let me try to do that instead.

86667 (approved):

PR #2345, opened 2025-02-13
Deploy essential contracts during after deploying new network

bzawisto:
As title says - we want to deploy known transactions after installing a new network (excluding mainnet and testnet).

### DIFF
--- a/z2/src/deployer.rs
+++ b/z2/src/deployer.rs
@@ -12,7 +12,7 @@ use ethers::{
     middleware::Middleware,
     prelude::{Bytes, TransactionRequest},
 };
-use primitive_types::{H160, U256};
+use primitive_types::{H160, H256, U256};
 use strum::Display;
 use tokio::{fs, sync::Semaphore, task};
 use zilliqa::{crypto::SecretKey, exec::BLESSED_TRANSACTIONS};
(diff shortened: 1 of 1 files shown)


z2/src/deployer.rs
```diff
@@ -89,10 +89,19 @@ pub async fn install_or_upgrade(
...
+        .cloned()
+        .ok_or(anyhow!("No chain nodes found!"))?;
+
     let _ = execute_install_or_upgrade(bootstrap_nodes, is_upgrade, max_parallel).await;
-    let _ = execute_install_or_upgrade(chain_nodes, is_upgrade, max_parallel).await;
+    let _ = execute_install_or_upgrade(chain_nodes.clone(), is_upgrade, max_parallel).await;
     let _ = execute_install_or_upgrade(apps_nodes, is_upgrade, max_parallel).await;
 
+    if !is_upgrade {
+        post_install(chain_node).await?;
```
frankmeds, line 102: if I run the install twice, will this fails?

z2/src/deployer.rs
```diff
@@ -144,6 +153,10 @@ async fn execute_install_or_upgrade(
     Ok(())
 }
 
+async fn post_install(node: ChainNode) -> Result<()> {
```
frankmeds, line 156: is there any reason not to call `node.post_install().await` directly in the change above?

z2/src/chain/node.rs
```diff
@@ -1415,6 +1418,46 @@ impl ChainNode {
 
         Ok(())
     }
+
+    pub async fn post_install(&self) -> Result<()> {
+        if self.chain.name().contains("prototestnet") || self.chain.name().contains("protomainnet")
+        {
+            log::info!("Skipping post install actions for chain: {}", self.name());
```
frankmeds, line 1425: self.name is the node name, not the chain name

z2/src/chain/node.rs
```diff
@@ -1415,6 +1418,46 @@ impl ChainNode {
 
         Ok(())
     }
+
+    pub async fn post_install(&self) -> Result<()> {
```
frankmeds, line 1422: this method looks more belonging to the chain rather than a node

z2/src/deployer.rs
frankmeds: is it fine then funding n times?

frankmeds (approved):

PR #2369, opened 2025-02-18
Hardfork at height in which serde_json started to preserve the ordering of deseralised data

86667:


zilliqa/src/cfg.rs
```diff
@@ -518,6 +519,8 @@ pub struct ForkDelta {
     /// the contract balance will be sum of the existing balance and the amount sent in the deployment transaction.
     /// If false, the contract balance will be the amount sent in the deployment transaction.
     pub scilla_contract_creation_increments_account_balance: Option<bool>,
+    // If true then the compiled code is using serde_json with feature "preserve_order"
+    pub serde_json_preserve_ordering: Option<bool>,
```
JamesHinshelwood, line 523: nit: I feel the feature should be named something like `init_data_preserve_order` instead. The user/configuration shouldn't need to know the implementation detail that we happen to use `serde_json` internally.

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):
Please verify this is the correct fork height by syncing a node across this point, as we discussed. Thanks!

PR #2372, opened 2025-02-19
Add custom genesis block number

bzawisto:
Converter doesn't skip blocks with numbers 0 and 1.


JamesHinshelwood:
Is this still needed now we have passive syncing @shawn-zil ?

shawn-zil:
> Is this still needed now we have passive syncing @shawn-zil ?

If the requirement is for nodes to start from a checkpoint, then we can rely on passive-sync to sync to any height. @JamesHinshelwood

PR #2373, opened 2025-02-19
Restart Node when Stuck Syncing

shawn-zil:
It is difficult to replicate the issue of a node getting stuck. However there are a couple of observations that:
1. that block is at a fork;
2. the stuck node is the Leader of that block;
3. restarting a stuck node gets past the issue; and
4. given enough time, the stuck node runs phase 2 all the way up to present, then restarts syncing; also gets past the issue.

So, this PR does some things to mitigate the issue.
1. It does not allow a syncing node to broadcast any Proposals i.e. when a Leader falls out of sync, it only syncs. I have a hunch that early proposals might be involved but I cannot prove it.
2. It restarts the Node when stuck syncing - effectively destroys/re-constructs its thread. Since this is only employed during a stuck sync, it is less drastic as the node is already stuck. It determines whether it is stuck at syncing by checking that phase 2 injections have not made progress for > 1000 blocks. It doesn't get triggered if the node is stuck for some other reason e.g. a long timeout.

This PR does some other minor things:
1. Return hex-encoded blocks/numbers in the RPC responses.
2. Not confusing start_at with checkpoint_at.
4. Minor cleanup.

### DIFF
--- a/zilliqa/src/api/types/eth.rs
+++ b/zilliqa/src/api/types/eth.rs
@@ -493,7 +493,6 @@ pub struct SyncingStruct {
 
 #[derive(Clone, Serialize)]
 #[serde(untagged)]
-#[allow(clippy::large_enum_variant)]
 pub enum SyncingResult {
     Bool(bool),
     Struct(SyncingStruct),
--- a/zilliqa/src/api/types/eth.rs
+++ b/zilliqa/src/api/types/eth.rs
@@ -463,21 +463,31 @@ pub struct TxPoolContent {
 #[serde(rename_all = "camelCase")]
 pub struct SyncingMeta {
     pub current_phase: String,
-    pub peer_count: String,
-    pub header_downloads: String,
-    pub block_downloads: String,
-    pub buffered_blocks: String,
-    pub empty_count: String,
-    pub retry_count: String,
-    pub timeout_count: String,
+    #[serde(serialize_with = "hex")]
+    pub peer_count: usize,
+    #[serde(serialize_with = "hex")]
+    pub header_downloads: usize,
+    #[serde(serialize_with = "hex")]
+    pub block_downloads: usize,
+    #[serde(serialize_with = "hex")]
+    pub buffered_blocks: usize,
+    #[serde(serialize_with = "hex")]
+    pub empty_count: usize,
+    #[serde(serialize_with = "hex")]
+    pub retry_count: usize,
+    #[serde(serialize_with = "hex")]
+    pub timeout_count: usize,
 }
 
 #[derive(Clone, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SyncingStruct {
-    pub starting_block: String,
-    pub current_block: String,
-    pub highest_block: String,
+    #[serde(serialize_with = "hex")]
+    pub starting_block: u64,
+    #[serde(serialize_with = "hex")]
+    pub current_block: u64,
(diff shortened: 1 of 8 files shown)


frankmeds:
> @frankmeds I've changed the output of the `eth_syncing` API to output the values using hex-strings instead. Not sure if this will impact your work.

can you add also the change in
https://github.com/Zilliqa/zq2/blob/main/z2/resources/node_provision.tera.py#L133-L134
 to convert from hex to decimal?

86667 (dismissed):

frankmeds (dismissed):

zilliqa/src/api/types/eth.rs
```diff
@@ -463,26 +463,27 @@ pub struct TxPoolContent {
...
-    pub empty_count: u64,
-    pub retry_count: u64,
-    pub timeout_count: u64,
+    pub peer_count: String,
+    pub header_downloads: String,
+    pub block_downloads: String,
+    pub buffered_blocks: String,
+    pub empty_count: String,
+    pub retry_count: String,
+    pub timeout_count: String,
```
JamesHinshelwood, line 472: Please use `#[serde` annotations like we do in the other APIs rather than converting manually.

zilliqa/src/sync.rs
```diff
@@ -303,9 +307,26 @@ impl Sync {
...
                     .get_highest_canonical_block_number()?
-                    .expect("no highest block"),
+                    .expect("no highest canonical block"),
             )?
-            .expect("missing highest block");
+            .expect("missing canonical block");
+
+        if let Some(finalized_view) = self.db.get_finalized_view()? {
+            // Cleanup tip of chain - https://github.com/Zilliqa/zq2/issues/2354
+            while let Some(head_block) = self.db.get_highest_recorded_block()? {
```
JamesHinshelwood, line 316: Don't we already have code that does this at the top of `consensus.rs`?

https://github.com/Zilliqa/zq2/blob/4b2a99ebcf9950ff8403d8ea2a588eee25281b68/zilliqa/src/consensus.rs#L276-L297

Why do both of these need to exist?

JamesHinshelwood (changes requested):

zilliqa/src/sync.rs
shawn-zil: Yes, this is based on that one, which runs at startup while this one runs every time syncing runs.

Not sure if anything else might need it too, outside of sync.

shawn-zil: I think that I'll remove this instead. And if we reallly want to opt for a restart, I'll implement a complete app-level restart mechanism that I did earlier instead.

JamesHinshelwood: I don't think doing the cleanup here is necessarily bad. I just don't like duplicating the logic in two places.

When is this called? Maybe if its called near startup we can remove the erasing from `consensus.rs` and just rely on this?

shawn-zil: The reason that I removed the cleanup and fast retry is that it didn't seem to help with Zoltan's node in `prototestnet`.

I'm guessing that there might be some state that didn't get cleared/reset in consensus (my hunch is due to the speculative block assembly stuff, or execution bypass stuff). So, merely clearing and doing a fast retry in sync, might encounter the same state and same issues. 

A complete restart (freeing up memory) might do better. So, I opted to do that instead.

Issue #2344, opened 2025-02-12
Syncing prototestnet node stuck on Scilla deployment


DrZoltanFazekas:
Started syncing a new node from an old checkpoint `7689600`. It was already downloading blocks but got stuck at `8190043`, which is a block that contains this Scilla contract deployment: https://explorer.zq2-prototestnet.zilliqa.com/tx/0x3fcb6f9fe73bc99aea21fef1dc8ec1628f58706db8c704173510a93141eac611

86667:
The issue is the adding of `preserve_order` feature to `serde_json` in _**z2**_ - https://github.com/Zilliqa/zq2/pull/2295/files#diff-402c6c70c07193aa9746960085fc1b514e6306217f6077a1ab86e1f294e6058e

The particular line is in `exec.rs`:
```rust
 let mut init_data: Vec<ParamValue> = serde_json::from_str(&txn.data)?;
```

Which results in a different ordering of `init_data`. 

## TODO 
- Implement custom deserialisation for struct `ParamValue` 
- Activate via hardfork - should use behaviour without order preservation up to a block before `v0.6.0` upgrade then begin using order preserving format after

86667:
Second PR for this issue - https://github.com/Zilliqa/zq2/pull/2374

86667:
Another PR correcting the hardfork height - https://github.com/Zilliqa/zq2/pull/2418

PR #2374, opened 2025-02-19
fix: Scilla calls json ordering

86667:
Follow on PR fixing https://github.com/Zilliqa/zq2/issues/2344


JamesHinshelwood (dismissed):

JamesHinshelwood:
We should rename the fork flag since its not just `init_data` any more. Maybe just `scilla_json_preserve_order` or similar?

JamesHinshelwood (approved):

PR #2375, opened 2025-02-19
feat: set withdrawalPeriod at initialise of deposit_v5 and add deposit contract tests

86667:
Adds `deposit_v5` with `withdrawalPeriod` configurable via `reinitialize()`.

Deposit tests are a work-in-progress but largely complete.

### DIFF
--- a/zilliqa/src/contracts/deposit_v5.sol
+++ b/zilliqa/src/contracts/deposit_v5.sol
@@ -375,6 +375,10 @@ contract Deposit is UUPSUpgradeable {
         bytes calldata blsPubKey,
         address signingAddress
     ) public onlyControlAddress(blsPubKey) {
+        require(
+            signingAddress != address(0),
+            "signingAddress cannot be set to zero address"
+        );
         DepositStorage storage $ = _getDepositStorage();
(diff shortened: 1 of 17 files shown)


zilliqa/src/contracts/deposit_v5.sol
```diff
@@ -0,0 +1,719 @@
...
+        $._stakersMap[blsPubKey].rewardAddress = rewardAddress;
+        emit StakerUpdated(blsPubKey);
+    }
+
+    function setSigningAddress(
+        bytes calldata blsPubKey,
+        address signingAddress
+    ) public onlyControlAddress(blsPubKey) {
+        require(
+            msg.sender != address(0),
```
DrZoltanFazekas, line 379: it should be `signingAddress != address(0)`

DrZoltanFazekas (changes requested):

zilliqa/src/contracts/deposit_v5.sol
86667: fixed, silly me

DrZoltanFazekas (dismissed):

DrZoltanFazekas (approved):

PR #2376, opened 2025-02-19
Fix pagination logic in ds_block_listing and improve error handling.

saeed-zil:


zilliqa/src/api/zilliqa.rs
```diff
@@ -927,14 +927,29 @@ pub fn get_current_ds_epoch(_params: Params, node: &Arc<Mutex<Node>>) -> Result<
...
 pub fn ds_block_listing(params: Params, node: &Arc<Mutex<Node>>) -> Result<DSBlockListingResult> {
     // Dummy implementation
-    let node = node.lock().unwrap();
-    let num_tx_blocks = node.get_latest_finalized_block_number()?;
-    let num_ds_blocks = (num_tx_blocks / TX_BLOCKS_PER_DS_BLOCK) + 1;
-    let max_pages = num_ds_blocks / 10;
+    let num_tx_blocks = {
+        let node = node.lock().unwrap();
+        node.get_latest_finalized_block_number()?
+    };
```
maxconway, line 933: Is this change in order to release the lock on the node earlier?
If so, maybe something like `let num_tx_blocks = node.lock().unwrap().get_latest_finalized_block_number()?;` might be a bit clearer? I think that'd still take the node out of scope, but I might be wrong?

zilliqa/src/api/zilliqa.rs
```diff
@@ -927,14 +927,29 @@ pub fn get_current_ds_epoch(_params: Params, node: &Arc<Mutex<Node>>) -> Result<
...
     let page_requested: u64 = params.one()?;
 
-    let base_blocknum = page_requested * 10;
-    let end_blocknum = num_ds_blocks.min(base_blocknum + 10);
+    if page_requested == 0 || page_requested > max_pages {
+        return Err(anyhow!(format!(
+            "Page out of range. Valid range is 1 to {}",
+            max_pages
+        )));
+    }
```
maxconway, line 949: Good idea having this

maxconway (commented):
Would it be possible to also update the tests? (`tx_block_listing`) in particular, the ds one is less important

zilliqa/src/api/zilliqa.rs
saeed-zil: I like the idea and I think it works.

maxconway (commented):
Would it be possible to have the tests check the pagination logic? So checking that the order and page size is correct?

maxconway (approved):
I'm happy

JamesHinshelwood (approved):

Issue #2400, opened 2025-02-25
Deque bug


DrZoltanFazekas:
Withdrawals from the deposit contract don't work properly because of a bug in `deque.sol`.

Here's an example of the incorrect behavior:
```
1. deposit 417 + 170 + 135 + 41
2. unstake 417
3. wait for the unbonding period
4. unstake 170
5. withdraw --> you receive 417 as expected
6. unstake 135
7. unstake  41
8. wait for the unbonding period
9. withdraw --> you receive only 170 + 41 instead of 170 + 135 + 41
```

DrZoltanFazekas:
Here's the code to reproduce the issue in Remix:
```solidity
// SPDX-License-Identifier: MIT OR Apache-2.0
pragma solidity ^0.8.20;

struct Withdrawal {
    uint256 startedAt;
    uint256 amount;
}

// Implementation of a double-ended queue of `Withdrawal`s, backed by a circular buffer.
library Deque {
    struct Withdrawals {
        Withdrawal[] values;
        // The physical index of the first element, if it exists. If `len == 0`, the value of `head` is unimportant.
        uint256 head;
        // The number of elements in the queue.
        uint256 len;
    }

    // Returns the physical index of an element, given its logical index.
    function physicalIdx(
        Withdrawals storage deque,
        uint256 idx
    ) internal view returns (uint256) {
        uint256 physical = deque.head + idx;
        // Wrap the physical index in case it is out-of-bounds of the buffer.
        if (physical >= deque.values.length) {
            return physical - deque.values.length;
        } else {
            return physical;
        }
    }

    function length(Withdrawals storage deque) internal view returns (uint256) {
        return deque.len;
    }

    // Get the element at the given logical index. Reverts if `idx >= queue.length()`.
    function get(
        Withdrawals storage deque,
        uint256 idx
    ) internal view returns (Withdrawal storage) {
        if (idx >= deque.len) {
            revert("element does not exist");
        }

        uint256 pIdx = physicalIdx(deque, idx);
        return deque.values[pIdx];
    }

    // Push an empty element to the back of the queue. Returns a reference to the new element.
    function pushBack(
        Withdrawals storage deque
    ) internal returns (Withdrawal storage) {
        // Add more space in the buffer if it is full.
        if (deque.len == deque.values.length) {
            deque.values.push();
        }

        uint256 idx = physicalIdx(deque, deque.len);
        deque.len += 1;

        return deque.values[idx];
    }

[...]

PR #2402, opened 2025-02-25
Fix deque implementation in deposit contract

JamesHinshelwood:
The old implementation was objectively broken and unnecessarily complicated. Instead, we just store elements in a `mapping(index => element)`.

### DIFF
--- a/zilliqa/src/contracts/utils/deque_v2.sol
+++ b/zilliqa/src/contracts/utils/deque_v2.sol
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity ^0.8.28;
+
+struct Withdrawal {
+    uint256 startedAt;
+    uint256 amount;
+}
+
+// A non-broken implementation of a double-ended queue of `Withdrawal`s. `deque.sol` doesn't work, don't use it.
+library Deque {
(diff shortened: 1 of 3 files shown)


zilliqa/src/contracts/utils/deque_v2.sol
```diff
@@ -0,0 +1,74 @@
...
+        uint256 idx = deque.tail;
+        deque.tail++;
+        return deque.values[idx];
+    }
+
+    // Pop an element from the front of the queue.
+    function popFront(
+        Withdrawals storage deque
+    ) internal returns (Withdrawal memory) {
+        require(deque.head < deque.tail, "queue is empty");
```
86667, line 45: I  think this should be `<=` ?

JamesHinshelwood: If `head == tail`, the queue is empty. So we require that `head != tail` (i.e. `head < tail`)

JamesHinshelwood:
Fixes #2400

zilliqa/src/contracts/utils/deque_v2.sol
86667: Sorry!

86667 (approved):

PR #2411, opened 2025-02-27
Fix panic inside leader_at_block()

86667:
This was required to fix the Bootstrap node after multiple block reverts caused some data issues. 

Before these changes a failure to find a leader caused the node to restart. With them the node errors but stays alive.


JamesHinshelwood (approved):
You can also use `.ok_or_else(|| anyhow!("failed to find leader"))?` if preferred

PR #2414, opened 2025-02-27
Ensure state is correct in both EVM and Scilla environments

JamesHinshelwood:
When an interop. transaction is run, we need to be careful to keep track of state changes in both sides of the world.

This PR takes the approach of:
* Let `revm` keep track of EVM state changes as normal.
* When an interop call is made to the Scilla interpreter, we make sure any 'pending' changes from the EVM side are visible to the Scilla contract.
* When the Scilla call finishes, if any 'pending' EVM changes were modified further, we apply those changes to the EVM state journal before transferring control back to `revm`.

This ensures the state we read is consistent and that when we come to apply state delta updates, there is no overlap between the EVM and Scilla state maps.

### DIFF
--- a/zilliqa/src/exec.rs
+++ b/zilliqa/src/exec.rs
@@ -17,7 +17,7 @@ use ethabi::Token;
 use jsonrpsee::types::ErrorObjectOwned;
 use libp2p::PeerId;
 use revm::{
-    Database, DatabaseRef, Evm, GetInspector, Inspector, inspector_handle_register,
+    Database, DatabaseRef, Evm, GetInspector, Inspector, JournaledState, inspector_handle_register,
     primitives::{
         AccountInfo, B256, BlockEnv, Bytecode, Env, ExecutionResult, HaltReason, HandlerCfg,
         KECCAK_EMPTY, Output, ResultAndState, SpecId, TxEnv,
@@ -758,6 +758,9 @@ impl State {
                 continue;
             }
 
+            // We shouldn't mutate accounts that were from EVM.
+            assert!(!account.from_evm);
+
             let mut storage = self.get_account_trie(address)?;
 
             /// Recursively called internal function which assigns `value` at the correct key to `storage`.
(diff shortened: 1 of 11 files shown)


86667 (approved):

zilliqa/src/precompiles/scilla.rs
```diff
@@ -622,6 +626,7 @@ fn scilla_call_precompile<I: ScillaInspector>(
     };
     trace!(?result, "scilla_call complete");
     if !&result.success {
+        evmctx.db = state;
```
rrw-zilliqa, line 629: This allows failed scilla transitions to modify the state - do we want to allow this?

zilliqa/src/precompiles/scilla.rs
```diff
@@ -631,7 +636,25 @@ fn scilla_call_precompile<I: ScillaInspector>(
             return err("scilla call failed");
         }
     }
-    // Move the new state back into `evmctx`.
+    state.new_state.retain(|address, account| {
+        if !account.touched {
```
rrw-zilliqa, line 640: Do you want to rely on side effects in retain() ? (it's safe, but .. )

zilliqa/src/exec.rs
```diff
@@ -1186,11 +1192,40 @@ pub struct PendingState {
...
+                } else {
+                    account
+                        .info
+                        .code
+                        .as_ref()
+                        .ok_or_else(|| anyhow!("account should have code"))?
+                        .original_bytes()
+                        .to_vec()
+                }),
+                storage_root: B256::ZERO, // There's no need to set this, since Scilla cannot query EVM contracts' state.
```
rrw-zilliqa, line 1218: It can't query the code either, so why set that?

rrw-zilliqa (approved):
I'll hit the approve button, but I'm not sure I understand the implications of either this patch or of the scilla execution mechanism generally - there's a lot of special-casing going on, and this is security critical code. That said, absent a lot of comments and probably some refactoring, I'm unlikely to get any better inside a week, and we do need to fix the bug...

One thing to check is that scilla remote contract reads still work - zeroing the storage location might be sufficiently safe, in that since scilla contracts will never end up in the journaled evm log, scilla can never legally read them via remote contract reads (or remote library references), but it would be nice to be more certain.

PR #2417, opened 2025-02-28
Update chain spec files

JamesHinshelwood:
PR #2402 forgot to update the chain specs.

This also adds a script to easily update the spec files for all networks.


shawn-zil (approved):
manually updated my devnet config.toml file according to changes here, and it worked.

z2/resources/chain-specs/zq2-devnet.toml
```diff
@@ -21,6 +21,5 @@ consensus.minimum_stake = "10_000_000_000_000_000_000_000_000"
 consensus.eth_block_gas_limit = 84000000
 consensus.gas_price = "4_761_904_800_000"
 consensus.scilla_call_gas_exempt_addrs = []
-consensus.contract_upgrade_block_heights = { deposit_v3 = 3600, deposit_v4 = 428400 }
```
86667, line 24: This doesnt look right - the default should be `deposit_v5 = 0` ?

z2/src/chain.rs
```diff
@@ -165,27 +165,22 @@ impl Chain {
...
+            }),
+            Self::Zq2ProtoTestnet => Some(ContractUpgradesBlockHeights {
                 deposit_v3: Some(8406000),
                 // estimated: 2025-02-03T13:55:00Z
                 deposit_v4: Some(10890000),
                 deposit_v5: None,
-            },
-            _ => ContractUpgradesBlockHeights::default(),
+            }),
+            _ => None,
```
86667, line 183: Yes, should be `ContractUpgradesBlockHeights::default` for Devnet here and i would argue default for all networks should be `deposit_v5` deployed. The case where we dont want any deposit contract upgrades is a rare one

JamesHinshelwood: `contract_upgrade_block_heights` is annotated with `#[serde(default)]`, so no entry uses the default. I think it makes sense to not mention the configuration at all if we're using the default?

PR #2438, opened 2025-03-04
Introduce block pruning

shawn-zil:
This PR introduces block pruning, via a `prune_interval` config value, which specifies the size of the window. It has a minimum size of 30, to avoid pruning forks. Pruning is done gradually, to avoid locking up a node; and done in ascending order from the oldest block in history, to avoid creating gaps in history.

It also introduces a `admin_blockRange` rpc end-point; and includes a `consensus::prune_interval` test to check that it is indeed pruning.


zilliqa/src/cfg.rs
```diff
@@ -159,6 +159,9 @@ pub struct NodeConfig {
     /// Maximum allowed RPC response size
     #[serde(default = "max_rpc_response_size_default")]
     pub max_rpc_response_size: u32,
+    /// The N number of historical blocks to be kept in the DB during pruning. N > 30.
```
JamesHinshelwood, line 162: `N > 30` or `N > 20`?

Also why `20/30`? Is this the maximum number of blocks we think could be non-finalized at once?

I'd consider just picking an obviously correct number like `1000`. Its trivially easy to prove to yourself that we couldn't have `1000` unfinalized blocks and `1000` is still a very small number of blocks in terms of storage space?

zilliqa/src/consensus.rs
```diff
@@ -177,9 +177,16 @@ pub struct Consensus {
...
     pub new_transaction_hashes: broadcast::Sender<Hash>,
+    /// Pruning interval i.e. how many blocks to keep in the database.
+    prune_interval: u64,
 }
 
 impl Consensus {
+    // gradual pruning interval
+    const PRUNE_INTERVAL: u64 = 1000;
+    // prune interval > 20 to ensure we don't prune forks
+    const MIN_PRUNE_INTERVAL: u64 = 20;
```
JamesHinshelwood, line 188: IMO we should validate the config to be greater than this (and fail to start if it isn't) instead of silently increasing the configured value.

zilliqa/src/consensus.rs
```diff
@@ -3091,9 +3101,30 @@ impl Consensus {
...
         }
 
+        self.prune_history(block.header.number)?;
         self.broadcast_commit_receipts(from, block, block_receipts)
     }
 
+    fn prune_history(&mut self, number: u64) -> Result<()> {
+        if self.prune_interval != u64::MAX {
+            let range = self.db.available_range()?;
+            let prune_max = range.start().saturating_add(Self::PRUNE_INTERVAL); // gradually prune N-blocks at a time
```
JamesHinshelwood, line 3111: `self.prune_interval` and `Self::PRUNE_INTERVAL` do different things but they have exactly the same name so its hard to work out the difference. Can we come up with better names?

zilliqa/src/consensus.rs
```diff
@@ -3091,9 +3101,30 @@ impl Consensus {
             ));
         }
 
+        self.prune_history(block.header.number)?;
```
JamesHinshelwood, line 3104: Hmm - It would be nice if we can do this out-of-band with block processing. But maybe there isn't any other convenient place to trigger it?

zilliqa/src/cfg.rs
shawn-zil: Sure. We just need to pick a number. I recall Zoltan telling me that any fork longer than 15 is improbable. So, anything more than that should be fine. Hence 20. But 100 or 1000 is ok too.

zilliqa/src/consensus.rs
shawn-zil: Alright. I'll change the behavior to fail instead of clamping.

zilliqa/src/consensus.rs
shawn-zil: Sure. I thought upper case as constant was obvious. I'll rename it.

zilliqa/src/consensus.rs
shawn-zil: I did think about either triggering it periodically on a timer, or doing it during an exponential back off, or doing it every N proposals.

I'll change it to doing it every N proposals where N is the interval size - middle ground between doing it too often vs letting the DB grow significantly larger than the desired interval.

zilliqa/src/cfg.rs
shawn-zil: I'll pick 1000.

zilliqa/src/consensus.rs
shawn-zil: Actually, after giving it more thought, it would probably be a better idea to just prune on every proposal, which means that the average prune rate would approach 1-prune/sec, rather than to do it by large batches, which could lock up the node.

zilliqa/src/cfg.rs
shawn-zil: I've picked 30 for now
https://github.com/Zilliqa/zq2/issues/1119#issuecomment-2699477630

JamesHinshelwood: It needs to be at least 256 (maybe 257?) to execute EVM transactions successfully.

shawn-zil: OK. Then I'll set the minimum to 300 to be safe. But this will make the unit-test run for 10x longer...

zilliqa/src/consensus.rs
```diff
@@ -317,6 +322,15 @@ impl Consensus {
             peers.clone(),
         )?;
 
+        let prune_interval = if config.prune_interval >= Self::MIN_PRUNE_INTERVAL {
```
86667, line 325: This would be better placed in `NodeConfig.validate()`

shawn-zil: Good idea!

86667 (approved):

PR #2439, opened 2025-03-04
Sync at fork, and other fixes.

shawn-zil:
1. Fixes the issue where, if an active-sync started on a losing branch of a fork, it may get stuck since all responses it receives are from the winning branch.
2. Fixes the issue where an active-sync might shoot past a checkpoint, by checking every single header downloaded against the blocks in the DB; and the first canonical block that hits the DB will turn the sync around.
3. Routes syncing blocks to lower-priority broadcast to avoid holding up the Requests queue with sync messages. Will hopefully reduce pressure on busy validators.
4. Minor fixes to syncing stats.

### DIFF
--- a/zilliqa/src/node.rs
+++ b/zilliqa/src/node.rs
@@ -944,10 +944,8 @@ impl Node {
             } else {
                 self.message_sender.broadcast_proposal(message)?;
             }
-        } else {
-            self.consensus.sync.sync_from_proposal(proposal)?;
         }
-
+        self.consensus.sync.sync_from_proposal(proposal)?;
         Ok(())
     }
 
--- a/zilliqa/src/node.rs
+++ b/zilliqa/src/node.rs
@@ -224,6 +224,11 @@ impl Node {
     pub fn handle_broadcast(&mut self, from: PeerId, message: ExternalMessage) -> Result<()> {
         debug!(%from, to = %self.peer_id, %message, "handling broadcast");
         match message {
+            // This just breaks down group block messages into individual messages to stop them blocking threads
+            // for long periods.
+            ExternalMessage::InjectedProposal(p) => {
+                self.handle_injected_proposal(from, p)?;
+            }
             // `NewTransaction`s are always broadcasted.
             ExternalMessage::NewTransaction(t) => {
                 // Don't process again txn sent by this node (it's already in the mempool)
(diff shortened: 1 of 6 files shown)


zilliqa/src/p2p_node.rs
```diff
@@ -384,7 +384,15 @@ impl P2pNode {
                             debug!(%from, %dest, %message, ?request_id, "sending direct message");
                             let id = format!("{:?}", request_id);
                             if from == dest {
-                                self.send_to(&topic.hash(), |c| c.requests.send((from, id, message, ResponseChannel::Local)))?;
+                                match message {
+                                    // Route sync messages as broadcast, to allow other requests to be prioritized.
+                                    ExternalMessage::InjectedProposal(_) => {
```
JamesHinshelwood, line 389: `InjectedProposal`s only appear after a block response has been processed right? I can see the argument that it makes sense to prioritise consensus messages over block **requests**, but not block responses. If I've requested a block and got the response, I am probably out-of-sync in which case the response is actually more important than any live consensus messages to me?

Or am I misunderstanding the message types?

zilliqa/src/node.rs
```diff
@@ -356,7 +356,7 @@ impl Node {
                     .into_iter()
                     .map(|bh| SyncBlockHeader {
                         header: bh,
-                        size_estimate: usize::MIN,
+                        size_estimate: (1024 * 1024 * bh.gas_used.0 / bh.gas_limit.0) as usize, // guesstimate
```
JamesHinshelwood, line 359: Why do we need to estimate? `std::mem::size_of<Blockheader>()` should be static?

JamesHinshelwood (approved):

zilliqa/src/node.rs
shawn-zil: It's not the size of the block header that we need. It's the size of the actual proposal. In the case of the new sync request, we actually encode it with cbor and return the value.

This estimate is just a placeholder in case of an older node that does not support the actual size computation.

After a future hardfork, this should be removed.

zilliqa/src/p2p_node.rs
JamesHinshelwood: I think its the case that if the validator is out-of-sync, it won't be able to deal with the newviews and votes, because it won't have the block which those votes are for.

shawn-zil: Alright. In that case, I'll switch it back to the previous behavior.

shawn-zil: Unless there is a fork, where the node is out-of-sync on one branch, but has the blocks for the other branch. Would that be possible?

zilliqa/src/node.rs
JamesHinshelwood: Ok if we will remove it later I won't complain too much that its a very bad estimate :)

JamesHinshelwood (approved):

zilliqa/src/p2p_node.rs
JamesHinshelwood: Most of the time the network will have continued producing blocks along some branch, so the node will have to download some blocks anyway.

JamesHinshelwood (approved):

PR #2443, opened 2025-03-04
Clean up contract upgrade config

86667:


zilliqa/src/cfg.rs
```diff
@@ -659,20 +684,70 @@ impl ContractUpgradesBlockHeights {
...
+                            k,
+                            toml::Value::Table(
+                                // ContractUpgradeConfig
+                                json!(v)
+                                    .as_object()
+                                    .unwrap()
+                                    .clone()
+                                    .into_iter()
+                                    .filter_map(|(k2, v2)| {
+                                        if v2.is_u64() {
```
JamesHinshelwood, line 697: This feels extremely confusing and error-prone. What exactly are we trying to achieve with this function? Surely there's a saner way?

JamesHinshelwood (dismissed):
LGTM, except `to_toml`!

zilliqa/src/cfg.rs
86667: Wow that is bad. Pushed improvement

JamesHinshelwood (approved):

PR #2450, opened 2025-03-05
Syncing leader must not propose.

shawn-zil:
The issue is difficult to reproduce. However, looking at the code, there are some parts that look suspicious i.e. the early proposal is assumed to be good, and gets proposed regardless of syncing state.

I feel that the basic principle should be, that an out-of-sync node should not make Proposals, and should only focus on active-sync. This PR handles the following cases:
1. If a node is out-of-sync; it prevents early proposal assembly.
2. If a node becomes out-of-sync, after early proposal assembly; it recovers the proposal.
3. If for some reason, an early proposal is out-of-sync when the early proposal is finished; it recovers the proposal.
4. If the node is back in-sync before time to finish the proposal, it will get the chance to produce a new proposal at the end.

In the cases (1)-(3), the result would be a timeout or missing proposal; which should be acceptable since the proposal would be a fork (possibly too far back to make a difference) as the node is out-of-sync. Case (4) results in the normal case.

This PR also moves InjectedProposal processing back to Requests, instead of Broadcasts.


zilliqa/src/consensus.rs
```diff
@@ -1254,8 +1259,17 @@ impl Consensus {
...
-            return Ok(());
+        if let Some((proposal, applied_txs, txns, rcpts)) = self.early_proposal.take() {
+            if proposal.view() == view {
+                // Do nothing if we are already in the correct view
+                self.early_proposal = Some((proposal, applied_txs, txns, rcpts));
+                return Ok(());
+            } else {
+                // view changed, recover from early proposal
+                for txn in applied_txs.into_iter().rev() {
+                    self.transaction_pool.insert_ready_transaction(txn)?;
```
86667, line 1270: The transactions have been added to the `transactions` table by this point, do we want to remove them from there also? I think elsewhere we add to the table only when a tx is included in an executed block

86667 (dismissed):

zilliqa/src/consensus.rs
shawn-zil: Since the transactions are `INSERT OR IGNORE INTO transactions...` I think it's ok for the txns to exist in the `transactions` table as it doesn't disrupt any subsequent insertion. Unless there's some reason not to? Does it disrupt Consensus somewhere down the line?

86667: > In fact, the transactions are inserted into the DB upon receiving a NewTransaction..

I may be missing the code but I dont think this is true.

Not one to fix now but something to consider: currently i believe we have txs in memory until they are executed as part of a valid block, at which point we add to `transactions` and `receipts` tables. I dont necessarily agree that this is the best way to do it but with these changes we have a situation in which txs may or may not be in the table before being included in a block.

We should either add all mempool txs to the table or only add those txs which are included in blocks which are a part of the current chain.

86667 (approved):

PR #2454, opened 2025-03-06
fix: Incorrect value is transferred to zero account as the gas fee

saeed-zil:


JamesHinshelwood (changes requested):
LGTM, but this needs to be conditionalised on a fork.

zilliqa/src/cfg.rs
```diff
@@ -547,6 +548,10 @@ pub struct ForkDelta {
     /// exempt addresses.
     #[serde(default)]
     pub scilla_call_gas_exempt_addrs: Vec<Address>,
+
+    /// If true, the total gas fee for all transactions in a block will be transferred to the zero account.
+    /// Before this fork, we were incorrectly transferring the gas used by all transactions to the zero account.
```
JamesHinshelwood, line 553:
```suggestion
    /// If true, the total gas paid by all transactions in a block is transferred to the zero address. This keeps the total supply of the network constant. If false, we still transfer funds to the zero address, but with an incorrect gas price of 1 Wei per gas.
```

minor wording changes - take it or leave it :)

zilliqa/src/consensus.rs
```diff
@@ -111,6 +111,7 @@ type EarlyProposal = (
     Vec<VerifiedTransaction>,
     EthTrie<MemoryDB>,
     EthTrie<MemoryDB>,
+    u128, // Cumulative gas fee which will be sunk to ZERO account
```
JamesHinshelwood, line 114:
```suggestion
    u128, // Cumulative gas fee which will be sent to ZERO account
```
?

zilliqa/src/consensus.rs
```diff
@@ -319,6 +321,7 @@ impl Consensus {
 
         let mut consensus = Consensus {
             secret_key,
+            forks: config.consensus.get_forks()?,
```
JamesHinshelwood, line 324: Can we use `state.forks` instead of making an extra copy here?

zilliqa/src/consensus.rs
```diff
@@ -3111,11 +3137,18 @@ impl Consensus {
         )?;
 
         // ZIP-9: Sink gas to zero account
+        let fork = self.forks.get(block.header.number);
```
JamesHinshelwood, line 3140: Would be nice in future if we can de-duplicate this logic?

z2/resources/chain-specs/zq2-protomainnet.toml
```diff
@@ -30,10 +30,10 @@ consensus.gas_price = "4_761_904_800_000"
...
     { at_height = 5342400, failed_scilla_call_from_gas_exempt_caller_causes_revert = true, call_mode_1_sets_caller_to_parent_caller = true },
     { at_height = 7685881, scilla_json_preserve_order = true },
-    { at_height = 7966800, scilla_messages_can_call_evm_contracts = true, scilla_contract_creation_increments_account_balance = true },
+    { at_height = 7966800, scilla_messages_can_call_evm_contracts = true, scilla_contract_creation_increments_account_balance = true, transfer_gas_fee_to_zero_account = true },
```
JamesHinshelwood, line 38: We obviously can't enable this fork at a height that has already passed. We should remove the fork from `mainnet` and enable it later. For `testnet`, you can add a new fork in a couple of days and deploy latest `main` to proto-testnet as soon as your PR is merged. Also let @DrZoltanFazekas know if you do this, so he can update his nodes.

z2/resources/chain-specs/zq2-prototestnet.toml
```diff
@@ -30,11 +30,11 @@ consensus.gas_price = "4_761_904_800_000"
...
 
 api_servers = [{ port = 4201, enabled_apis = [{ namespace = "eth", apis = ["blockNumber"] }] }, { port = 4202, enabled_apis = ["admin", "debug", "erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa"] }]
-consensus.genesis_fork = { at_height = 0, call_mode_1_sets_caller_to_parent_caller = false, failed_scilla_call_from_gas_exempt_caller_causes_revert = false, scilla_messages_can_call_evm_contracts = false, scilla_contract_creation_increments_account_balance = false, scilla_json_preserve_order = false, scilla_call_respects_evm_state_changes = false, only_mutated_accounts_update_state = false, scilla_call_gas_exempt_addrs = [] }
+consensus.genesis_fork = { at_height = 0, call_mode_1_sets_caller_to_parent_caller = false, failed_scilla_call_from_gas_exempt_caller_causes_revert = false, scilla_messages_can_call_evm_contracts = false, scilla_contract_creation_increments_account_balance = false, scilla_json_preserve_order = false, scilla_call_respects_evm_state_changes = false, only_mutated_accounts_update_state = false, scilla_call_gas_exempt_addrs = [] , transfer_gas_fee_to_zero_account = false }
```
JamesHinshelwood, line 33: These files need to be generated by `z2/update-chain-specs.sh` (or run the `z2` command manually).

The source of truth is in `chain.rs` which needs to be updated.

JamesHinshelwood (changes requested):
The actual change LGTM, but we need to change some things in the configuration.

JamesHinshelwood (approved):
LGTM, thanks

JamesHinshelwood (approved):

Issue #2378, opened 2025-02-21
Improve local docker network


shawn-zil:
Continue fixing local docker network issues, which result also in syncing only against 2 nodes.

shawn-zil:
```
zilliqanode3  | 2025-03-06T04:21:55.226526Z  INFO zilliqa::p2p_node: 261: identify event peer_id=12D3KooWLA4xVjiGszqmYJmt8E1NTurVeCujDi17FoSzSDDDKUjT info=Info { public_key: PublicKey { publickey: Ed25519(PublicKey(compressed): 99a1653308455fd47a5e3284210d443618495fcdb57dc5240a7e5519f341ffa) }, protocol_version: "zilliqa/1.0.0", agent_version: "rust-libp2p/0.46.0", listen_addrs: [/ip4/198.52.100.102/tcp/42492], protocols: ["/meshsub/1.0.0", "/zq2-message/1", "/ipfs/id/push/1.0.0", "/meshsub/1.1.0", "/ipfs/id/1.0.0", "/ipfs/kad/1.0.0"], observed_addr: /ip4/198.52.100.103/tcp/5643 }
zilliqanode2  | 2025-03-06T04:21:55.268225Z  INFO zilliqa::p2p_node: 261: identify event peer_id=12D3KooWPXw2dXBRH1bT4vcNos9f6W2KoFTiarqptBuTzxaXg7zu info=Info { public_key: PublicKey { publickey: Ed25519(PublicKey(compressed): cbcde5f3e253d88c94a96052f7e6936ed569f35b17547b149ec1ee8bd4d9ce) }, protocol_version: "zilliqa/1.0.0", agent_version: "rust-libp2p/0.46.0", listen_addrs: [/ip4/198.52.100.103/tcp/5643], protocols: ["/ipfs/kad/1.0.0", "/zq2-message/1", "/libp2p/autonat/2/dial-back", "/meshsub/1.1.0", "/meshsub/1.0.0", "/ipfs/id/1.0.0", "/ipfs/id/push/1.0.0", "/libp2p/autonat/2/dial-request"], observed_addr: /ip4/198.52.100.102/tcp/42492 }
[...]

PR #2464, opened 2025-03-07
Bind libp2p kademlia and identify

shawn-zil:
Making things more compatible as-per documentation. It is unclear if it solves the issue of nodes seemingly unable to talk to each other - sometimes. Running on `perftest` and does not seem to do any harm. So, I think we should merge this first.


zilliqa/src/p2p_node.rs
```diff
@@ -255,6 +256,16 @@ impl P2pNode {
                         SwarmEvent::NewListenAddr { address, .. } => {
                             info!(%address, "P2P swarm listening on");
                         }
+                        SwarmEvent::Behaviour(BehaviourEvent::Identify(identify::Event::Received { peer_id, info, .. })) => {
+                            info!(%peer_id, ?info, "identify event");
+                            self.swarm.add_external_address(info.observed_addr);
```
JamesHinshelwood, line 261: This is wrong.

> Add a confirmed external address for the local node.
> This function should only be called with addresses that are guaranteed to be reachable. The address is broadcast to all [NetworkBehaviour](https://docs.rs/libp2p/latest/libp2p/swarm/trait.NetworkBehaviour.html)s via [FromSwarm::ExternalAddrConfirmed](https://docs.rs/libp2p/latest/libp2p/swarm/enum.FromSwarm.html#variant.ExternalAddrConfirmed).

Autonat will add the external address for us once it is **confirmed**.

JamesHinshelwood (approved):
LGTM, except the `add_external_address`. Feel free to merge and fix after, but we should fix it.

zilliqa/src/p2p_node.rs
shawn-zil: OK. I'll remove it.

JamesHinshelwood (approved):

JamesHinshelwood (dismissed):

zilliqa/src/p2p_node.rs
86667: This line is required in order for nodes running locally to add their addresses to Kademlia, else they will only connect with nodes they heard about from Bootstrap at startup.

We either need to include this line or understand why Autonat never adds the addresses itself

86667: https://github.com/Zilliqa/zq2/pull/2285/files#r1942372969

JamesHinshelwood: There are two things going on here:

* We are setting our external address once a single peer tells us it observed an address for us. This is what I think is bad and we should be relying on autonat for. As you say, if autonat isn't working then we should find and fix that problem.
* We are adding peer addresses once a peer tells us they exist. This is what the documentation tells us to do and I agree with.

zilliqa/src/p2p_node.rs
```diff
@@ -255,6 +256,15 @@ impl P2pNode {
                         SwarmEvent::NewListenAddr { address, .. } => {
                             info!(%address, "P2P swarm listening on");
                         }
+                        SwarmEvent::Behaviour(BehaviourEvent::Identify(identify::Event::Received { peer_id, info, .. })) => {
+                            info!(%peer_id, ?info, "identify event");
+                            // this is necessary - https://docs.rs/libp2p-kad/latest/libp2p_kad/#important-discrepancies
+                            if info.protocols.iter().any(|p| *p == kad::PROTOCOL_NAME) {
+                                for addr in info.listen_addrs {
+                                    self.swarm.behaviour_mut().kademlia.add_address(&peer_id, addr);
```
JamesHinshelwood, line 264: Should this be `self.swarm.add_peer_address(peer_id, addr);`? The old version did that instead. I don't know what the difference is exactly, but I guess if there were other protocols that need to know peer addresses this would help? Then again you are filtering by the protocol name above so maybe we shouldn't?

shawn-zil: I cannot find evidence that `add_peer_address()` actually adds the value to Kademlia. If we wanted to, we could do both, just to be sure.

zilliqa/src/p2p_node.rs
shawn-zil: Added some code to `add_external_address()` upon receiving the `ExternalAddrConfirmed` event. So, it's not some random peer informing us of an observed address, but when autonat receives a successful dialback.

86667:
FYI this does not solve local peers being unable to talk to eachother and thus wont close - https://github.com/Zilliqa/zq2/issues/2378

Happy to merge if there are other benefits, though.

zilliqa/src/p2p_node.rs
```diff
@@ -255,12 +251,37 @@ impl P2pNode {
                         SwarmEvent::NewListenAddr { address, .. } => {
                             info!(%address, "P2P swarm listening on");
                         }
+                        // this is necessary - https://docs.rs/libp2p-kad/latest/libp2p_kad/#important-discrepancies
+                        SwarmEvent::Behaviour(BehaviourEvent::Identify(identify::Event::Received { peer_id, info, .. })) => {
+                            debug!(%peer_id, ?info, "identify event");
+                            if info.protocols.iter().any(|p| *p == kad::PROTOCOL_NAME) {
```
JamesHinshelwood, line 257: Should we remove the protocol check now we are adding peers to the whole swarm?

zilliqa/src/p2p_node.rs
```diff
@@ -255,12 +251,37 @@ impl P2pNode {
                         SwarmEvent::NewListenAddr { address, .. } => {
                             info!(%address, "P2P swarm listening on");
                         }
+                        // this is necessary - https://docs.rs/libp2p-kad/latest/libp2p_kad/#important-discrepancies
+                        SwarmEvent::Behaviour(BehaviourEvent::Identify(identify::Event::Received { peer_id, info, .. })) => {
+                            debug!(%peer_id, ?info, "identify event");
```
JamesHinshelwood, line 256: This is already logged on line 249 (and the same applies for the other new `debug!` logs)

zilliqa/src/p2p_node.rs
```diff
@@ -255,12 +251,37 @@ impl P2pNode {
...
+                        }
+                        SwarmEvent::NewExternalAddrOfPeer{peer_id, address} => {
+                            debug!(%peer_id, %address, "new peer");
+                            self.swarm.behaviour_mut().kademlia.add_address(&peer_id, address);
+                            self.swarm.behaviour_mut().gossipsub.add_explicit_peer(&peer_id);
+                        }
+                        SwarmEvent::ExternalAddrConfirmed{address} => {
+                            debug!(%address, "confirmed");
+                            self.swarm.add_external_address(address);
+                        }
```
JamesHinshelwood, line 276: The `ExternalAddrExpired` and `ExternalAddrConfirmed` handlers are already implemented by `Swarm`:
https://github.com/libp2p/rust-libp2p/blob/88f7875ad1a3e240aa2d9b9fb6f6c5354f1a62eb/swarm/src/lib.rs#L1139-L1148

zilliqa/src/p2p_node.rs
```diff
@@ -255,12 +251,37 @@ impl P2pNode {
...
+                            }
+                        }
+                        SwarmEvent::ExternalAddrExpired{address} => {
+                            debug!(%address, "expired");
+                            self.swarm.remove_external_address(&address);
+                        }
+                        SwarmEvent::NewExternalAddrOfPeer{peer_id, address} => {
+                            debug!(%peer_id, %address, "new peer");
+                            self.swarm.behaviour_mut().kademlia.add_address(&peer_id, address);
+                            self.swarm.behaviour_mut().gossipsub.add_explicit_peer(&peer_id);
```
JamesHinshelwood, line 271: This adds the address as an 'explicit peer' in Gossipsub, which means it ignores lots of the gossipsub logic around scoring, etc.

Is that intended?

`gossipsub::Behaviour` already adds 'normal' peers when a connection is established:
https://github.com/libp2p/rust-libp2p/blob/88f7875ad1a3e240aa2d9b9fb6f6c5354f1a62eb/protocols/gossipsub/src/behaviour.rs#L2888

zilliqa/src/p2p_node.rs
JamesHinshelwood: And maybe add it in the `NewExternalAddrOfPeer` handler instead?

JamesHinshelwood: Or I guess we could:
1. Call `kademlia.add_address` **here** if the protocol matches
2. Call `swarm.add_peer_address` unconditionally here.
?

JamesHinshelwood (approved):

JamesHinshelwood (approved):

PR #2466, opened 2025-03-07
Fix reading and returning complex maps

bzawisto:
Scilla accepts quoted strings whereas `serde_json` unescapes them. In addition, complex map should be returned at a proper key.


JamesHinshelwood:
Replaced by https://github.com/Zilliqa/zq2/pull/2486

PR #2468, opened 2025-03-07
Return correct block for Scilla `CURRENTBLOCK` query

JamesHinshelwood:
The previous implementation incorrectly returned `CURRENTBLOCK` as the number of the last mined block. This adds a fork which fixes the behaviour to return the current block. Also, the latest block of the chain was always used. Instead we now pass the current block from the execution layer, meaning transactions can be replayed and produce the same result.

This also limits the maximum range of `BLOCKHASH` and `TIMESTAMP` queries to 256 blocks from the current block.


zilliqa/src/scilla.rs
```diff
@@ -355,6 +356,7 @@ impl Scilla {
         value: ZilAmount,
         init: &ContractInit,
         ext_libs_dir: &ScillaExtLibsPathInScilla,
+        current_block: u64,
```
JamesHinshelwood, line 359: Investigating another related discrepancy at 
https://explorer.zq2-protomainnet.zilliqa.com/tx/0x472f1dc54ce1aa317f222f28c2dec97ef9ac1a600043633446d263e97406c413

Leave unresolved until done.

86667 (approved):

PR #2482, opened 2025-03-10
Return correct object for GetSmartContractSubState endpoint.

saeed-zil:
* The second and third parameters are not optional for `GetSmartContractSubState` based on ZQ1
* If the second parameter to `GetSmartContractSubState` is an empty string, it returns the entire state
* If not, `GetSmartContractSubState` response doesn't include `_balance`


maxconway (changes requested):
I'm getting a failure for `cargo test get_smart_contract_sub_state`. I'm guessing the test needs updating to match the changes?

maxconway (approved):
LGTM :)

PR #2483, opened 2025-03-10
feat: allow specification of Scilla server port in start_up script

86667:


z2/resources/start_node.tera.sh
```diff
@@ -8,17 +8,19 @@ NODE_PRIVATE_KEY=""
 CONFIG_FILE="${CHAIN_NAME}.toml"
 CHECKPOINT_FILE=""  # The optional checkpoint file to mount, if provided.
 DATA_FOLDER=$(pwd)/data
+SCILLA_SERVER_PORT=3111
```
JamesHinshelwood, line 11: FWIW I'd pick something between 10,000 and 65,000 if we want collisions to be most unlikely.

z2/resources/start_node.tera.sh
```diff
@@ -8,17 +8,19 @@ NODE_PRIVATE_KEY=""
...
 
         Options:
-            -k, --key           NODE_PRIVATE_KEY (mandatory)
-            -c, --config        Path to the config file (optional, default: ${CHAIN_NAME}.toml)
-            -p, --checkpoint    Path to the checkpoint file (optional, it is needed only the first time a node is started)
-            -h, --help          Display this help message
+            -k, --key                   NODE_PRIVATE_KEY (mandatory)
+            -c, --config                Path to the config file (optional, default: ${CHAIN_NAME}.toml)
+            -p, --checkpoint            Path to the checkpoint file (optional, it is needed only the first time a node is started)
+                --scilla-server-port    Specify Scilla Server port
```
JamesHinshelwood, line 22: Should we mention here that `consensus.scilla_address` needs to match?

Also if we're updating the default, we should update the default address in `cfg.rs` too.

JamesHinshelwood (approved):

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):

JamesHinshelwood (dismissed):

JamesHinshelwood (approved):

PR #2498, opened 2025-03-12
fix: ensure empty maps are included in smart contract state response

saeed-zil:


maxconway (commented):
Might be worth adding in a test?

saeed-zil:
> Might be worth adding in a test?

Tests have already been added in SimpleMap.ts.

maxconway (approved):
lgtm

PR #2510, opened 2025-03-13
Implement txpool_contentFrom

### LINKED ISSUE #2502: Implement txpool_contentFrom

maxconway:
Implement txpool_contentFrom

- Plus docs
- Plus tests
- Closes #2502

Note: Depends on #2509 - do not merge until #2509 is merged, then do a rebase or merge for a cleaner diff


zilliqa/src/api/txpool.rs
```diff
@@ -58,8 +59,41 @@ fn txpool_content(
...
-fn txpool_content_from(_params: Params, _node: &Arc<Mutex<Node>>) -> Result<()> {
-    todo!("Endpoint not implemented yet")
+fn txpool_content_from(
+    params: Params,
+    node: &Arc<Mutex<Node>>,
+) -> Result<types::txpool::TxPoolContent> {
+    let address: super::zilliqa::ZilAddress = params.one()?;
+    let address: Address = address.into();
+    let node = node.lock().unwrap();
+    let content = node.txpool_content()?;
```
JamesHinshelwood, line 69: It seems very wasteful to get the full txpool content and then only later filter it to a single `from` address. Can we filter earlier and save a bunch of effort?

JamesHinshelwood (dismissed):

zilliqa/src/api/txpool.rs
maxconway: I've had a look at this but unless I'm missing something we don't have an index into the txpool by address?
So I the best approach would be to add such an index, e.g. a probably `HashMap<Address, HashSet<TxIndex>> to map from transaction signers to transaction incidences, which can then index into the main transactions BTreeMap. But obviously there's the overhead to that of maintaining the map.

Would you like me to do that or is there something else you had in mind?

JamesHinshelwood (approved):

PR #2522, opened 2025-03-14
Only inject newer blocks during Phase-3.

shawn-zil:
This PR does 2 things:
1. Extracts a chain-segment from the cache-blocks during Phase-3, instead of blindly injecting everything, which can trigger forking issues not handled by `deal_with_fork()`.
2. Speeds-up Phase-2 by driving it even without Proposals.

It's currently deployed in `perftest`

### DIFF
--- a/zilliqa/src/sync.rs
+++ b/zilliqa/src/sync.rs
@@ -244,16 +244,15 @@ impl Sync {
         match self.state {
             // Check if we are out of sync
             SyncState::Phase0 if self.in_pipeline == 0 => {
-                // Guarantee that active-sync only requests missing blocks - https://github.com/Zilliqa/zq2/issues/2520
-                let meta = self.recent_proposals.front().unwrap().header;
+                let meta = self.recent_proposals.back().unwrap().header;
                 let parent_hash = meta.qc.block_hash;
                 // No parent block, trigger sync
                 if !self.db.contains_canonical_block(&parent_hash)? {
                     self.active_sync_count = self.active_sync_count.saturating_add(1);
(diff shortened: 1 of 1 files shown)


zilliqa/src/sync.rs
```diff
@@ -1002,6 +1046,16 @@ impl Sync {
     pub fn mark_received_proposal(&mut self, number: u64) -> Result<()> {
         tracing::trace!(%number, "sync::MarkReceivedProposal : received");
         self.in_pipeline = self.in_pipeline.saturating_sub(1);
+        // speed-up block transfers, w/o waiting for proposals
+        if Self::DO_SPECULATIVE {
```
86667, line 1050: I dont understand this, we're still waiting for a Proposal to arrive to get to this point in the code?

shawn-zil: This function is called as part of the normal execution - when an injected proposal is processed. So, at this point, we trigger fetching the next batch of proposals, when necessary.

86667: This fixes the issue of a node starting up during a period where there are no new proposals will not sync, right? I didnt think there would be any injected proposals in this time either as `recent_blocks` would not be added too

86667 (approved):

JamesHinshelwood:
Can we test this against proto-mainnet before merging?

shawn-zil:
> Can we test this against proto-mainnet before merging?

having trouble getting past a scilla txn (I've had trouble with Scilla tests). I've been running this on `perftest` and local docker network. So, I think that I'll merge this first.

PR #2524, opened 2025-03-17
fix: add a separator after variable name in storage key generation to prevent prefix collisions.

saeed-zil:


evm_scilla_js_tests/artifacts/scilla.cache
saeed-zil, line 1: This file can be treated like pnpm-lock.json or Cargo.lock. Sometimes the content is incorrect, and the developer needs to delete this file manually to generate the correct one. Instead, we can include it in our codebase to ensure we always have the right version. Please object if you disagree.

JamesHinshelwood (commented):
I think the fix makes perfect sense, but we can't merge it because all the existing data exists in `proto-{main,test}net`. We either need a solution that works with the existing data or we need to do some kind of data migration. I guess we can discuss on Slack what makes sense.

evm_scilla_js_tests/artifacts/scilla.cache
saeed-zil: Executing `scilla-fmt` through Docker for numerous contracts and parsing the S-expression output for every test run can be quite inefficient. Moreover, when the `hardhat-scilla-plugin` is updated, the `scilla.cache` file may become outdated without being automatically regenerated. This happens because the file is only refreshed when modifications are made to the contracts. While we can maintain the approach of not tracking it in the codebase, there are instances when test failures occur, requiring manual deletion of the cached file to trigger regeneration.

saeed-zil:
> I think the fix makes perfect sense, but we can't merge it because all the existing data exists in `proto-{main,test}net`. We either need a solution that works with the existing data or we need to do some kind of data migration. I guess we can discuss on Slack what makes sense.

What if, instead of an assertion, we do a `continue` with a warning?

https://github.com/Zilliqa/zq2/blob/9f2f246393e9d35280e079a3741d66dd2c37204e/zilliqa/src/exec.rs#L1429

chetan-zilliqa:
Is this PR still in progress, or good to merge?

PR #2541, opened 2025-03-17
fix: apply scilla state changes only if transaction was successful.

saeed-zil:


JamesHinshelwood (changes requested):
This needs a fork and means failed transactions won't charge gas or increase the nonce, which is not what we want.

maxconway:
> This needs a fork and means failed transactions won't charge gas or increase the nonce, which is not what we want.

Is this fixed now, @saeed-zil?

saeed-zil:
> > This needs a fork and means failed transactions won't charge gas or increase the nonce, which is not what we want.
> 
> Is this fixed now, @saeed-zil?

Yes :)

zilliqa/src/exec.rs
```diff
@@ -715,7 +715,29 @@ impl State {
...
+                self.apply_delta_scilla(&state, current_block.number)?;
+            } else {
+                // If the transaction rejected, we must update the nonce and balance of the sender account.
+                let from_account = state
+                    .get(&from_addr)
+                    .ok_or(anyhow!("from account not found"))?;
+
+                let mut storage = self.get_account_trie(from_addr)?;
+                let account = Account {
+                    nonce: from_account.account.nonce,
```
bzawisto, line 733: Is the updated `nonce = previous nonce + 1`?
The same for balance - is `balance = balance - txn.gas_limit() * gas_price()`?

its-saeed: Tests in `RejectedTransition.ts` can verify this.

bzawisto (approved):

PR #2549, opened 2025-03-19
Move sync data from SQL to in-memory

shawn-zil:
This PR:
1. Moves the syncing metadata functionality from SQL/DB into in-memory structures; and
2. Minor change from Range to RangeInclusive that is semantically correct.

### DIFF
--- a/zilliqa/src/sync.rs
+++ b/zilliqa/src/sync.rs
@@ -1356,9 +1356,7 @@ impl SyncSegments {
 
     /// Peeks into the top of the segment stack.
     fn last_sync_segment(&self) -> Option<(BlockHeader, PeerInfo)> {
-        let Some((hash, peer)) = self.segments.last() else {
-            return None;
-        };
+        let (hash, peer) = self.segments.last()?;
         let header = self.headers.get(hash).cloned()?;
         let peer = PeerInfo {
             last_used: Instant::now(),
@@ -1370,14 +1368,14 @@ impl SyncSegments {
 
     /// Pushes a particular segment into the stack.
     fn push_sync_segment(&mut self, peer: &PeerInfo, meta: &BlockHeader) {
-        self.headers.insert(meta.hash, meta.clone());
+        self.headers.insert(meta.hash, *meta);
         self.segments.push((meta.hash, peer.clone()));
     }
 
     /// Bulk inserts a bunch of metadata.
     fn insert_sync_metadata(&mut self, metas: &Vec<BlockHeader>) {
         for meta in metas {
-            self.headers.insert(meta.hash, meta.clone());
+            self.headers.insert(meta.hash, *meta);
         }
     }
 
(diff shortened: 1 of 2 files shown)


zilliqa/src/sync.rs
```diff
@@ -773,8 +775,8 @@ impl Sync {
             } else {
                 false
             }
-        }) {
-            self.db.push_sync_segment(&segment_peer, header)?;
+        }).collect_vec().into_iter().rev() {
+            self.segments.push_sync_segment(&segment_peer, header);
         }
```
JamesHinshelwood, line 780: nit: The flow is confusing here since we're mixing a `for` loop with two internal iterators and doing an allocation in the middle.

zilliqa/src/sync.rs
```diff
@@ -1327,3 +1322,76 @@ impl ToSql for PeerVer {
         Ok((self.clone() as u32).into())
     }
 }
+
+#[derive(Debug, Default)]
+struct SyncSegments {
+    headers: BTreeMap<Hash, BlockHeader>,
```
JamesHinshelwood, line 1328: Is this a `BTreeMap` for any reason? Do you need ordering of `Hash`es? Surely that isn't meaningful?

zilliqa/src/sync.rs
```diff
@@ -1327,3 +1322,76 @@ impl ToSql for PeerVer {
...
+    fn count_sync_segments(&self) -> usize {
+        self.segments.len()
+    }
+
+    /// Retrieves bulk metadata information from the given block_hash (inclusive)
+    fn get_sync_segment(&self, block: &BlockHeader) -> Vec<Hash> {
+        let mut result = vec![];
+
+        let mut hash = block.qc.block_hash;
+        // This implementation skips the final segment in the chain. I don't know if that's intended or not.
```
JamesHinshelwood, line 1348: I wrote this because I believe the `db.rs` implementation didn't iterate all of the returned results. Since you're the author it would be nice to find out if that is by-design or not and make this comment more assertive in either direction.

JamesHinshelwood (approved):
Some nits, nothing blocking. Thanks!

zilliqa/src/sync.rs
shawn-zil: OK. I'll break it up, and add a comment in a subsequent PR. There's a reason for the double-reverse.

zilliqa/src/sync.rs
shawn-zil: No, a HashMap would probably work just as well. I can switch to that in a subsequent PR.

zilliqa/src/sync.rs
shawn-zil: Ah, blame it on cut-n-paste. I'll comment that it is by-design, and why, in a subsequent PR.

chetan-zilliqa (approved):

PR #2570, opened 2025-03-24
Implement debug_traceTransaction

### LINKED ISSUE #2204: Implement debug_traceTransaction

maxconway:
- Closes #2204
- Adds docs
- Adds tests/it/debug.rs test file with placeholders
- Adds tests for debug_TraceTransaction
- Adds tests for debug_trace_block_by_number


saeed-zil (approved):
LGTM

zilliqa/tests/it/debug.rs
```diff
@@ -0,0 +1,143 @@
...
+        .expect("Failed to call debug_traceBlockByNumber API");
+
+    assert!(!response.is_empty());
+
+    // Test with callTracer
+    let tracer_options = serde_json::json!({
+        "tracer": "callTracer"
+    });
+
+    let response: Vec<TraceResult> = wallet
```
JamesHinshelwood, line 49: IMO just adding a test that asserts the API exists isn't really worth anything. I don't think its worth spending any test-time budget on this.

JamesHinshelwood (dismissed):

zilliqa/tests/it/debug.rs
maxconway: Just pushed a fix for this making sure the tests check that the correct trace type has occurred.
I like to have a test exercise the method to catch any panics even when there's nothing that particularly needs testing about the actual return value

JamesHinshelwood (approved):

PR #2576, opened 2025-03-25
feat: Remove high_block from decisions made at startup. Always use finalized

86667:


zilliqa/src/consensus.rs
```diff
@@ -233,39 +233,26 @@ impl Consensus {
...
-                        high_block.view(),
-                        finalized_number,
-                        start_view
+                        "recovery: finalized_view {0}, start_view {1}",
+                        finalized_view, start_view
                     );
 
-                    if finalized_number > high_block.view() {
-                        // We know of a finalized view higher than the view in finalized_number; start there.
-                        state.set_to_root(finalized_block.header.state_root_hash.into());
```
JamesHinshelwood, line 263: Should we keep this line?

JamesHinshelwood (approved):

zilliqa/src/consensus.rs
86667: Its sitll there I moved it up a few lines (:243)

PR #2579, opened 2025-03-25
#2578 improved pruning config

### LINKED ISSUE #2578: Not possible to write a configuration file that disables pruning
prune_interval is a u64, where u64::max() implies no pruning. However, toml only allows signed 64-bit values.

Attempting to write a config file with prune_interval = u64::MAX() causes the toml renderer to panic (and if you could write such a file, it would not be readable).

rrw-zilliqa:
Fixes #2578 

(I believe that `consensus::network_can_die_restart` fails independent of this patch - eg. for seed 6601524667959767463 at 14ed53aedbce24ca087d62164d91c461f165194d  - and so does `persistence::checkpoints_test`)

### DIFF
--- a/z2/src/setup.rs
+++ b/z2/src/setup.rs
@@ -552,7 +552,7 @@ impl Setup {
                 failed_request_sleep_duration: failed_request_sleep_duration_default(),
                 enable_ots_indices: false,
                 max_rpc_response_size: max_rpc_response_size_default(),
-                prune_interval: cfg::u64_max(),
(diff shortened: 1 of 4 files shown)


shawn-zil:
Would it make sense to just not write a toml file, to disable pruning i.e. leave it out of the config file? - it defaults to u64::MAX internally.

rrw-zilliqa:
Closing due refactor upstream.

PR #2584, opened 2025-03-26
feat: trigger sync when we see a high qc for a block we do not have

86667:
We trigger a block request probe upon seeing a high qc for a block we dont have to a single peer. If the peer chosen does not have the block either, then the next `NewView` will trigger the same code to probe the next peer.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -389,30 +389,6 @@ impl Consensus {
                 // Build NewView so that we can immediately contribute to consensus moving along if it has halted
                 consensus.build_new_view()?;
             }
-
-            // Remind block_store of our peers and request any potentially missing blocks
-            let high_block = consensus
-                .db
-                .get_block_by_hash(&high_qc.block_hash)?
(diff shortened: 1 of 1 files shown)


shawn-zil (dismissed):

zilliqa/src/consensus.rs
```diff
@@ -1754,6 +1730,12 @@ impl Consensus {
             return Ok(None);
         }
 
+        if let Ok(None) = self.get_block(&new_view.qc.block_hash) {
```
JamesHinshelwood, line 1733:
```suggestion
        if self.get_block(&new_view.qc.block_hash)?.is_none() {
```
Maybe I'm nitpicking, but the current version will silently ignore an `Err` from `get_block` which I don't think makes sense.

JamesHinshelwood (approved):

zilliqa/src/consensus.rs
86667: No that's fair. Updated

JamesHinshelwood (approved):

PR #2585, opened 2025-03-26
feat: add admin_forceSetView API

86667:


zilliqa/src/api/admin.rs
```diff
@@ -85,3 +86,13 @@ fn checkpoint(params: Params, node: &Arc<Mutex<Node>>) -> Result<CheckpointRespo
         block: block.number().to_hex(),
     })
 }
+
+fn force_set_view(params: Params, node: &Arc<Mutex<Node>>) -> Result<bool> {
+    let mut params = params.sequence();
+    let view: U64 = params.next()?;
+    let timeout_mins: U64 = params.next()?;
```
shawn-zil, line 93: instead of putting a duration, would it be better to input an exact future time-stamp e.g. `2025-12-12T00:00:00Z` or something like that? Otherwise, we would need to ensure that whatever script that fires this can hit the necessary nodes, within a fairly short amount of time.

JamesHinshelwood: I agree this is probably slightly cleaner, since even if we 'miss' the timeout all the nodes should be close to aligned on the subsequent views.

zilliqa/src/api/admin.rs
```diff
@@ -24,6 +24,7 @@ pub fn rpc_module(
             ("admin_consensusInfo", consensus_info),
             ("admin_generateCheckpoint", checkpoint),
             ("admin_blockRange", admin_block_range),
+            ("admin_forceSetView", force_set_view),
```
JamesHinshelwood, line 27: nit: I don't like that this has 2 verbs. Is `admin_forceView` or `admin_setView` okay?

zilliqa/src/consensus.rs
```diff
@@ -171,6 +171,8 @@ pub struct Consensus {
     pub new_transaction_hashes: broadcast::Sender<Hash>,
     /// Pruning interval i.e. how many blocks to keep in the database.
     prune_interval: u64,
+    /// Used for testing and test network recovery
+    force_set_view: Option<(u64, Duration)>,
```
JamesHinshelwood, line 175: `forced_view`?

zilliqa/src/consensus.rs
```diff
@@ -3208,6 +3221,21 @@ impl Consensus {
     pub fn get_sync_data(&self) -> Result<Option<SyncingStruct>> {
         self.sync.get_sync_data()
     }
+
+    /// This function is intended for use only by admin_forceSetView API. It is dangerous and should not be touched outside of testing or test network recovery.
+    ///
+    /// Force set our view and override exponential timeout value with given value for this view only
+    /// View value must be larger than current
+    pub fn force_set_view(&mut self, view: u64, timeout_mins: u64) -> Result<()> {
+        if self.get_view()? >= view {
```
JamesHinshelwood, line 3230: nit 
```suggestion
        if self.get_view()? > view {
```

if `view == self.get_view()?` everything is fine and the API is just saying "stay in this view for the time I'm telling you, not the time you've calculated".

zilliqa/src/consensus.rs
```diff
@@ -3208,6 +3221,21 @@ impl Consensus {
...
+            return Err(anyhow!(
+                "view cannot be forced into lower view than current"
+            ));
+        }
+        self.set_view(view)?;
+        self.force_set_view = Some((view, Duration::from_secs(timeout_mins * 60)));
```
JamesHinshelwood, line 3236:
```suggestion
        self.force_set_view = Some((view, Duration::from_mins(timeout_mins)));
```
but I guess this will change if we make @shawn-zil 's proposed change anyway.

86667: `Duration::from_` anything larger than secs is nightly

zilliqa/src/api/admin.rs
86667: Agreed :+1:

zilliqa/src/consensus.rs
86667: true

zilliqa/src/consensus.rs
```diff
@@ -610,6 +615,16 @@ impl Consensus {
             .block_time
             .saturating_sub(milliseconds_since_last_view_change);
 
+        // Override milliseconds_remaining_of_block_time in forced set view scenario
+        match self.force_view {
+            Some((forced_view, timeout_instant)) if view == forced_view => {
+                milliseconds_remaining_of_block_time = SystemTime(timeout_instant.to_system_time())
```
JamesHinshelwood, line 621: Can we add and use `impl From<DateTime> for zilliqa::time::SystemTime` in `time.rs` instead? I'm pretty sure this line of code won't compile when the `fake_time` is not enabled.

JamesHinshelwood (changes requested):
LGTM, but I think there's one issue with the time conversion.

zilliqa/src/time.rs
```diff
@@ -58,6 +59,12 @@ mod time_impl {
         }
     }
 
+    impl From<DateTime> for SystemTime {
```
JamesHinshelwood, line 62: Don't you need this when `#[cfg(not(feature = "fake_time"))]` too?

86667: The `impl From<DateTime> for SystemTime {` implementation already exists in `datetime`.

Confirmed with build with `--build-arg="is_release=true"`, that's correct right?

JamesHinshelwood: Ah sorry I missed that

JamesHinshelwood (approved):

PR #2586, opened 2025-03-27
Fix broken tests

shawn-zil:


maxconway (commented):
When I run cargo test on main, I only get `consensus::network_can_die_restart` failing.

When I run cargo test on this, I get:
```
consensus::network_can_die_restart
    zil::get_num_txns_ds_epoch_1
    zil::get_num_txns_tx_epoch_1
    zil::get_tx_rate_1
    zil::get_txns_for_tx_block_ex_1
```

Are these expected? Are they tests that were previously passing when they shouldn't have been?


Edit: I just noticed this is still a draft, ignore this if appropriate :)

maxconway:
Yeah i'm still getting additional test failures over main:

```
failures:

---- consensus::network_can_die_restart stdout ----

thread 'tokio-runtime-worker' panicked at zilliqa/tests/it/main.rs:618:67:
called `Result::unwrap()` on an `Err` value: missing block that high QC points to!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Reproduce this test run by setting ZQ_TEST_RNG_SEED=2958932326436504804

---- zil::get_num_txns_tx_epoch_1 stdout ----

thread 'tokio-runtime-worker' panicked at zilliqa/tests/it/zil.rs:2923:5:
assertion `left == right` failed
  left: 0
 right: 1
Reproduce this test run by setting ZQ_TEST_RNG_SEED=17326652032902044921

---- zil::get_tx_rate_1 stdout ----

thread 'tokio-runtime-worker' panicked at zilliqa/tests/it/zil.rs:2151:5:
Transaction block rate should be positive
Reproduce this test run by setting ZQ_TEST_RNG_SEED=15509853319132783021


failures:
    consensus::network_can_die_restart
    zil::get_num_txns_tx_epoch_1
    zil::get_tx_rate_1
```

Looks like the CI is getting additional ones.


From looking at get_num_txns_tx_epoch_1 as an example, what's going on is that previously the test transaction was added after block 1 and before block 2, so it's in block 2, which is where GetNumTxnsTXEpoch picks it up.

After the change, the transaction is added after 1 and before 8, presumably ending up in block 2, but when GetNumTxnsTXEpoch is run at block 8 it is no longer in the current tx epoch.

I can go through and add some fixes for stuff I see if you like?

maxconway:
I've pushed in fixes for some tests that were failing for me

shawn-zil:
@maxconway finally, passes tests.

maxconway (dismissed):
LGTM :)

PR #2614, opened 2025-04-07
Add transmission_size column to blocks table

shawn-zil:
This PR is not a blocker. It is probably a useful feature to have, before 1.0.0 launch. This PR reduces the need to compute the transmission-size in real-time, for syncing purposes, by storing the value in the DB.

I'm not completely happy with this 'lazy at sync' approach, but it is the simplest to implement with little impact on the rest of the code-base.


86667 (approved):

JamesHinshelwood:
Does `cbor4ii::serde::to_vec` actually take a meaningful amount of time that this is worth caching? My gut feeling is no.

shawn-zil:
> Does `cbor4ii::serde::to_vec` actually take a meaningful amount of time that this is worth caching? My gut feeling is no.

My concern is more on the `block_to_proposal()` than the serialization. So, trying to reduce how often the proposals need to be reconstructed, which includes pulling out the list of transactions.

shawn-zil:
Implemented an in-memory cache solution in #2661

PR #2630, opened 2025-04-09
Fix some slow SQL.

shawn-zil:


86667:
The benchmark report shows an decrease in performance because of the sql execution timings. We should revert that commit before merging.

86667 (approved):

PR #2632, opened 2025-04-09
feat: log long running queries

86667:


bzawisto (approved):

86667:
I expected a minuscule performance hit from this but it seems from the bencher that it is large. Best not to merge then.

PR #2646, opened 2025-04-11
Use redb instead of SQLite

JamesHinshelwood:
This replaces the persistent SQLite datastore with redb. The goal is to achieve faster read and write performance and more reliable query behaviour.

Our benchmarks indicate redb achieves this goal and I beleive the resulting interface is also easier to use.

When a node running this version starts for the first time, it will automatically migrate its data from SQLite to redb. Here are the timings for this process on my machine with full history. I would expect this to be much less for persistence files that do not include full history.

* `migrating blocks` - 1053 seconds
* `migrating receipts` - 1275 seconds
* `migrating transactions` - 613 seconds
* `migrating touched address index` - 1 second
* `migrating consensus info` - 0 seconds
* `migrating state trie` - 1232 seconds
* `committing writes` - 924 seconds
* `compacting database` - 827 seconds

Total time: 5925 seconds = 1 hour, 38 minutes, 45 seconds


shawn-zil (commented):
is there a reason we're reintroducing `block_store.rs` or is that an oversight?

JamesHinshelwood:
Thanks @shawn-zil , it was an oversight :)

86667 (dismissed):
:+1: :+1: :+1: 

The need for a `write.commit()` is bound to cost developers time down the line debugging when it is forgotten. I wonder if there are any potential disadvantages to including it in a custom `Drop` implementation.

zilliqa/src/node.rs
```diff
@@ -651,16 +659,22 @@ impl Node {
...
         let parent = self
             .get_block(block.parent_hash())?
             .ok_or_else(|| anyhow!("missing block: {}", block.parent_hash()))?;
         let mut state = self
             .consensus
             .state()
             .at_root(parent.state_root_hash().into());
-        if state.is_empty() {
-            return Err(anyhow!("State required to execute request does not exist"));
-        }
```
86667, line 663: We re-execute these txs on the state at parent, so do we not need to check that the parent's state exists

86667 (commented):
Did you remove the code which sets currently stored block's state root hash to zero:

```
    fn get_block(&self, filter: BlockFilter) -> Result<Option<Block>> {
        let Some(mut block) = self.get_transactionless_block(filter)? else {
            return Ok(None);
        };
        if self.executable_blocks_height.is_some()
            && block.header.number < self.executable_blocks_height.unwrap()
        {
            debug!("fetched ZQ1 block so setting state root hash to zeros");
            block.header.state_root_hash = Hash::ZERO;
        }
...
``` 

We could remove the need for this by instead writing it at migration

zilliqa/src/node.rs
JamesHinshelwood: Good point, thanks - I moved the check to `apply_transaction` instead which I think makes more sense.

shawn-zil (approved):
It's good to get this in first. I'll have to make a number of changes for syncing/pruning after this is merged.

shawn-zil (approved):

PR #2656, opened 2025-04-14
fix: Bug which checks for loops in forking code at the wrong time

86667:
This PR fixes the issue we've been seeing in which nodes get stuck with 2 blocks both marked `is_canoncial = FALSE` and so cannot move forward: We check for loops slightly too early in the forking logic.

Also fixed is handling of self proposals which fail to finalise. Without this code I saw nodes attempt and fail to propose their bad blocks repeatedly, failing in a way which will never become successful.

One day I'll rewrite this forking code. I'm sure it can be made clearer but its not immediately obvious exactly how.

This code was deployed to prototestnet validators at 8pm on 14th April.


JamesHinshelwood (approved):
Oops, good spot, thank you!

PR #2661, opened 2025-04-15
Implement passive sync

shawn-zil:


zilliqa/src/sync.rs
```diff
@@ -465,30 +585,190 @@ impl Sync {
...
+        &mut self,
+        from: PeerId,
+        request: RequestBlocksByHash,
+    ) -> Result<ExternalMessage> {
+        tracing::debug!(hash = %request.hash, count = %request.count, %from,
+            "sync::PassiveRequest : received",
+        );
+
+        // Check if we should service this request - https://github.com/Zilliqa/zq2/issues/1878
+        if self.is_validator {
```
JamesHinshelwood, line 620: Maybe its cleaner to just have a configuration variable which says whether you're willing to serve passing syncing responses?

Happy to change it later though.

zilliqa/src/sync.rs
```diff
@@ -161,12 +168,24 @@ impl Sync {
...
 
+        // If set, sync_base_height must be sane
+        if sync_base_height != u64::MAX && latest_block_number < sync_base_height {
+            return Err(anyhow::anyhow!("sync_base_height > highest_block"));
+        }
+
+        let zq1_ceil_height = config
+            .consensus
+            .get_forks()?
+            .find_height_fork_first_activated(crate::cfg::ForkName::ExecutableBlocks)
```
JamesHinshelwood, line 186: As discussed on Slack, I think this would be cleaner with a call to something like `forks.get(block.number).is_executable` when you receive a block.

JamesHinshelwood (approved):
LGTM, only minor comments.

JamesHinshelwood:
@shawn-zil I resolved my comments for now, so we can merge this - Feel free to ignore/fix in a follow-up.

PR #2664, opened 2025-04-15
feat: Create separate GossipSub topic for Validators

86667:
And Broadcast `NewView`s on that topic.

For backwards compatibility we'll have to send to both topics until the version after these changes are included.

### DIFF
--- a/zilliqa/src/p2p_node.rs
+++ b/zilliqa/src/p2p_node.rs
@@ -44,6 +44,8 @@ use crate::{
     sync::SyncPeers,
 };
 
+static VALIDATOR_TOPIC_SUFFIX: &str = "-validator";
+
 /// Messages are a tuple of the destination shard ID and the actual message.
 type DirectMessage = (u64, ExternalMessage);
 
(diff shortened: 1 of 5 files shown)


shawn-zil (approved):
I can understand the appeal of having a 'private' topic. My concern is that we should have a strict policy of what gets sent on this topic. Otherwise, we risk having N-topics for N-things.

shawn-zil (changes requested):
I believe that when a validator is first re/started, there will be a problem - it is unclear whether the node should:
1. automatically subscribe to the validator topic until proven otherwise; or 
2. should not subscribe until it is proven otherwise.

In (1) the node would be receiving validator only messages, rendering the topic as a public topic; and in (2) the node could possibly miss out on validator only messages, making it miss out on critical things.

Therefore, it is likely that the messages would need to be broadcast on both topics regardless.

shawn-zil (approved):
LGTM. As explained by @86667

86667:
@shawn-zil thanks for your comments.

The policy (which i'll document somewhere in the code) is:

1) Broadcasts are so sent to the public topic (`Proposal`)
2) Direct messages are not sent to any topic (`Vote`, `NewView`)
3) Re-sending of `NewView` is sent to the Validator-only topic 

So:
A node who is not in the committee on restart will receive 1).
A node who is in the committee on restart will receive 1) and 2), learn it is in the committee after a proposal then begin to receive 3).

This leaves a circumstance in which a node is restarted and in the committee, and whilst being down the network hasnt produced any blocks, the node then will not be able to participate in recovering the network for the current view. I'll change the default so that the node assumes it is in the committee on startup, then can remove itself when it learns that it isnt.

86667:
@shawn-zil changes pushed

zilliqa/src/p2p_node.rs
```diff
@@ -210,11 +240,14 @@ impl P2pNode {
...
-        self.shard_nodes.insert(topic.hash(), input_channels);
+        self.shard_peers.insert(shard_id, peers);
+        self.shard_nodes.insert(shard_id, input_channels);
         self.shard_threads
             .spawn(async move { node.start_shard_node().await });
-        self.swarm.behaviour_mut().gossipsub.subscribe(&topic)?;
+        self.swarm
+            .behaviour_mut()
+            .gossipsub
+            .subscribe(&Self::shard_id_to_topic(shard_id, None))?;
```
shawn-zil, line 250: Isn't this where the node is supposed to subscribe to the validator topic, upon a re/start?

86667: Yes, thanks. updated

JamesHinshelwood (approved):

PR #2686, opened 2025-04-24
Fix active-sync flaw.

shawn-zil:
Description in the issue ticket


shawn-zil:
Since it doesn't really present itself, I'll incorporate this into #2661

PR #2697, opened 2025-04-29
feat: DEVOPS-1991 not deploy faucet in mainnet

pavlops:
- Adding a enable_faucet variable in the chain definitions.
- Conditional deployment of faucet/spout resources.
- Separating the start/stop logic of the faucet from the apps.
- Tested with `z2 deployer install --select zq2-infratest.yaml`.

### DIFF
--- a/z2/src/chain.rs
+++ b/z2/src/chain.rs
@@ -117,7 +117,7 @@ pub enum Chain {
             subdomain = "zq2-mainnet.zilliqa.com",
             project_id = "prj-p-zq2-mainnet-sn5n8wfl",
             log_level = "zilliqa=trace",
-            enable_faucet = "true"
+            enable_faucet = "false"
         )
     )]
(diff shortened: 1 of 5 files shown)


z2/src/chain.rs
```diff
@@ -108,7 +116,8 @@ pub enum Chain {
             genesis_amount = "900_000_000_000_000_000_000",
             subdomain = "zq2-mainnet.zilliqa.com",
             project_id = "prj-p-zq2-mainnet-sn5n8wfl",
-            log_level = "zilliqa=trace"
+            log_level = "zilliqa=trace",
+            enable_faucet = "true"
```
frankmeds, line 120: enable_faucet here should be false

frankmeds (changes requested):

frankmeds (approved):

PR #2710, opened 2025-05-02
Removed extraneous price conversion

maxconway:
As per conversation with @bzawisto and @chetan-zilliqa, I'd got a multiplication here I shouldn't have


zilliqa/src/api/zilliqa.rs
```diff
@@ -758,13 +758,7 @@ fn get_txn_fees_for_block(node: &Node, hash: Hash) -> Result<u128> {
...
         .get_transaction_receipts_in_block(hash)?
         .iter()
-        .fold(0, |acc, txnrcpt| {
-            let txn = node
-                .get_transaction_by_hash(txnrcpt.tx_hash)
-                .unwrap()
-                .unwrap();
-            acc + ((txnrcpt.gas_used.0 as u128) * txn.tx.gas_price_per_evm_gas())
-        }))
+        .fold(0, |acc, txnrcpt| acc + txnrcpt.gas_used.0 as u128))
```
JamesHinshelwood, line 761: Does `get_txn_fees_for_block` return the fees paid (in ZIL) or the gas used? The return type of `u128` suggests it should be the fee in ZIL?

maxconway: Good shout @JamesHinshelwood , this was left over from previously. I've changed the type to match

chetan-zilliqa (approved):

PR #2714, opened 2025-05-05
Post-REDB issues

shawn-zil:


zilliqa/src/db/migrate.rs
```diff
@@ -484,14 +484,22 @@ impl Db {
...
         write.commit()?;
 
+        let path = self.path.clone().expect("valid path");
+        let db = self.into_raw();
+
         info!("compacting");
-        let path = self.path.clone();
-        let mut db = self.into_raw();
+        mem::drop(db);
+        let mut db = redb::Database::builder()
```
shawn-zil, line 492: I believe that compaction is unnecessary - it had no significant impact on the database size on API1-1 in prototestnet.

maxconway:
Looks like this breaks the log indices. What's do the log index changes here do?

shawn-zil:
> Looks like this breaks the log indices. What's do the log index changes here do?

The eager logs alter the memory structure of the Log Structs. These structs are serialised with bincode and stored as binary blobs in the DB. As a result, deserialization fails because the in-memory layout no longer matches the on-disk layout.

maxconway (approved):
Seems fine to me, if we get any knock on problems I'll fix them as they appear

PR #2724, opened 2025-05-06
Revert "Use redb instead of SQLite (#2646)"

86667:
This reverts commit 6d3e0de91d1b59e4a3d8a9d23ebb9fdff40ecf77.


shawn-zil (approved):
LGTM. But @JamesHinshelwood should have a look.

PR #2730, opened 2025-05-08
Minor passive-sync improvements

shawn-zil:
This PR adds 2 minor passive-sync improvements:
1. Downgrade, but do not drop failing peers. Otherwise, it eventually leads to no more peers.
2. Use a config flag to service passive-sync requests, which allows the user to decide if they want to block it e.g. for validators.


maxconway (approved):
Looks good as far as I'm concerned

PR #2731, opened 2025-05-08
Extend admin_getPeers() RPC

shawn-zil:
Extends the `admin_getPeers()` RPC to return the totality of peers `swarm_peers`, not just the peers used for syncing `sync_peers`. The peers used for syncing is a subset as it only contains peers that are part of the topic, and are still good. So, it won't match the numbers in `net_peerCount` which will match `swarm_peers` returned in this PR.

If combined with #2730 the sync peers are the known peers in the topic. So, there will be no need to return a separate `topic_peers`.


maxconway (approved):
No issues as far as I'm concerned

PR #2784, opened 2025-05-20
Block out-of-sync proposals

shawn-zil:
I chose this approach - to refuse/block the broadcast of the proposal - as it has the least impact on the rest of the code, as opposed to checking/failing at every step of the proposal assembly. So, a node is allowed to go thru the motions of assembling a proposal but will only broadcast a proposal if-and-only-if it is in-sync at the time of broadcast.

In addition, the proposer checks that the proposed block is not lagging the chain, to address the situation where the node might have been out-of-sync at the start of proposal assembly but caught up by the time it had to broadcast the proposal.

The proposal itself could very well be a fork, and that is dealt with elsewhere, as usual.

### DIFF
--- a/zilliqa/src/consensus.rs
+++ b/zilliqa/src/consensus.rs
@@ -1681,7 +1681,7 @@ impl Consensus {
         )))
     }
 
-    // Returns Ok(None) if the node cannot propose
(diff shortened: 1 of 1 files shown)


shawn-zil:
Closing this PR until it is proven to be an issue.

PR #2799, opened 2025-05-21
Performance enhancement to log filters

### LINKED ISSUE #2794: Speed up trace logs filter by avoiding unneeded block queries

maxconway:
@JamesHinshelwood hopefully this is what you were looking for?
This reintroduces the filters, with the log filter now adjusted to only gather logs for blocks that match both the filter criteria and haven't yet been returned, rather than gathering everything that matches the criteria then throwing away everything that's already been returned.

It also fixes what I believe was a bug in the previous implementation where the filter wasn't getting updated with the last block retrieved, which definitely wouldn't have been helping the performance.

I haven't added any benchmarking or performance testing, if there's something you'd like to see in that respect, let me know

Fixes #2794


JamesHinshelwood (approved):
LGTM  :)

PR #2800, opened 2025-05-22
added explicit begin-commit-rollback around execute_block().

shawn-zil:


86667 (approved):
:+1:

zilliqa/src/db.rs
```diff
@@ -673,13 +673,36 @@ impl Db {
         })
     }
 
+    pub fn begin(&self) -> Result<()> {
+        let db = self.db.lock().unwrap();
+        Ok(db.execute_batch("BEGIN")?)
```
JamesHinshelwood, line 678: Can we use `db.transaction()` instead which will ensure we can't accidentally leave the transaction open?

shawn-zil: I believe that would require refactoring much of `execute_block()` to make the borrow-checker happy. I can do that, unless there's a different way of doing it?

JamesHinshelwood: Shouldn't you be able to do something like acquire a `db.transaction()` at the start of `execute_block()` and remember to `.commit()` it (or whatever) on the happy paths?

shawn-zil: First thing I tried, and a number of others, but the borrow-checker complained.
So, I resorted to an explicit BEGIN-COMMIT-ROLLBACK pattern instead.

PR #2822, opened 2025-05-26
Remove write lock when calling filter APIs

JamesHinshelwood:
This pushes the lock down to the inner filter map, meaning we can remove the top-level write locks.

I also simplified to cleanup implementation in filters. Now we only delete expired filters when we are attempting to add a new filter. This is the only time we might need to reclaim some memory.


maxconway (approved):
Fine by me, I just pushed in a quick fix for a lint clippy complained about

shawn-zil (approved):

PR #2829, opened 2025-05-27
Reroute sync-requests to broadcast

shawn-zil:
To avoid HOL-blocking issues on the request queue, reroute sync-related messages to broadcasts.

### DIFF
--- a/zilliqa/src/node.rs
+++ b/zilliqa/src/node.rs
@@ -241,7 +241,12 @@ impl Node {
         Ok(node)
     }
 
-    pub fn handle_broadcast(&mut self, from: PeerId, message: ExternalMessage) -> Result<()> {
+    pub fn handle_broadcast(
+        &mut self,
+        from: PeerId,
+        message: ExternalMessage,
+        response_channel: ResponseChannel,
+    ) -> Result<()> {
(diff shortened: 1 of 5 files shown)


86667:
Sorry if I'm missing something, I dont see how this helps. If the `requests` channel is blocking on `Node` then it will block the broadcasts channel from `Node` also?

shawn-zil:
> Sorry if I'm missing something, I dont see how this helps. If the `requests` channel is blocking on `Node` then it will block the broadcasts channel from `Node` also?

It mitigates HOL-blocking.

Assuming that the incoming requests are Sync-Sync-Sync-Vote, in the present case, the node will process it in-order; and a Sync request could be quite time consuming since it needs to retrieve a bunch of full blocks. By re-routing things this way, the node can process it as Sync-Vote-Sync-Sync. Effectively, deprioritising Sync vs NewViews, Votes and Proposals.

We did a similar thing the other way around previously, by routing Proposals such that they aren't blocked by a bunch of NewTransactions.

86667 (dismissed):
This feels like quite a shoe-horn. I'm not against merging as-is for now but would much prefer a read-only queue be implemented right away. I'll look into doing that  if i get time

86667 (approved):

PR #2837, opened 2025-05-29
Speedup pending_transactions()

shawn-zil:
Refer to issue for details. These small changes result in the following speedup:
```
time cast rpc --rpc-url http://localhost:44201 txpool_status
{"pending":4000,"queued":1000}

real    0m0.029s
user    0m0.003s
sys     0m0.011s

```


maxconway (commented):
I'm a bit suspicious of limiting the number of pending transactions based on a timer - I'd have thought we'd want the pending transactions list to be deterministic

shawn-zil:
I'll leave this here @maxconway you may extend this, cherry pick bits into #2838 or abandon this. I leave that decision to you.

PR #2838, opened 2025-05-29
Yet another speeding up pending_transactions by reducing calls to get_account

maxconway:
This does roughly the same thing as #2837 in terms of reducing calls to the slow `state.get_account()`, I'm happy to go with either implementation.

I've also added two benchmarks for measuring the speed up. The results are as follows:

Before caching (as of afeaf9):
Benchmark completed: pending_transactions took 4.095512ms to execute.
After caching (as of 245fae0):
Benchmark completed: pending_transactions took 387.258µs to execute.

Clearly the root problem is somewhere in state.get_account but this makes things a bit better until that's fixed

### DIFF
--- a/zilliqa/src/pool.rs
+++ b/zilliqa/src/pool.rs
@@ -822,8 +822,6 @@ mod tests {
             "Benchmark completed: pending_transactions took {:?} to execute.",
             duration
         );
-
         Ok(())
     }
-
 }
--- a/zilliqa/src/pool.rs
+++ b/zilliqa/src/pool.rs
@@ -192,6 +192,8 @@ impl TransactionPool {
         // Keeps track of [account, cumulative_txns_cost]
         let mut tracked_accounts = HashMap::new();
 
+        let mut balances = HashMap::new();
+
         let mut ready = self.gas_index.clone();
 
         let mut pending_txns = Vec::new();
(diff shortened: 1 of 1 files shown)


zilliqa/src/pool.rs
```diff
@@ -192,6 +192,8 @@ impl TransactionPool {
         // Keeps track of [account, cumulative_txns_cost]
         let mut tracked_accounts = HashMap::new();
 
+        let mut balances = HashMap::new();
```
shawn-zil, line 195: Instead of having two HashMaps, you can also consider doing it the other way - decrementing and tracking balances-per-account instead.

maxconway: Yeah I'm happy to take that approach - this felt a little clearer to me but you're right there's something to be said for one less hashmap. I've pushed the change. 

One thing to take a look at is that I've put the line `tracked_balances.insert(txn.signer, balance.saturating_sub(tx_cost));` after `if tx_cost > balance {continue};`. That seemed right to me and passed the tests but it does differ from #2837 so I'm not sure if there's something I'm missing

shawn-zil: Since you've inserted the balance earlier in line-215, it's not an issue.

shawn-zil (approved):

PR #2839, opened 2025-05-29
Set Kademelia protocol name to include network name

86667:
Also includes network name in request-response and identify protocol names.

These changes will be followed by [these changes](https://github.com/Zilliqa/zq2/pull/2840/files) to remove the legacy kademlia. 

The nodes will then slowly clean out peers from the wrong network, or at least will not add anymore peers from different networks. We can consider a coordinated Kademlia routing table purge if this does not happen naturally.

### DIFF
--- a/config-single-node.toml
+++ b/config-single-node.toml
@@ -1,4 +1,5 @@
 p2p_port = 0
+network = "Localnet"
 
 [[nodes]]
 api_servers = [


zilliqa/src/p2p_node.rs
```diff
@@ -146,9 +161,14 @@ impl P2pNode {
...
-                    kademlia: kad::Behaviour::new(peer_id, MemoryStore::new(peer_id)),
+                    kademlia_legacy: kad::Behaviour::new(peer_id, MemoryStore::new(peer_id)),
+                    kademlia: kad::Behaviour::with_config(
+                        peer_id,
+                        MemoryStore::new(peer_id),
+                        kad::Config::new(kad_protocol.clone()),
+                    ),
                     identify: identify::Behaviour::new(
-                        identify::Config::new("zilliqa/1.0.0".into(), key_pair.public())
+                        identify::Config::new(protocol_version.clone(), key_pair.public())
```
86667, line 171: Protocol versions do not need to match for nodes to communicate - https://github.com/libp2p/specs/blob/master/identify/README.md#protocolversion

shawn-zil (approved):

shawn-zil:
I think that we missed something with the JS tests - they're unable to connect to the network (which is desirable, if they were incorrectly configured).

JamesHinshelwood (approved):

PR #2843, opened 2025-05-29
Do not process own broadcasts

86667:
This never felt correct to me. It feels like something which was implemented as a hacky fix for a less mature code base.


shawn-zil (approved):
Since we now explicitly handle our own Proposals, then it no longer requires re-broadcasting back to ourselves.

PR #2854, opened 2025-06-04
Rewrite of transaction pool for superior performance

maxconway:
Pool now keeps the pending transactions up to date with each change, rather than reconstructing it for each query. Internally, the pool is a set of pools for individual accounts, each of which maintains its own pending lists of nonced and nonceless transactions.

All the tests pass, but it is more difficult to test that consensus is interacting with the pool correctly, particularly that the pool is receiving new state when necessary.

In terms of speed, we've got about a 1564x speedup in pending transaction query time at the cost of about a 2x slowdown in insertion speed since we're now doing more work upfront, but insertion time is 1-2µs so this doesn't seem like it'll be slowing anything. If it turns out to be a problem there's more optimisation that can be done there. Either way it seems best to get this merged as is and then work on any further optimisations required


maxconway:
Some benchmarks of the changes:

## Insertion Benchmarks

| Operation | Old Version | New Version | Change |
|-----------|-------------|-------------|---------|
| **Insert 1M transactions (ascending nonces)** | 853ms | 1777ms | **2.1x slower** |
| **Insert 1M transactions (descending nonces)** | 1,046ms | 1886ms | **1.8x slower** |

## Query Benchmarks (1000 iterations on 256K transactions)

| Operation | Old Version | New Version | Change |
|-----------|-------------|-------------|---------|
| **Get all pending transactions** | 138.99s | 15.07s | **9.2x faster** |
| **Get pending transaction count** | 138.28s | 88.4ms | **1,564x faster** |
| **Get best transaction** | 761.6µs | 55.3µs | **13.8x faster** |

zilliqa/src/consensus.rs
```diff
@@ -834,7 +826,9 @@ impl Consensus {
                         let (_, txns, _, _, _) = early_proposal.take().unwrap();
                         let mut pool = self.transaction_pool.write();
                         for txn in txns.into_iter().rev() {
-                            pool.insert_ready_transaction(txn)?;
+                            let account = self.state.get_account(txn.signer)?;
+                            let added = pool.insert_transaction(txn, &account, false);
+                            assert!(added.was_added())
```
86667, line 831: There is a situation in which a transaction coming in as part of a proposal is not inserted into our mempool here. 
Ie:
- tx `A` received 
- tx `B` received with same nonce but larger `gas_price`
- Proposal received which includes tx `A`
Then `pool.insert_transaction` returns `TxAddResult::SameNonceButLowerGasPrice;` when called from here. I'm not sure of the knock-on effects of this but I presume something would fail later on because we must be adding transactions from proposals into the mempool for some reason.

maxconway: By my reading, that reinsertion of transactions from a failed early proposal was line 1683? Or is 1683 for if this node's proposal fails whereas 830 is for if another node's proposal fails?

Could 830 just be in case we receive a proposal with a transaction we haven't seen yet, we need to insert it?

Either way, it looks to me like it was a bug in the original implementation, no? If we receive a proposal with a new transaction that's inferior to one we already have, then even if the proposal as a whole is good we don't want to replace our transaction with the worse one?

zilliqa/src/pool.rs
```diff
@@ -43,270 +45,802 @@ impl TxAddResult {
...
+                    Some(result)
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        }
+    }
+    // Must be followed by maintain()
```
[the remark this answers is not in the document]
maxconway: It's so that in `update_with_account` we don't move transactions pending->queue during `update_balance` which we're only going to have to move queue-> pending during `update_nonce`.

Specifically, if we called it in the function itself, then in the scenario where we see new state which increases the nonce significantly but also decreases the balance (e.g. when an account sends a lot of transactions) , the following would occur:

1. balance decreases, decreasing `balance_after_pending`, so we move lots of transactions out of pending and into queued to get it positive again
2. nonce increases, so we remove lots of transactions from pending, meaning we increase `balance_after_pending` again because their gas no longer counts against it
3. because `balance_after_pending` is up again, we re-add the transactions to pending which we removed in step `

shawn-zil:
I've just noticed that there is a potential issue with our existing pool. Not sure if we want to do it here, or in an immediate follow up PR: https://github.com/Zilliqa/zq2/issues/2890

maxconway:
> I've just noticed that there is a potential issue with our existing pool. Not sure if we want to do it here, or in an immediate follow up PR: #2890

The suggestion of storing the encoded sizes in the pool feels nice, it'd require wrapping the transactions inside another structure with some metadata which is already something I wanted to do as an optimization. But I want to make sure we've got the txpool as is working correctly before I merge it in (or put in in a followup change)

maxconway:
Latest timings as of b45e156a:

Operation | Operations Count | Total Time | Time per Operation |
|-----------|------------------|------------|--------------------|
| Insert transactions (ascending nonces) | 236,544 | 1.000004459s | 4,227 ns/op |
| Insert transactions (descending nonces) | 242,739 | 1.000001744s | 4,119 ns/op |
| Get all pending transactions | 6,231 | 1.000118484s | 160,506 ns/op |
| Get pending transaction count | 16,226,874 | 1.000000108s | 61 ns/op |
| Get best transaction | 18,144,947 | 1.000000036s | 55 ns/op |
| Popping best transaction | 10,001 | 8.802878ms | 880 ns/op

shawn-zil (approved):
The size-limit edits LGTM.

PR #2872, opened 2025-06-10
[WIP] replace bincode with bitcode for SQL data.

shawn-zil:
just a quick test to see if there is a measurable impact


shawn-zil:
Based on the benchmark results, the performance improvement seems minimal ~5% and may not be worth the trouble.

PR #2908, opened 2025-06-17
A couple transaction pool write lock efficiency gains and a fix for clearing `early_proposal` without adding txs back to pool

86667:
- Update taking of a write lock until its required
- Take a single write lock per tx batch rather than taking a fresh one for each tx in the batch
- Fix a bug in which an `early_proposal` was cleared without its transactions being added back to the pool
- Some `consensus.rs` docs


maxconway (approved):
LGTM

shawn-zil (approved):

PR #2915, opened 2025-06-18
Changed GetLatestTxBlock to return latest finalized block

maxconway:
Plus I fixed the docs


JamesHinshelwood:
Will take on this change and fix the other APIs, as agreed.

PR #2921, opened 2025-06-19
Fixed benchmarks/integration-tests due to size limits

shawn-zil:
- Fixed the failing benchmark, that actually exceeded size limits.
- Added payload size checks into integration tests.


maxconway:
Looks like there's some issues with the benchmarking in CI, is this relevant?

86667 (approved):

PR #2930, opened 2025-06-23
Alter TxZilliqa serializer to keep it < 230 bytes

shawn-zil:
By keeping ZIL transfers < 230 bytes, we can squeeze 4000 of them into a single Proposal. It was previously taking too much space due to double-byte encoding.

I'll still need to make this a compatible change.
1. ~~compatible with cbor~~
2. ~~compatible with bincode~~
3. ~~cut-over height~~

Just need to pick a height to deploy this.


shawn-zil:
closed in favour of #3030

PR #2937, opened 2025-06-23
Additional tests for

maxconway:
- eth_getCode
- eth_getBalance

### DIFF
--- a/zilliqa/tests/it/eth.rs
+++ b/zilliqa/tests/it/eth.rs
@@ -1,6 +1,6 @@
 use std::{fmt::Debug, ops::DerefMut};
 
-use alloy::primitives::{Address, hex};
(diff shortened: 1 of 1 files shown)


86667 (commented):
There is a lot of overhead here for tests which dont need to be separate. Can we bring all these all into a single test network, or at least one for getBalance and another for getCode

maxconway:
> There is a lot of overhead here for tests which dont need to be separate. Can we bring all these into a single test

Now consolidated into two tests, one for eth_getBalance and one for eth_getCode. They're a bit over a second each on a debug build

I think that probably the correct next step would be to go for a single test fixture that builds a "typical" network that we can reuse across tests but that's probably beyond the scope of this PR

PR #2946, opened 2025-06-25
Added credits-based RPC rate-limiter

shawn-zil:
The global rate-limits would have to be managed using a hybrid GCP+APP level management. It is not feasible, otherwise.

IP-based connection limits are best managed at the GCP level, which has a global view on connectivity. However, GCP does not inspect the payload to count RPC call rate.

This PR attempts to bridge this gap, with a simple RPC per connection rate limit. So, for each connection that GCP lets through, it imposes a limit on how many RPC calls it makes. This also includes RPC batch calls as seen in the log output below.

This PR uses a credits-based mechanism. Each connection is allocated 500 (configurable) credits/second. Each RPC method is priced e.g. eth_blockNumber is priced at 80 credits. So, a single connection can make six eth_blockNumber calls/sec before hitting the rate limit (as seen in the log below).

Some calls are cheap e.g. net_version is priced at 5 credits, which allows 100 calls/sec and debug* calls are priced at 1000, meaning that they always hit the limit. Unknown (to me) calls are priced at the default value (e.g. 500) to allow for exactly 1 call/sec. It is expected that the price of all implemented calls will eventually be listed on the pricing table in `ratelimit.rs`

So, the global rate limit would be a combination of GCP + APP level limits e.g. 200 connections/minute/IP should result in around twenty (20) eth_blockNumber calls/sec on average, if it uses multiple RPC calls/connection. Otherwise, about three (3) eth_blockNumber calls/sec on average, per IP address.

```
POST / HTTP/1.0
Host: localhost:34201
Accept: */*
Accept-Encoding: gzip, deflate
User-Agent: Mozilla/5.0 (pc-x86_64-linux-gnu) Siege/4.0.7
Connection: keep-alive
Content-Type: application/json
Content-Length: 470

[{"jsonrpc":"2.0","id":"1","method":"eth_blockNumber","params": []},
{"jsonrpc":"2.0","id":"2","method":"eth_blockNumber","params": []},
{"jsonrpc":"2.0","id":"3","method":"eth_blockNumber","params": []},
{"jsonrpc":"2.0","id":"4","method":"eth_blockNumber","params": []},
{"jsonrpc":"2.0","id":"5","method":"eth_blockNumber","params": []},
{"jsonrpc":"2.0","id":"6","method":"eth_blockNumber","params": []},
{"jsonrpc":"2.0","id":"7","method":"eth_blockNumber","params": []}]

[{"jsonrpc":"2.0","id":"1","result":"0x31db"},
{"jsonrpc":"2.0","id":"2","result":"0x31db"},
{"jsonrpc":"2.0","id":"3","result":"0x31db"},
{"jsonrpc":"2.0","id":"4","result":"0x31db"},
{"jsonrpc":"2.0","id":"5","result":"0x31db"},
{"jsonrpc":"2.0","id":"6","result":"0x31db"},
{"jsonrpc":"2.0","id":"7","error":{"code":-32005,"message":"Limit exceeded"}}]
```

Currently deployed on `testnet` API nodes. Feel free to test it out.

### DIFF
--- a/config-single-node.toml
+++ b/config-single-node.toml
@@ -3,7 +3,7 @@ network = "Localnet"
 
 [[nodes]]
 api_servers = [
-    { port = 4201, enabled_apis = [ "admin", "debug", "erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa" ] }
+    { rate_credit = 0, port = 4201, enabled_apis = [ "admin", "debug", "erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa" ] }
 ]
 eth_chain_id = 0x8001
 consensus.scilla_address = "http://localhost:3000"
--- a/z2/src/chain/node.rs
+++ b/z2/src/chain/node.rs
@@ -681,7 +681,7 @@ impl ChainNode {
             json!({"port": 4201, "enabled_apis": [ { "namespace": "eth", "apis": ["blockNumber"] } ] })
         };
         // 4202 is not exposed, so enable everything for local debugging.
-        let private_api = json!({ "port": 4202, "enabled_apis": ["admin", "debug", "erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa"] });
+        let private_api = json!({ "rate_credit": 0, "port": 4202, "enabled_apis": ["admin", "debug", "erigon", "eth", "net", "ots", "trace", "txpool", "web3", "zilliqa"] });
         let api_servers = json!([public_api, private_api]);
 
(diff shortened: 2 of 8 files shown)


shawn-zil:
Currently breaks Otterscan as the `erigon_*` and `ots_*` prices for some operations hit the rate limit.

shawn-zil:
> Currently breaks Otterscan as the `erigon_*` and `ots_*` prices for most calls hit the rate limit.

Updated default credits to allow bursty Otterscan calls, up to a reasonable limit.

PR #2949, opened 2025-06-26
Minor log message upgrade

shawn-zil:
Make it obvious that the skip happened, rather than silence on the Log Explorer.


shawn-zil:
Closed, in favour of #2947

PR #2966, opened 2025-06-27
Fixed total coin supply APIs

maxconway:
- Adjusted amount down by 10^6, and subtracted null address balance
- Updated test
- Extra methods for ZilAmount

### DIFF
--- a/zilliqa/src/api/zilliqa.rs
+++ b/zilliqa/src/api/zilliqa.rs
@@ -1381,13 +1381,12 @@ fn get_total_coin_supply(_params: Params, node: &Arc<RwLock<Node>>) -> Result<St
         .get_account(Address::ZERO)
         .unwrap()
         .balance;
-    let null_address_balance_zil = ZilAmount::from_amount(null_address_balance);
-    let native_supply: ZilAmount = ZilAmount::from(node.config.consensus.total_native_token_supply);
-    Ok((native_supply - null_address_balance_zil).to_string())
+    let native_supply = node.config.consensus.total_native_token_supply.0;
+    Ok(ZilAmount::from_amount(native_supply - null_address_balance).to_string())
 }
 
 // GetTotalCoinSupplyAsInt
-fn get_total_coin_supply_as_int(_params: Params, node: &Arc<RwLock<Node>>) -> Result<u128> {
+fn get_total_coin_supply_as_int(_params: Params, node: &Arc<RwLock<Node>>) -> Result<ZilAmount> {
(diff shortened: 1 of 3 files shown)


zilliqa/src/transaction.rs
```diff
@@ -914,6 +927,12 @@ impl FromStr for ZilAmount {
     }
 }
 
+impl From<crate::cfg::Amount> for ZilAmount {
```
JamesHinshelwood, line 930: We deliberately didn't include this to force the caller to use `from_amount`, where they will hopefully acknowledge the caveats around rounding.

zilliqa/src/transaction.rs
```diff
@@ -894,6 +899,14 @@ impl Add for ZilAmount {
     }
 }
 
+impl Sub for ZilAmount {
```
JamesHinshelwood, line 902: This isn't used?

zilliqa/src/api/zilliqa.rs
```diff
@@ -1373,13 +1373,31 @@ fn get_num_txns_ds_epoch(_params: Params, node: &Arc<RwLock<Node>>) -> Result<St
...
+    let finalized_block_number = node.get_finalized_block_number()?;
+    let null_address_balance = node
+        .consensus
+        .state_at(finalized_block_number)?
+        .unwrap()
+        .get_account(Address::ZERO)
+        .unwrap()
+        .balance;
+    let native_supply = node.config.consensus.total_native_token_supply.0;
+    Ok(ZilAmount::from_amount(native_supply - null_address_balance).to_string())
```
JamesHinshelwood, line 1385: It might make sense to just call `get_total_coin_supply_as_int` here and de-duplicate?

zilliqa/src/transaction.rs
```diff
@@ -872,6 +872,11 @@ impl ZilAmount {
         self.0.checked_mul(10u128.pow(6)).expect("amount overflow")
     }
 
+    /// Get the underlying ZIL amount in units of (10^-12) ZILs.
+    pub fn into_inner(self) -> u128 {
```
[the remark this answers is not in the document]
JamesHinshelwood: If we aren't using this, we should remove it.

zilliqa/src/transaction.rs
maxconway: Done

zilliqa/src/api/zilliqa.rs

JamesHinshelwood (approved):

PR #2984, opened 2025-07-03
Build missing ots indices

bzawisto:
This PR adds a background thread which creates missing ots indices for transactions converted from zq1. It's called only once (by checking entry in new aux table).


zilliqa/src/consensus.rs
```diff
@@ -393,6 +393,8 @@ impl Consensus {
             // treat genesis as finalized
             consensus.set_finalized_view(latest_block_view)?;
             return Ok(consensus);
+        } else {
+            aux::check_and_build_ots_indices(db, latest_block_view)?;
```
JamesHinshelwood, line 397: Should we only call this if `enable_ots_indices` is true?

JamesHinshelwood (approved):
Some docs around what the batch job is doing might be useful in future too.

JamesHinshelwood (dismissed):

Issue #2693, opened 2025-04-28
Verify ZQ2 transactions


shawn-zil:
Some ZQ1 transactions in `prototestnet` fail verification e.g.
```
<google-sheets-html-origin><style type="text/css"><!--td {border: 1px solid #cccccc;}br {mso-data-placement:same-cell;}--></style>
999616 | 0 | d6d64bc4e43b26430a464ed6504b583506355b587ae88019ff334eb818255c1f |   |   |   |  
-- | -- | -- | -- | -- | -- | --
999577 | 0 | 062d22e8ddb2bf04f7c2f767fa537ce2978486d76caf41ce1aa7f77115da2cae |   |   |   |  
999379 | 0 | 2bdc0474a02ad3c83f900633d0e26664401fe9cdc7764ed9e3ed5d56e09f3c1b |   |   |   |  
999375 | 0 | 56b5fa44e1ff0358c4b70d64d77488bb518dff92922abb39a1983e3e13717b29 |   |   |   |  
999371 | 0 | 27b51b001a657c0eca7b95832f37b3bbb8714bf6ec51e97488d1eecfb72038cd |   |   |   |  
999360 | 0 | dc7357c04f2e52eed82a010265900df1ad703c90a190337daaf46f900312c63d |   |   |   |  
999341 | 0 | bc383594ea9511653a6e5528613083c1b8a44e6e6f81eb78cd13de7a1c5884e6 |   |   |   |  
997969 | 0 | c3294e8ab9d3a14837e03762971fcbaee0cd31fbe0500c160ed30138d5bf4946 |   |   |   |  
997949 | 0 | f264c5088c4a4349272b1899e0cb3ad9bfa1f5e013f906fc0b38a37207eb6f88 |   |   |   |  
962780 | 0 | 36e747c1b0b904b003dcb9f00f81ba69839d6a41e61ba780e818d563471eb083 |   |   |   |  
960159 | 0 | b45e204e8fee5884ba84a1d5a9bfef555c1f5caf08a28f20e616add9fb2bb7ed |   |   |   |  
960155 | 0 | 01c0642756d8f26273d02c12bf920bd657a04707d92ebe37f5bab8a0240c0bc0 |   |   |   |  
933768 | 0 | 4b4d472a6f3eb0f8ceb2dcd566b4d261b7d81710aed58805e319d0c7754b94e8 |   |   |   |  
[...]

PR #2986, opened 2025-07-03
Moved db wrapper to accepting and returning verified transactions

maxconway:
- actual verification is skipped on retrieval
- Should fix error when retrieving some ZQ1 transactions
- I haven't actually changed what's stored in the db, that could come later if we like this approach

Let me know if this doesn't work for some reason - it just seemed nicer and more consistent than adding one-off ignoring verification

### DIFF
--- a/zilliqa/src/transaction.rs
+++ b/zilliqa/src/transaction.rs
@@ -382,6 +382,14 @@ impl SignedTransaction {
     }
 
     pub fn verify(self) -> Result<VerifiedTransaction> {
+        self.verify_inner(false)
+    }
+
+    pub fn verify_override(self) -> VerifiedTransaction {
+        self.verify_inner(true).unwrap()
+    }
+
+    fn verify_inner(self, force: bool) -> Result<VerifiedTransaction> {
         let (tx, signer, hash) = match self {
             SignedTransaction::Legacy { tx, sig } => {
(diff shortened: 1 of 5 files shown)


shawn-zil:
I actually think that the correct solution involve fixing the ZQ1 verification. I've not had time to look into it yet, tho.
https://github.com/Zilliqa/zq2/issues/2693

So, I feel that a more localised bypass - where needed - might serve as a temporary workaround.

zilliqa/src/transaction.rs
```diff
@@ -382,6 +382,14 @@ impl SignedTransaction {
     }
 
     pub fn verify(self) -> Result<VerifiedTransaction> {
+        self.verify_inner(false)
+    }
+
+    pub fn verify_override(self) -> VerifiedTransaction {
+        self.verify_inner(true).unwrap()
```
shawn-zil, line 389: Probably a bad idea to `unwrap()` since there are other things that may return errors in `verify_inner()`.

shawn-zil (dismissed):
I had thought that this PR changes the on-disk format but that's not the case. It merely bypasses the verification checks for transactions loaded from disk, with the implicit assumption that the transactions in the DB are valid; and should never have made it in otherwise.

PR #3013, opened 2025-07-10
fix: Hardfork revert_restore_xsgd_contract

86667:


shawn-zil (commented):
will need to add hardforks to z2/src/chain.rs as well.

shawn-zil (commented):
Would this benefit from having a test as well?

PR #3014, opened 2025-07-11
Disable EIP-3607, for ZQ1 compatibility.

shawn-zil:
ZQ1 used SputnikVM (London) which did not have EIP3607 support. ZQ2 uses REVM (Shanghai) which has EIP3607 turned on by default.

This PR selectively disables EIP3607 for calls from `eth_call` and `eth_estimateGas` to emulate the previous ZQ1 behaviour.


maxconway (approved):
LGTM

PR #3033, opened 2025-07-15
fix: eth_getAccount should return code_hash rather than code

86667:


zilliqa/src/api/eth.rs
```diff
@@ -1025,7 +1026,13 @@ fn get_account(params: Params, node: &Arc<RwLock<Node>>) -> Result<Account> {
     let block = node.get_block(block_id)?;
     let block = build_errored_response_for_missing_block(block_id, block)?;
 
-    node.get_state(&block)?.get_account(address)
+    let account = node.get_state(&block)?.get_account(address)?;
+    Ok(GetAccountResult {
+        balance: account.balance,
+        nonce: account.nonce,
+        storage_root: account.storage_root,
+        code_hash: keccak256(serde_json::to_string(&account.code)?),
```
shawn-zil, line 1034: should it be a hash on the JSON representation? or a hash on the byte content?

86667:
Okay this is a better solution, I've hash'd the `code` which is returned from `get_code`.

zilliqa/src/api/eth.rs
```diff
@@ -1025,7 +1026,21 @@ fn get_account(params: Params, node: &Arc<RwLock<Node>>) -> Result<Account> {
     let block = node.get_block(block_id)?;
     let block = build_errored_response_for_missing_block(block_id, block)?;
 
-    node.get_state(&block)?.get_account(address)
+    let account = node.get_state(&block)?.get_account(address)?;
+    let return_code = if account.code.is_eoa() {
+        vec![].to_hex()
```
shawn-zil, line 1031: Does the keccak of this return the well known  KECCAK_EMPTY value?

shawn-zil (approved):

PR #3042, opened 2025-07-17
Make an exception for scilla precompile in view/estimate calls

bzawisto:
This is an extension to previous fix - we shouldn't make entire run of  eth_call/eth_estimate as failed if the conditions are met. In addition, I reduced the scope the extra_opts are passed further down.


shawn-zil (approved):
LGTM.

PR #3049, opened 2025-07-18
local docker network was broken.

shawn-zil:
If you start the local docker network and try to use the faucet to send yourself some funds, it will fail. This fixes things.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -3,7 +3,7 @@ version: "3"
 services:
   node0:
     depends_on:
-      - node3 
+      - node3
     environment:
       RUST_BACKTRACE: 1
       RUST_LOG: zilliqa=info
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -86,14 +86,14 @@ services:
   faucet:
     depends_on:
       - node0
-    image: asia-docker.pkg.dev/prj-p-devops-services-tvwmrf63/zilliqa/eth-spout:v1.3.72
+    image: asia-docker.pkg.dev/prj-p-devops-services-tvwmrf63/zilliqa/eth-spout:v1.3.84
     environment:
-      RPC_URL: http://node0:4201
+      RPC_URL: http://198.51.100.101:4201
```
maxconway, line 91: It looks like this changes from pointing to node0 by name to node1 by address,, is that what you were going for?

shawn-zil: So I picked a different node, since node0's port is exposed to the host machine.

maxconway (approved):
LGTM

PR #3072, opened 2025-07-25
Changed returned chainid to 0x8001 for Zilliqa transactions

maxconway:


zilliqa/src/transaction.rs
```diff
@@ -664,7 +664,7 @@ impl Transaction {
             Transaction::Legacy(TxLegacy { chain_id, .. }) => *chain_id,
             Transaction::Eip2930(TxEip2930 { chain_id, .. }) => Some(*chain_id),
             Transaction::Eip1559(TxEip1559 { chain_id, .. }) => Some(*chain_id),
-            Transaction::Zilliqa(TxZilliqa { chain_id, .. }) => Some(*chain_id as u64),
+            Transaction::Zilliqa(TxZilliqa { .. }) => Some(0x8001),
```
DrZoltanFazekas, line 667: please change it to `chain_id + 0x8000` to make it work on all Zilliqa networks

DrZoltanFazekas (changes requested):

zilliqa/src/transaction.rs
maxconway: Done

DrZoltanFazekas (approved):

PR #3074, opened 2025-07-28
Store sync metadata on-disk

shawn-zil:
This PR moves all the sync metadata on-disk, enabling active-sync across a larger range of blocks.

It uses a separate DB to avoid further fragmenting the existing DB. The sync db will be sporadically creating, deleting, records, etc. Since this DB is specific for syncing use, and does not have any specific performance requirements, I used a DB that was more ergonomic i.e. has a BTree-like interface. The choice of SLED was arbitrary.

It has been tested by syncing ~500K blocks on `devnet` that consumed < 500MB on-disk space.

The actual code changes are fairly small - mainly in the `SyncSegments` struct. But this PR also includes some minor cleanups:
1. Removed redundant "sync::" log message prefix.
2. Removed deprecated PeerVer.

### DIFF
--- a/zilliqa/src/sync.rs
+++ b/zilliqa/src/sync.rs
@@ -360,7 +360,7 @@ impl Sync {
                 self.request_missing_headers()?;
             }
             // Continue phase 2, until we have all segments.
-            SyncState::Phase2(_) if self.in_pipeline < self.max_blocks_in_flight => {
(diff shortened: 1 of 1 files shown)


maxconway (approved):
I don't love adding in a new db but I do agree that it makes sense in the current situation, so LGTM

PR #3080, opened 2025-07-30
Implement jailing of faulty proposers

DrZoltanFazekas:
Suspend a validator as leader if it failed to propose a block in 3 of the last 100 views and select another leader.


zilliqa/src/contracts/deposit_v5.sol
```diff
@@ -220,11 +220,34 @@ contract Deposit is UUPSUpgradeable {
 
     function leaderAtView(
         uint256 viewNumber
-    ) public view returns (bytes memory) {
-        uint256 randomness = uint256(
-            keccak256(bytes.concat(bytes32(viewNumber)))
-        );
-        return leaderFromRandomness(randomness);
+    ) public view returns (bytes memory stakerKey) {
```
shawn-zil, line 223: would this require a `deposit_v6.sol` or just a change to `deposit_v5.sol`?

DrZoltanFazekas: definitely, just took this "shortcut" for local testing, will introduce `deposit_v6.sol` later

maxconway (approved):
LGTM

PR #3086, opened 2025-07-31
Added an expiry mechanism to the transaction pool

### LINKED ISSUE #3040: TXpool needs expiry mechanic
Transactions older than 24h should be dropped

maxconway:
First of all, I added `delete_transactions` functions to TransactionPoolCore and TransactionsAccount

Then, I added the transaction expiry mechanism itself It's fairly approximate but should be good enough.

The mechanism is:

1. `hash_to_txn_map` is augmented to record the time when transactions were added to the pool
2. When the first transaction is added to the pool, `oldest_insertion_time` is set
3. When `oldest_insertion_time` is older than `expiry_time_clearout_threshold` (currently 48h), a clearout is initiated on the next add_txn()
4. During the clearout:
  5. All transactions older than `pool_expiry_time_minimum` (currently 24h) are deleted
  6. `oldest_insertion_time` is updated to the new oldest transaction

The end result should be that every 24h, we clear out everything older than 24h

My first pass at this I tried to implement something that would delete all transactions the moment they passed the 24h threshold in O(1). However, to do that I needed to package the creation time up inside a wrapper round the transactions which made everything much more long and complicated, which is why I instead went for this approximate approach

Closes #3040

### DIFF
--- a/zilliqa/src/pool.rs
+++ b/zilliqa/src/pool.rs
@@ -422,6 +422,40 @@ impl TransactionsAccount {
             })
     }
 
+    #[allow(clippy::collapsible_else_if)] // Clearer semantics
+    fn delete_transactions<I>(&mut self, txns: I) -> Vec<VerifiedTransaction>
+    where
+        I: IntoIterator<Item = VerifiedTransaction>,
+    {
+        let mut recalculate_needed = false;
+        let mut result = Vec::new();
+        for txn in txns {
+            if let Some(nonce) = txn.tx.nonce() {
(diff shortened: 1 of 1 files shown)


shawn-zil (commented):
Is it possible to add a test for this? Should be possible to advance the fake time by 24h to check.

shawn-zil:
My opinion is that an 'approximate' version is good enough for its intended purpose, if the alternatives are too complex.

maxconway:
> Is it possible to add a test for this? Should be possible to advance the fake time by 24h to check.

Done. That was a good call - I caught a bug

PR #3100, opened 2025-08-05
fix failing benchmark

shawn-zil:
Standardised some constants, so that they're same across different parts of the program.


maxconway (approved):
LGTM

PR #3123, opened 2025-08-14
Fix for incorrect receipt format in ZQ1 api responses

maxconway:
Modified GetTxnBodiesForTxBlock and GetTxnBodiesForTxBlockEx to use correct return format

Previously these had a different receipt format to GetTxResponse etc

Now it's unified in the ReceiptResponse type


shawn-zil (approved):
LGTM

Issue #3114, opened 2025-08-12
Improve slow API nodes


shawn-zil:
General finding: timeouts at gateway due to *slow* API nodes, results in `504` responses, and freezes, that result in a subsequent sync to catch up once the operation is completed.

Find ways to speed things up, generally, to mitigate the issue.

shawn-zil:
Early tests on `devnet` with #3126 seems promising - doubled txns/block for simple EVM transfers. More testing needed with other txn types before merging the feature.

shawn-zil:
Initial ERC20 baseline tests:

SQL

```
Block        | View         | Time                      | TXs   | Gas Used / Limit (%)                | Time diff  | Total TXs
------------------------------------------------------------------------------------------------------------------------------
495046       | 495056       | 2025-08-20 03:20:58 UTC   | 22    | 545104     / 84000000   (  0.65%)   | 1          | 4294
495047       | 495057       | 2025-08-20 03:21:01 UTC   | 0     | 0          / 84000000   (  0.00%)   | 3          | 4294
495048       | 495058       | 2025-08-20 03:21:01 UTC   | 134   | 3320108    / 84000000   (  3.95%)   | 0          | 4428
495049       | 495059       | 2025-08-20 03:21:06 UTC   | 46    | 1139740    / 84000000   (  1.36%)   | 5          | 4474
495050       | 495060       | 2025-08-20 03:21:07 UTC   | 61    | 1511398    / 84000000   (  1.80%)   | 1          | 4535
495051       | 495061       | 2025-08-20 03:21:09 UTC   | 166   | 4113016    / 84000000   (  4.90%)   | 2          | 4701
495052       | 495062       | 2025-08-20 03:21:11 UTC   | 118   | 2923684    / 84000000   (  3.48%)   | 2          | 4819
495053       | 495063       | 2025-08-20 03:21:14 UTC   | 60    | 1486656    / 84000000   (  1.77%)   | 3          | 4879
495054       | 495064       | 2025-08-20 03:21:16 UTC   | 22    | 545068     / 84000000   (  0.65%)   | 2          | 4901
495055       | 495065       | 2025-08-20 03:21:17 UTC   | 122   | 3022796    / 84000000   (  3.60%)   | 1          | 5023
495056       | 495066       | 2025-08-20 03:21:18 UTC   | 201   | 4980162    / 84000000   (  5.93%)   | 1          | 5224
[...]

PR #3126, opened 2025-08-18
Migrate state_trie to RocksDB

shawn-zil:
This PR proposes to migrate the `state_trie` to RocksDB, as a mitigation for #3114 

This has a couple of possible benefits:
1. RocksDB is a specialised key-value store, which should perform better than SQLite.
2. It should reduce DB locking by moving some requests to RocksDB instead of SQLite.

Existing state will be migrated using two mechanisms:
1. Lazily whenever a value is retrieved that does not exist in RocksDB, it will be migrated from SQLite.
2. Actively by setting the `active_state_migration = true` config setting (default: false), which will sequentially migrate state going backwards in history by loading each state and triggering (1) above. Active migration can be heavy-duty as it force loads the entire state trie, which is why it is disabled by default.

The hope is an overall improvement in node performance - this still needs to be fully tested, but initial testing on `devnet` looks promising - https://github.com/Zilliqa/zq2/issues/3114#issuecomment-3218690996

This feature should be pushed out with `v0.19.0` to allow for more testing.

### DIFF
--- a/z2/src/plumbing.rs
+++ b/z2/src/plumbing.rs
@@ -404,7 +404,6 @@ pub async fn run_persistence_converter(
         // This is None because it makes no difference to the conversion: var is required for fetching ZQ1 blocks and setting their state root hash to zero
         None,
         zilliqa::cfg::DbConfig::default(),
-        false,
     )?;
     let zq1_db = zq1::Db::new(zq1_dir)?;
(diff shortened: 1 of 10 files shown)


maxconway (approved):
I like this based on the conversation we had the other day, but we should check the performance in due course

maxconway (approved):

PR #3146, opened 2025-08-29
Fix possible state-sync issue

shawn-zil:
This PR improves state-sync by removing the DashSet, which stores all past state changes. This DashSet is unnecessary as we only need to detect a state change, which only requires a difference between the current state, and previous state. So, we only store the `last_hash` instead of the full history of changes. 

This is done to avoid possible (not actual) OOM issues.


maxconway (approved):
LGTM

PR #3153, opened 2025-09-01
fix benchmarks error.

shawn-zil:


shawn-zil:
Force merging, since this just fixes the CI/benchmark.

PR #3155, opened 2025-09-01
Update revm

bzawisto:
This PR introduced newer version of revm. There's also a dependency updated (eth-trie.rs) which we vendored and keep as submodule (see my two commits here: https://github.com/Zilliqa/eth-trie.rs/commits/zilliqa-revm/)

Don't merge until we confirm node can be synced.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -404,7 +404,7 @@ checksum = "64b728d511962dda67c1bc7ea7c03736ec275ed2cf4c35d9585298ac9ccf3b73"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.106",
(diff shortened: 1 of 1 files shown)


shawn-zil (commented):
You should reformat `zilliqa/tests/it/contracts/BytesArray.sol` to pass CI/CD.
```
npx prettier --write --plugin=prettier-plugin-solidity './zilliqa/**/*.sol'
```

Otherwise, LGTM overall. Let's test it properly.

shawn-zil (approved):

shawn-zil (changes requested):
Setting this to prevent accidental merge.

shawn-zil (approved):
LGTM.

PR #3182, opened 2025-09-09
Some performance improvements to a few api handlers

bzawisto:


zq2-devnet.yaml
```diff
@@ -9,4 +9,4 @@ roles:
   - persistence
   - private-api
 versions:
-  zq2: v0.18.0
+  zq2: v0.18.1
```
shawn-zil, line 12: Note: if we want to release a `v0.18.1` then the changes should be merged into the `release/v0.18.0` branch; and tagged from there as `main` already contains other breaking changes.

PR #3184, opened 2025-09-10
Replace `unsafe` code with `ArcSwap` crate

shawn-zil:
This PR is to improve code hygiene. Replace use of `unsafe` code with the ArcSwap crate.


maxconway (approved):
LGTM

PR #3187, opened 2025-09-10
feat: DEVOPS-2231 trace logs enabled for specific nodes

pavlops:
if NODE_NAME in [
    "zq2-testnet-api-ase1-0",
    "zq2-testnet-validator-ase1-0",
    "zq2-mainnet-api-ase1-0",
    "zq2-mainnet-bootstrap-ase1-0",
    "zq2-mainnet-checkpoint-ase1-0",
    "zq2-mainnet-persistence-ase1-0",
    "zq2-mainnet-apps-ase1-0",
] set zilliqa trace log level for specific nodes, if not use the default.

Compatible with z2 and Ansible installations.


maxconway (approved):
TBH I suspect in future we'll want to turn off debug logging entirely but this seems good for now :)

PR #3209, opened 2025-09-23
Release/v0.18.0 ex - Wip

bzawisto:


zilliqa/src/db.rs
```diff
@@ -580,8 +586,8 @@ impl Db {
                 if let rusqlite::trace::TraceEvent::Profile(statement, duration) = profile_event {
                     let statement_txt = statement.expanded_sql();
                     let duration_secs = duration.as_secs();
-                    if duration_secs > 5 {
-                        tracing::warn!(duration_secs, statement_txt, "sql execution took > 5s");
+                    if duration_secs > 1 {
```
shawn-zil, line 589: Use of `as_millis() > 1000` would be more accurate than `as_secs() > 1`, which is `2s`.

zilliqa/src/node.rs
```diff
@@ -100,18 +101,19 @@ impl MessageSender {
         Ok(())
     }
 
-    pub fn next_request_id(&mut self) -> RequestId {
-        let request_id = self.request_id;
-        self.request_id.0 = self.request_id.0.wrapping_add(1);
-        request_id
+    pub fn next_request_id(&self) -> RequestId {
```
shawn-zil, line 104: we might be able to get rid of any locking entirely, if we return a random instead of sequential number.

shawn-zil (dismissed):

zilliqa/src/db.rs
```diff
@@ -577,17 +579,21 @@ impl Db {
...
                     let statement_txt = statement.expanded_sql();
-                    let duration_secs = duration.as_secs();
-                    if duration_secs > 1 {
-                        tracing::warn!(duration_secs, statement_txt, "sql execution took > 1s");
+                    let query_duration = duration.as_millis();
+                    const DURATION_TIME_THRESHOLD_MS: u128 = 1000;
+                    if query_duration > DURATION_TIME_THRESHOLD_MS {
+                        warn!(
+                            DURATION_TIME_THRESHOLD_MS,
+                            statement_txt, "sql execution took > {DURATION_TIME_THRESHOLD_MS}"
```
shawn-zil, line 595: This should be `duration_secs` to record the actual time it took.

PR #3236, opened 2025-10-01
Call do_migration with sleep_time in NodeLauncher

shawn-zil:
State migration was only previously triggered upon consensus timing out. So, it would rarely get triggered and migration will be slow. Moved it so that it gets triggered upon both normal consensus processing; and during timeouts.


maxconway (approved):
LGTM

PR #3240, opened 2025-10-02
Adjust passive-sync logic to handle blocks with missing transactions

shawn-zil:
A previous fix https://github.com/Zilliqa/zq2/pull/3145 attempted to solve the issue of checkpoint parents missing transactions, by simply re-syncing the lowest block, at all times - forcibly syncing any checkpoint parent.

This causes an issue with a rare case, where only 1 block is able to fit into a response payload (either due to size or time). In such a case, it is possible for the passive-sync to get stuck repeatedly syncing only the lowest block. Re-syncing the lowest block is needed only for the checkpoint parent, which is stored without any associated transactions.


maxconway (approved):
LGTM

PR #3247, opened 2025-10-06
Integrate jailing into state-syncing

DrZoltanFazekas:
When jailing gets activated, it should also support state-syncing, i.e. replaying passive synced blocks to reconstruct their state. This requires building up a temporary history of missed views as blocks are being replayed, which gets merged with the node's primary missed view history.


zilliqa/src/db.rs
```diff
@@ -486,6 +486,80 @@ impl Db {
...
+    }
+
+    pub fn reset_ckpt_view_history(&self) -> Result<()> {
+        self.pool
+            .get()?
+            .prepare_cached("DELETE FROM ckpt_view_history WHERE leader NOT NULL")?
+            .execute([])?;
+        self.pool
+            .get()?
+            .prepare_cached(
```
shawn-zil, line 526: Please use a more structured e.g. `format!("UPDATE... {LARGE_OFFSET} ...");`.

zilliqa/src/db.rs
```diff
@@ -486,6 +486,80 @@ impl Db {
...
         }
 
+        if version < 6 {
+            connection.execute_batch(
+                ("
+                BEGIN;
+
+                INSERT INTO schema_version VALUES (6);
+
+                CREATE TABLE IF NOT EXISTS ckpt_view_history (view INTEGER NOT NULL PRIMARY KEY, leader BLOB) WITHOUT ROWID;
```
shawn-zil, line 496: Since `leader` is used in `WHERE` clause elsewhere, shouldn't it be indexed to prevent table SCAN?

zilliqa/src/db.rs
```diff
@@ -810,6 +884,25 @@ impl Db {
             .optional()?)
     }
 
+    pub fn get_lowest_block_view_number(&self) -> Result<Option<u64>> {
+        Ok(self
+            .pool
+            .get()?
+            .prepare_cached("SELECT MIN(view) FROM blocks")?
```
shawn-zil, line 891: I'm unsure, so I'm asking:
1. Does it matter if the row with the `min(view)` corresponds to the row with the `min(height)`?
2. Does it matter if the `min(view)` block is not a `canonical` block and therefore, not a valid block in the chain?

zilliqa/src/db.rs
```diff
@@ -486,6 +486,80 @@ impl Db {
...
+            .query_row([], |row| row.get(0))
+            .unwrap_or_default();
+        // to prevent primary key collision with missed views stored in the table
+        Ok(min_view - LARGE_OFFSET)
+    }
+
+    pub fn set_min_view_of_ckpt_view_history(&self, min_view: u64) -> Result<()> {
+        self.pool
+            .get()?
+            .prepare_cached("UPDATE ckpt_view_history SET view = ?1 WHERE leader IS NULL")?
```
shawn-zil, line 560: Add this query to the set of QPSG test queries below.

zilliqa/src/db.rs
```diff
@@ -486,6 +486,80 @@ impl Db {
...
+            //.execute(rusqlite::params![view, leader])?;
+            .execute((view, leader))?;
+        Ok(())
+    }
+
+    pub fn get_min_view_of_ckpt_view_history(&self) -> Result<u64> {
+        let min_view: u64 = self
+            .pool
+            .get()?
+            .prepare_cached("SELECT view FROM ckpt_view_history WHERE leader IS NULL LIMIT 1")?
```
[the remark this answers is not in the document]
DrZoltanFazekas: There is only one row with leader IS NULL so the result is consistent

zilliqa/src/consensus.rs
```diff
@@ -3664,6 +3773,71 @@ impl Consensus {
         // done
         tracing::info!("State-sync complete!");
         state_trie.finish_migration()?;
+        self.merge_missed_view_history()?;
         Ok(true)
     }
+
+    pub fn merge_missed_view_history(&mut self) -> Result<()> {
+        // state syncing finished at view >= cutover_at so the state
```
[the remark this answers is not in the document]
DrZoltanFazekas: Just a misleading comment, pushed to change to clarify what is actually meant

PR #3248, opened 2025-10-07
Ensure receipt-order matches transactions

shawn-zil:
When a `SELECT` is performed in SQLite with the `IN` filter, the results are returned in db-order, not in filter-order. This PR fixes the case when db-order is not the same as receipt-order.


DrZoltanFazekas (approved):

maxconway (approved):
LGTM - I think a different variant on this bug cropped up a while back

PR #3313, opened 2025-11-13
Fix z2 withdraw function call redirecting to unstake function instead of withdraw function.

chetan-zilliqa:


frankmeds (dismissed):

frankmeds (approved):

DrZoltanFazekas (approved):
The `withdraw()` function exists since deposit v2, so you could also add it to all versions before v4, but it doesn't make any difference for the actual fix.

PR #3341, opened 2025-12-02
Wrap `ViewHistory` in `Arc<RwLock<>>` to prevent excessive cloning

DrZoltanFazekas:
The goal of this PR is to stop cloning the whole `VecDeque` of the `ViewHistory` when finalizing blocks and calling the jailing precompile.


shawn-zil (approved):
Looks ok. Needs to be tested and checked in mainnet.

PR #3344, opened 2025-12-04
fix ci

shawn-zil:
This massive PR fixes a few things:
1. Fixes the upstream breakage in `svm-rs`; also updated some other crates with issues.
2. Migrates all `ethers` code to `alloy` code. The EOL-ed `ethers` crate is still using a broken `svm-rs` version. This crate is mostly used in `z2` and `tests`.
3. Migrated nearly all the tests. Commented out a couple of exceptions that I've yet to figure out how to migrate. I'll continue working on them, but it is better to get the rest in first.
4. Fixed a handful of tests that were not working; or were outdated.
5. Speed up jailing tests - now runs for ~25mins, down from previous ~50mins
6. Upgraded MSRV to 1.90.0 which is required by some crates.
7. General updates/cleanups.


frankmeds:
> Regarding the `z2` changes. I did some quick smoke testing and it seemed to work. The changes are minor but I'd be happier if @pabops / @frankmeds helped check.

@shawn-zil looks good to me

PR #3353, opened 2025-12-12
Randao support

bzawisto:
This PR introduces randao - which works as follows:
1. validators include `RANDAO_REVEAL(block) = BLS_SIGN(proposer.privkey, block.view)` in the randao_reveal field of block they propose
2. validator include `RANDAO_MIX(block) = RANDAO_MIX(block.parent) XOR KECCAK256(RANDAO_REVEAL(block))` in the mixHash field of block they propose
3. nodes check if `BLS_VERIFY(proposer.pubkey, block.view, RANDAO_REVEAL(block))` is true
4. nodes compute `RANDAO_MIX(block) = RANDAO_MIX(block.parent) XOR KECCAK256(RANDAO_REVEAL(block))` and check if included mixHash included in a block is the same

### DIFF
--- a/.github/workflows/docker-release.yaml
+++ b/.github/workflows/docker-release.yaml
@@ -15,6 +15,7 @@ jobs:
     permissions:
       id-token: write
       contents: write
+      security-events: write
     if: ${{ github.ref_name != 'main' && github.ref_type == 'tag' }} 
     runs-on: ubuntu-latest-16-cores
     env:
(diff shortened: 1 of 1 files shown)


zilliqa/src/checkpoint.rs
```diff
@@ -280,6 +280,28 @@ pub fn load_ckpt_blocks(path: &Path) -> Result<(Block, Vec<SignedTransaction>, B
     };
     ensure!(block.parent_hash() == parent.hash(), "Parent hash mismatch");
 
+    let grandparent = {
+        if let Ok(mut file) = zipreader.by_name("grandparent.bincode") {
```
shawn-zil: This is OK too, but I would prefer we check the `comment()` to see if it is `CKPT_VERSION_2_0` or `CKPT_VERSION_2_1` and take the appropriate action.

PR #3410, opened 2026-01-13
feat: DEVOPS-190 disks from snapshot in nodes parameter

pabops:
Required in case a disk makes a reference to a snapshot.

Reference example:
```
      api:
        boot_disk_size: 100
        data_disk_size: 50
        instance_type: e2-standard-2
        provisioning_model: STANDARD
        generate_external_ip: true
        detach_load_balancer: false
        alternative_ssl_domains: 
          api: ["api.testnet.zilliqa.com"]
        nodes:
        - count: 3
          region: asia-southeast1
        disk_snapshot_override:
          zq2-testnet-api-ase1-2: zq2-testnet-checkpoint-ase1-0-recovery-test-2
```


pabops:
Using ignore lifecycle instead.

PR #3443, opened 2026-01-29
Fix simple transfer gas estimate issue

bzawisto:
In case the recipient is an eoa and there's no calldata in request - we should try running simple transfer with default gas (21k).

### DIFF
--- a/zilliqa/src/exec.rs
+++ b/zilliqa/src/exec.rs
@@ -1288,7 +1288,7 @@ impl State {
             self.get_account(dest_addr)?.code.is_eoa()
         } else {
             false
-        }
+        };
 
         // For simple transfer try running with the default gas limit (21k) first
         if is_simple_transfer {


zilliqa/src/exec.rs
```diff
@@ -1282,6 +1282,37 @@ impl State {
     ) -> Result<u64> {
         let gas_price = gas_price.unwrap_or(self.gas_price);
 
+        let mut is_simple_transfer = false;
+        if data.is_empty()
+            && let Some(dest_addr) = to_addr
+        {
+            is_simple_transfer = self.get_account(dest_addr)?.code.is_eoa()
+        }
```
shawn-zil:
```suggestion
        let is_simple_transfer = if data.is_empty()
            && let Some(dest_addr) = to_addr
        {
            self.get_account(dest_addr)?.code.is_eoa()
        } else {
            false
        }
```

PR #3467, opened 2026-02-10
New feature: state pruning

shawn-zil:
This PR introduces the state-pruning feature. It has been tested successfully on a local node in `testnet` by restoring from a recent checkpoint, enabling `prune_interval` and monitoring the database statistics.

As an example:
```
@25474000 - 2795227 nodes (602MB)
@25537402 - 2745840 nodes (535MB)
```
The *nodes* here refer to nodes in the Patricia-Merkle-Tree.

As a comparison, a `testnet` full node has a 6.5GB database storage, and it only grows in size as more and more data is added to the PMT, never reduced. The data is validated by checking account balances across time.

Details are described in `docs/state_pruning.md` documentation.

Tested on `testnet` under different scenarios:
1. Successful key migration observed for non-pruning node.
2. Successful key migration and pruning observed for pruning node.
3. Successful pruning observed for new pruning node.

### DIFF
--- a/zilliqa/src/checkpoint.rs
+++ b/zilliqa/src/checkpoint.rs
@@ -481,7 +481,7 @@ pub fn save_ckpt(
 
     let mut account_count = u64::MIN;
     let mut record_count = u64::MIN;
-    let num_workers = crate::available_threads().max(1);
+    let num_workers = crate::available_threads();
     let (work_tx, work_rx) = crossbeam::channel::bounded::<(Vec<u8>, Vec<u8>)>(num_workers * 2); // 2 x Threads
     let (blob_tx, blob_rx) = crossbeam::channel::bounded::<AccountBlob>(num_workers * 2); // 2 x Threads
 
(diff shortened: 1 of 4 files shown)


shawn-zil:
Running this on `devnet`, we can see the difference in state nodes:

non-pruning `API0` node:
```
Keys in range: 225001
Value size distribution: 
Count: 225001 Average: 346.3749  StdDev: 265.51
Min: 1  Median: 394.4585  Max: 22311
Percentiles: P50: 394.46 P75: 487.25 P99: 576.33 P99.9: 579.68 P99.99: 1524.98
```

pruning `VAL0` node:
```
Keys in range: 186531
Value size distribution: 
Count: 186531 Average: 164.3262  StdDev: 273.02
Min: 1  Median: 143.1736  Max: 22310
Percentiles: P50: 143.17 P75: 164.03 P99: 561.53 P99.9: 578.54 P99.99: 16376.40
```

There are less keys in the pruned node because the tree is getting pruned periodically.

PR #3572, opened 2026-04-06
fix: route /health through Flask and add port 4201 reachability check

frankmeds:
## Summary

- Route `api.*/health` requests through the Flask health check (port 8080) instead of the Rust trivial endpoint (port 4201)
- Add a port 4201 TCP reachability check to the Flask health check so the LB detects when the user-facing server is overwhelmed
- Add explicit health check thresholds (`unhealthy_threshold = 3`) to avoid false positives from brief spikes

## Context

On 2026-04-04, the ZQ2 mainnet API returned HTTP 500 for ~3h while the GCP LB considered all 10 backends HEALTHY. The LB health check probed port 8080 (Flask process querying internal port 4202), but user traffic goes to port 4201 (Rust server, max 100 connections). When port 4201 was overwhelmed, port 8080 stayed healthy because it queries a different internal port.

### Why not just switch the health check to port 4201?

The Rust `/health` endpoint on port 4201 only returns `{"health": true}` unconditionally; it checks nothing. The Flask health check on port 8080 performs substantive checks:
- **Sync status**: queries `eth_blockNumber` and `eth_syncing` on port 4202; returns 503 if no blocks for 60s
- **Block production pace**: measures blocks/minute; returns 500 if < 5 blocks/min
- **Checkpoint status**: verifies GCS checkpoint freshness

Switching the LB to port 4201 would bypass all of these checks; solving the connection saturation problem but losing sync/stall detection.

### Approach: best of both worlds

Instead, this PR:
1. **Keeps** the LB health check on port 8080 (Flask) to preserve sync/block-production detection
2. **Adds** a TCP reachability check for port 4201 to Flask's `/health` endpoint, so it returns 503 when the Rust server is overwhelmed
3. **Routes** `api.*/health` requests to the Flask backend via a URL map `path_rule`, so users get meaningful health data

## Changes

| File | Change |
|------|--------|
| `infra/tf/api.tf` | Add explicit thresholds to `google_compute_health_check.api` (port stays 8080) |
| `infra/tf/api.tf` | Add `path_rule` in `api` path matcher to route `/health` and `/health/*` to `backend_service.health` (Flask) |
| `infra/tf/private_api.tf` | Add explicit thresholds to `google_compute_health_check.private_api` (port stays 8080) |
| `infra/ansible/.../healthcheck.py.j2` | Add `check_jsonrpc_port()`: TCP socket probe to `localhost:4201` with 2s timeout; integrated as first check in `check_sync_status()` |

No firewall changes needed: port 8080 firewalls already exist for both API and private API nodes.

No new Terraform resources: removed the `api_health_endpoint` resource from the previous approach.

## Test plan

- [ ] `terraform plan` on testnet: expect in-place update on health checks (add thresholds) + URL map update (add path_rule); no destroys
- [ ] `terraform apply` on testnet → `gcloud compute backend-services get-health` confirms all backends HEALTHY
- [ ] Deploy healthcheck.py via Ansible on testnet → verify `/health` returns 503 when port 4201 is stopped
- [ ] Verify `api.<testnet>/health` returns Flask health data (sync status), not `{"health": true}`
- [ ] Blackbox exporter probe for testnet still returns 200
- [ ] Apply to mainnet after successful testnet validation
- [ ] Monitor `probe_http_status_code` for `api.zq2-mainnet.zilliqa.com/health` for 24h

Closes DEVOPS-288

### DIFF
--- a/infra/tf/api.tf
+++ b/infra/tf/api.tf
@@ -56,7 +56,7 @@ resource "google_compute_health_check" "api" {
   name = "${var.chain_name}-jsonrpc"
 
   http_health_check {
-    port               = "4201"
+    port               = "8080"
     port_specification = "USE_FIXED_PORT"
     request_path       = "/health"
   }
@@ -65,16 +65,6 @@ resource "google_compute_health_check" "api" {
   unhealthy_threshold = 3
   check_interval_sec  = 5
   timeout_sec         = 5
-}
-
-resource "google_compute_health_check" "api_health_endpoint" {
-  name = "${var.chain_name}-health-endpoint"
-
-  http_health_check {
-    port               = "8080"
-    port_specification = "USE_FIXED_PORT"
-    request_path       = "/health"
-  }
 }
 
 resource "google_compute_backend_service" "api" {
@@ -101,7 +91,7 @@ resource "google_compute_backend_service" "api" {
 
 resource "google_compute_backend_service" "health" {
   name                  = "${var.chain_name}-api-health-nodes"
-  health_checks         = [google_compute_health_check.api_health_endpoint.id]
+  health_checks         = [google_compute_health_check.api.id]
   port_name             = "health"
   load_balancing_scheme = "EXTERNAL_MANAGED"
   enable_cdn            = false
@@ -147,6 +137,11 @@ resource "google_compute_url_map" "api" {
   path_matcher {
     name            = "api"
     default_service = google_compute_backend_service.api.id
+
+    path_rule {
+      paths   = ["/health", "/health/*"]
+      service = google_compute_backend_service.health.id
+    }
 
     default_route_action {
       cors_policy {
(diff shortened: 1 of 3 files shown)


infra/tf/api.tf
```diff
@@ -55,6 +55,21 @@ resource "google_compute_instance_group" "api" {
 resource "google_compute_health_check" "api" {
   name = "${var.chain_name}-jsonrpc"
 
+  http_health_check {
+    port               = "4201"
+    port_specification = "USE_FIXED_PORT"
+    request_path       = "/health"
```
shawn-zil, line 61: Could you quickly double check whether it responds on `/health` or `/health_check`?

https://github.com/Zilliqa/zq2/blob/main/zilliqa%2Fsrc%2Fapi%2Fmod.rs#L52

frankmeds: all the current check are to `/health`

shawn-zil (dismissed):

shawn-zil (dismissed):

PR #3612, opened 2026-04-27
Distribute validator rewards every epoch

bzawisto:
Once the feature is activated, the rewards are distributed every epoch block (3600). This should reduce the growth of the state.


zilliqa/src/cfg.rs
```diff
@@ -448,7 +448,7 @@ pub fn failed_request_sleep_duration_default() -> Duration {
 }
 
 pub fn max_missed_view_age_default() -> u64 {
-    MISSED_VIEW_WINDOW
+    MISSED_VIEW_WINDOW * (blocks_per_epoch_default() + 1)
```
shawn-zil, line 451: unclear to me why it requires `3601 * MISSED_VIEWS`. do we need more than `MISSED_VIEW_WINDOW + 3600` to compute the last set of rewards?

shawn-zil (approved):

zilliqa/src/cfg.rs
bzawisto: This would work iff there are no gaps in views. However, these are possible and we expect that 3600-wide window is sufficient for each view.

